PR #3, opened 2017-08-28
Tests hang on travis (WIP)

olydis:


olydis:
will close since this has 0 changes and was essentially a probe

PR #6, opened 2017-08-29
client-side-validation flag support

olydis:
replay of https://github.com/Azure/autorest/pull/2490

### DIFF
--- a/src/Model/MethodGo.cs
+++ b/src/Model/MethodGo.cs
@@ -14,6 +14,7 @@ using System.Globalization;
 using System.Linq;
 using System.Net;
 using System.Text;
+using static AutoRest.Core.Utilities.DependencyInjection;
 
 namespace AutoRest.Go.Model
 {
@@ -211,6 +212,10 @@ namespace AutoRest.Go.Model
             }
 
             return $"{retVal}, {errVal}";
+        }
+
+        public bool ShouldValidate() {
+            return !ParameterValidations.IsNullOrEmpty();
         }
 
         public string NextMethodName => $"{Name}NextResults";
--- a/src/Model/MethodGo.cs
+++ b/src/Model/MethodGo.cs
@@ -216,7 +216,8 @@ namespace AutoRest.Go.Model
 
         public bool ShouldValidate()
         {
-            return AutoRest.Core.Settings.Instance.Host?.GetValue<bool?>("client-side-validation").Result ?? false;
+            return true == AutoRest.Core.Settings.Instance.Host?.GetValue<bool?>("client-side-validation").Result
+                && !ParameterValidations.IsNullOrEmpty();
         }
 
         public string NextMethodName => $"{Name}NextResults";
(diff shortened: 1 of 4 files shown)


src/Model/MethodGroupGo.cs
```diff
@@ -86,7 +87,7 @@ internal void Transform(CodeModelGo cmg)
                     {
                         mg.ReturnType.Body.AddImports(imports);
                     }
-                    if (!string.IsNullOrEmpty(mg.ParameterValidations))
+                    if (mg.ShouldValidate())
                         validationImports = true;
```
jhendrixMSFT: Minor nit, you could just do this.
```
validationImports = mg.ShouldValidate()
```

jhendrixMSFT: Actually no you can't, never mind.

marstr: You could do this though!

``` C#
useValidation |= mg.ShouldValidate();
```
That said, I'm not sure this is any more readable.

src/Model/MethodGo.cs
```diff
@@ -213,6 +214,12 @@ public string MethodReturnSignature(bool helper)
             return $"{retVal}, {errVal}";
         }
 
+        public bool ShouldValidate()
+        {
+            return true == AutoRest.Core.Settings.Instance.Host?.GetValue<bool?>("client-side-validation").Result
+                && !ParameterValidations.IsNullOrEmpty();
```
marstr: [nit] This is sort of a needlessly long line. I think it would be more readable as:
``` C#
public bool ShouldValidate()
{
    var clientValidation = AutoRest.Core.Settings.Insance.Host?.GetValue<bool?>("client-side-validation").Result;
    return clientValidation && !ParameterValidations.IsNullOrEmpty();
}
```

Preferably, you could reduce down that first line even more :)

olydis: blame me for the long line, that was my interactive rebasing and adjusting to new project structure on the fly... didn't even check if there's a `using AutoRest.Core;` in that file

jhendrixMSFT: While we're discussing this chunk of code I don't understand why the "ParameterValidations.IsNullOrEmpty()" clause is necessary.  If the settings that we get from core say "disable validation" isn't that enough?

marstr: @jhendrixMSFT, I believe that `IsNullOrEmpty()` check still makes sense. It could spare us generating code to allocate what would be an empty slice of validations to enforce, for example.

jhendrixMSFT: If that's the case then there's more to fix here IMO.  It really should be as simple as "if the flag is set to true then don't emit any validation code" right?

marstr: Fair enough :) Truthfully, I'm not trying to say what should be, just trying to explain why that call would be there.

mcardosos:
Reopened https://github.com/Azure/autorest.go/pull/11

PR #7, opened 2017-08-29
Added pkg and bin from GOPATH to gitignore

mcardosos:
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -290,6 +290,5 @@ __pycache__/
 package-lock.json
 test/src/tests/generated/*
 test/src/tests/vendor/*
-glide.lock
 test/pkg
 test/bin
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -290,4 +290,6 @@ __pycache__/
 package-lock.json
 test/src/tests/generated/*
 test/src/tests/vendor/*
-glide.lock
\ No newline at end of file
+glide.lock
```
marstr: I see that you're not adding `glide.lock` to the ignore file, but I don't think it should have present in the first place. Mind removing it?

mcardosos: Well, the glide up goal is to grab the latest version for the runtime. Dependencies for testing should not be tied to specific commits.

marstr: Hmm, I guess I disagree. I think dependencies for testing should be handled the same way as others. Also, I think that `glide install` is the more typical command that should be run over `glide up`.

PR #9, opened 2017-08-31
Print full autorest command

mcardosos:
PTAL
@jhendrixMSFT @marstr @olydis


mcardosos:
@marstr this is so we know all arguments used by `npm test`. One of the arguments is currently hidden and there is a big change on including it or not on the command. This is just for transparency.

PR #10, opened 2017-08-31
Get rid of autorest.home var

mcardosos:
PTAL @marstr @jhendrixMSFT @fearthecowboy


mcardosos:
Fixed on https://github.com/Azure/autorest.go/pull/12

PR #12, opened 2017-08-31
debugger

olydis:


mcardosos:
>   autorest_home: process.env["autorest.home"]

Is this still needed?

---
Refers to: .gulp/common.iced:410 in 35d47fd. [](commit_id = 35d47fd274a2d1bc38f8b9401e6da0aa01a5037b, deletion_comment = False)

olydis:
@mcardosos nope, well spotted!

PR #13, opened 2017-09-01
x-ms-discriminator-value

mcardosos:
Closing https://github.com/Azure/autorest.go/pull/4 and opening this
PTAL @jhendrixMSFT @marstr


olydis:
@mcardosos I cancelled `win32-x64`, consider this passed 😄

olydis:
@mcardosos great, and commenting makes the job think a new commit was done, so now Linux reruns... feel free to merge X'D

mcardosos:
Sanity checks
https://github.com/mcardosos/postgresql-sample
https://github.com/mcardosos/commerce-sample

src/Templates/EnumTemplate.cshtml
```diff
@@ -13,10 +13,13 @@
 type @Model.Name string
 
 @EmptyLine
-const (
+@if (constants.Count() > 0)
```
marstr: Changing this to `constants.Any()` will reduce this line from O(n) to O(1)

src/Extensions.cs
```diff
@@ -113,6 +113,11 @@ public static string[] ToWords(this string value)
             return WordSplitPattern.Split(value).Where(s => !string.IsNullOrEmpty(s)).ToArray();
         }
 
+        public static string ToIdiomaticShortName(this string value)
+        {
+            return CodeNamerGo.Instance.GetVariableName(string.Concat(value.ToPhrase().Split(new char[]{' '}).Select(w => w[0])));
```
marstr: I think this could be rewritten to be a little bit easier to read. Maybe,

``` C#
public static string ToIdiomaticShortName(this string longName)
{
    var initials = from word in longName.ToPhrase().Split(new []{' '})
                         select word[0];
    var acronym = string.Concat(initials);
    return CodeNamerGo.Instance.GetVariableName(acronym);
}
```

src/Model/CompositeTypeGo.cs
```diff
@@ -26,6 +27,28 @@ public class CompositeTypeGo : CompositeType
...
+                var st = (BaseModelType as CompositeTypeGo).DerivedTypes;
+                if (BaseModelType.BaseModelType != null && BaseModelType.BaseModelType.IsPolymorphic)
+                {
+                    st = st.Union((BaseModelType as CompositeTypeGo).SiblingTypes);
+                }
+                return st;
+            }
+        }
+
+        public bool HasPolymorphicFields => AllProperties.Any(p => (p.ModelType is CompositeType && (p.ModelType as CompositeTypeGo).IsPolymorphic)
```
marstr: Too much shortening! Lets give this a real method body.

src/Model/CompositeTypeGo.cs
```diff
@@ -102,16 +126,88 @@ public CompositeTypeGo(IModelType wrappedType)
         }
 
         /// <summary>
+        /// If PolymorphicDiscriminator is set, makes sure we have a PolymorphicDiscriminator property.
+        /// </summary>
+        private void AddPolymorphicPropertyIfNecessary()
+        {
+            if (!string.IsNullOrEmpty(PolymorphicDiscriminator) && Properties.All(p => p.SerializedName != PolymorphicDiscriminator))
+            {
+                var newProp = base.Add(New<Property>(new
```
marstr: Why are we saving the result of `base.Add`? It doesn't look like it gets used.

src/Extensions.cs
```diff
@@ -113,6 +113,11 @@ public static string[] ToWords(this string value)
             return WordSplitPattern.Split(value).Where(s => !string.IsNullOrEmpty(s)).ToArray();
         }
 
+        public static string ToIdiomaticShortName(this string value)
```
marstr: Shouldn't all names be Idiomatic? Maybe a more appropriate name would be `ToShortName`

src/Extensions.cs
jhendrixMSFT: To throw another option into the mix, I wrote the follow function for creating vars from pascal-cased strings.
```
public static string ToShortVarName(this string name)
        {
            var sb = new StringBuilder();
            for (int i = 0; i < name.Length; ++i)
            {
                if (char.IsUpper(name[i]))
                {
                    sb.Append(char.ToLowerInvariant(name[i]));
                }
            }
            return sb.ToString();
        }
```

src/Extensions.cs
```diff
@@ -355,8 +360,23 @@ public static void AddImports(this IModelType type, HashSet<string> imports)
         }
 
         public static bool ShouldBeSyntheticType(this IModelType type)
+        {            
+            return (type is PrimaryType || type is SequenceType || type is DictionaryType || type is EnumType || 
+                (type is CompositeTypeGo && (type as CompositeTypeGo).IsPolymorphicResponse()));
```
jhendrixMSFT: Why do we need to generate a synthetic type for polymorphic responses?

mcardosos: Because they need a wrapper to be unmarshalled

src/Templates/EnumTemplate.cshtml
mcardosos: Fixed

src/Extensions.cs
mcardosos: Fixed

src/Extensions.cs
mcardosos: Fixed

src/Model/CompositeTypeGo.cs
mcardosos: Made all this more readable

src/Model/CompositeTypeGo.cs
mcardosos: Fixed

PR #15, opened 2017-09-18
Replaced FixedValue with Value

mcardosos:


mcardosos:
PTAL @jhendirxMSFT @marstr

mcardosos:
Fixes [this Travis failure](https://travis-ci.org/Azure/azure-sdk-for-go/jobs/275144765#L667).
According to @olydis , `FixedValue` should not be used unless it is strictly necessary,

mcardosos:
Most important effect: there are no breaking changes in the SDK.
Interesting effects:
- `whatevergroup.go` files got renamed to `whatever.go`
- Bunch of dumb validation code generated

src/Model/ParameterGo.cs
```diff
@@ -234,6 +234,11 @@ public static IEnumerable<ParameterGo> QueryParameters(this IEnumerable<Paramete
 
         public static string Validate(this IEnumerable<ParameterGo> parameters, HttpMethod method)
         {
+            if (!parameters.Any())
+            {
+                return null;
```
marstr: Nit: I think it would be more appropriate to return string.Empty here. :)

mcardosos: Discussing this with Joel, this lines were not needed at all. So deleted

PR #16, opened 2017-09-21
Added 429 as possible response status code

mcardosos:
PTAL
@jhendrixMSFT @marstr 
This will allow us to gen from `web` swagger

### DIFF
--- a/src/CodeNamerGo.cs
+++ b/src/CodeNamerGo.cs
@@ -151,7 +151,8 @@ namespace AutoRest.Go
                 HttpStatusCode.BadGateway,
                 HttpStatusCode.ServiceUnavailable,
                 HttpStatusCode.GatewayTimeout,
-                HttpStatusCode.HttpVersionNotSupported
+                HttpStatusCode.HttpVersionNotSupported,
+                (HttpStatusCode)429
             })
             {
                 statusCodeMap.Add(sc, string.Format("http.Status{0}", sc));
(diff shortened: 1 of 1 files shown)


src/CodeNamerGo.cs
```diff
@@ -151,7 +151,8 @@ public CodeNamerGo()
                 HttpStatusCode.BadGateway,
                 HttpStatusCode.ServiceUnavailable,
                 HttpStatusCode.GatewayTimeout,
-                HttpStatusCode.HttpVersionNotSupported
+                HttpStatusCode.HttpVersionNotSupported,
+                (HttpStatusCode)429
```
marstr: Nit it seems like these are clustered by their first digit. That would mean that 429 is currently lumped with the 500s, but it should be with the 400s.

Separately, I've verified that this doesn't seem to be part of .NET's Status enumeration, documentation can be found here: https://msdn.microsoft.com/en-us/library/system.net.httpstatuscode(v=vs.110).aspx

mcardosos: Fixed, lololol

marstr: Lol race conditions

src/CodeNamerGo.cs
```diff
@@ -145,6 +145,7 @@ public CodeNamerGo()
                 HttpStatusCode.UnsupportedMediaType,
                 HttpStatusCode.RequestedRangeNotSatisfiable,
                 HttpStatusCode.ExpectationFailed,
+                (HttpStatusCode)429,
```
jhendrixMSFT: Can you define this as a const?

mcardosos:
Merging when Johannes goes green

PR #18, opened 2017-09-22
Next Link is named correctly and no longer duplicated

mcardosos:
Fixes https://github.com/Azure/autorest/issues/2219
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -260,6 +260,7 @@ namespace AutoRest.Go.Model
                 }
                 else
                 {
+                    // NextLinks might have differences in casing, but they need to be consistent
                     if (property.Name.EqualsIgnoreCase(NextLink))
                     {
                         property.Name = NextLink;
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -260,6 +260,10 @@ namespace AutoRest.Go.Model
                 }
                 else
                 {
+                    if (property.Name.EqualsIgnoreCase(NextLink))
+                    {
+                        property.Name = NextLink;
+                    }
                     indented.AppendFormat("{0} *{1} {2}\n", property.Name, property.ModelType.Name, property.JsonTag());
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Model/CompositeTypeGo.cs
```diff
@@ -226,35 +226,6 @@ public string Fields()
                 indented.Append(((CompositeTypeGo)BaseModelType).Fields());
             }
 
-            // If the type is a paged model type, ensure the nextLink field exists
-            // Note: Inject the field into a copy of the property list so as to not pollute the original list
-            if (!string.IsNullOrEmpty(NextLink))
```
jhendrixMSFT: If NextLink is no longer required please remove it.

mcardosos: It is still used on the template for paged methods convinience.

src/Model/MethodGo.cs
```diff
@@ -557,27 +557,27 @@ public string NextLink
...
-                        var nextLinkName = (string)pageableExtension["nextLinkName"];
-                        if (!string.IsNullOrEmpty(nextLinkName))
+                        var nextLink = (string)pageableExtension["nextLinkName"];
+                        if (!string.IsNullOrEmpty(nextLink))
                         {
-                            nextLink = CodeNamerGo.PascalCaseWithoutChar(nextLinkName, '.');
+                            var nextLinkProperty = CodeNamerGo.Instance.GetPropertyName(CodeNamerGo.PascalCaseWithoutChar(nextLink, '.'));
+                            if ((ReturnType.Body as CompositeTypeGo).Properties.Any(p => p.Name.EqualsIgnoreCase(nextLinkProperty)))
+                            {
+                                (ReturnType.Body as CompositeTypeGo).Properties.First(p => p.Name.EqualsIgnoreCase(nextLinkProperty)).Name = nextLinkProperty;
```
jhendrixMSFT: It seems dubious that a getter is mutating state like this.  Shouldn't the name have been fixed up during transformation?

mcardosos: Fixed

olydis:
# consider this as passing

Retriggered locally, went through. Sorry about the servers failing, we're fine tuning auto-publishing on PR merge right now...

olydis:
# publish job

PR #23, opened 2017-09-25
Casing and Pascal case

mcardosos:
PTAL @jhendrixMSFT
@marstr

### DIFF
--- a/src/CodeNamerGo.cs
+++ b/src/CodeNamerGo.cs
@@ -269,7 +269,7 @@ namespace AutoRest.Go
         /// <param name="name"></param>
         /// <param name="splitter"></param>
         /// <returns>The formatted string</returns>
-        public static string PascalCaseWithoutChar(string name, char splitter)
+        public override string PascalCase(string name)
         {
             if (string.IsNullOrWhiteSpace(name))
             {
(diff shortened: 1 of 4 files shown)


mcardosos:
I could create some breaking changes around casing

olydis:
> linux-x64 restart

olydis:
> win32-x64 restart

olydis:
@jhendrixMSFT @mcardosos @marstr Note that CI is now failing due to the new regression test: your code generator seems to be non-deterministic in terms of the order it emits certain stuff within a file... always feel free to merge anyways after visually inspecting that the diff is okay - this is just supposed to be a way to bring your attention to *unintended* changes you made. Note that in this case for example I see a casing change from "Pname" to "PName" and "Odatavalue" to "OdataValue" that may or may not have been intended (probably the former, looking at the title of this PR) :) @mcardosos did you commit after running `gulp test`/`npm test`?

mcardosos:
Yes, I ran npm test and modified some stuff in the tests, yes

olydis:
@mcardosos well, but the question is whether it's ***committed***, or why does CI see this change?
image

mcardosos:
Expected diff

olydis:
# publish job

olydis:
@mcardosos I think we have an educational issue about what the CI job does - you're supposed to *commit* expected changes, so the job turns green! Otherwise (as will be the case now), this diff will grow and grow with each PR. Whoever makes the next PR to this repo will see the above diff I posted *again*, although it has nothing to do with their change. The idea is to *commit* the generated files, so reviewers see and can judge whether those changes are reasonable and expected!

olydis:
What I meant with my previous comment: You also have inconsistencies between running the Go generator on Linux vs Windows, so I don't expect both jobs to turn green unless you make your internal sorting platform agnostic. For anything else, you should *not* see a diff in the job or something is wrong with your PR ;)

PR #24, opened 2017-09-26
Do not name files with file names reserved for autorest

mcardosos:
Some of the new swaggers happens to have a method group called `version`.
PTAL
@jhendrixMSFT @marstr

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -65,17 +65,19 @@ namespace AutoRest.Go
 
             // by convention the methods in the method group with an empty
             // name go into the client template so skip them here.
-            HashSet<string> ReservedFiles = new HashSet<string>();
-            ReservedFiles.Add("models");
-            ReservedFiles.Add("client");
-            ReservedFiles.Add("version");
+            HashSet<string> ReservedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase){
+                "models",
+                "client",
+                "version",
+            };
 
             foreach (var methodGroup in codeModel.MethodGroups.Where(mg => !string.IsNullOrEmpty(mg.Name)))
             {
-                if (ReservedFiles.Contains(methodGroup.Name.Value.ToLowerInvariant()))
+                if (ReservedFiles.Contains(methodGroup.Name.Value))
                 {
                     methodGroup.Name += "group";
                 }
+                ReservedFiles.Add(methodGroup.Name);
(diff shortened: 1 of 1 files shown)


src/CodeGeneratorGo.cs
```diff
@@ -64,8 +65,17 @@ public override string ImplementationFileExtension
...
             // by convention the methods in the method group with an empty
             // name go into the client template so skip them here.
+            HashSet<string> ReservedFiles = new HashSet<string>();
+            ReservedFiles.Add("models");
+            ReservedFiles.Add("client");
+            ReservedFiles.Add("version");
+
             foreach (var methodGroup in codeModel.MethodGroups.Where(mg => !string.IsNullOrEmpty(mg.Name)))
             {
+                if (ReservedFiles.Contains(methodGroup.Name.Value.ToLowerInvariant()))
```
jhendrixMSFT: You can get rid of the ToLowerInvariant() if you specify a comparison object in the ctor for the hash set.

src/CodeGeneratorGo.cs
```diff
@@ -64,8 +65,17 @@ public override string ImplementationFileExtension
 
             // by convention the methods in the method group with an empty
             // name go into the client template so skip them here.
+            HashSet<string> ReservedFiles = new HashSet<string>();
+            ReservedFiles.Add("models");
+            ReservedFiles.Add("client");
+            ReservedFiles.Add("version");
```
marstr: [nit] This could be abbreviated to the following:

``` C#
HashSet<string> ReservedFiles = new HashSet<string>(){
    "models",
    "client",
    "version",
};
```

src/CodeGeneratorGo.cs
mcardosos: Fixed

src/CodeGeneratorGo.cs
mcardosos: Fixed

src/CodeGeneratorGo.cs
```diff
@@ -64,8 +65,19 @@ public override string ImplementationFileExtension
 
             // by convention the methods in the method group with an empty
             // name go into the client template so skip them here.
+            HashSet<string> ReservedFiles = new HashSet<string>(StringComparer.OrdinalIgnoreCase){
```
jhendrixMSFT: Nit, opening brace in wrong location.

mcardosos: Fixed

olydis:
# publish job

PR #25, opened 2017-09-28
Add generated code

mcardosos:
After @olydis explained stuff here https://github.com/Azure/autorest.go/pull/23


olydis:
# publish job

PR #30, opened 2017-10-16
Autoregister  RPs

mcardosos:
Gens code that uses https://github.com/Azure/go-autorest/pull/176
PTAL @marstr @jhendrixMSFT

### DIFF
--- a/.gulp/gulpfile.iced
+++ b/.gulp/gulpfile.iced
@@ -24,7 +24,6 @@ task 'init', "" ,(done)->
 task 'test', "", ['regenerate'], (done) ->
   process.env.GOPATH = "#{basefolder}/test"
   await execute "glide install",          { cwd: './test/src/tests' }, defer code, stderr, stdout
-  await execute "go fmt ./generated/...", { cwd: './test/src/tests' }, defer code, stderr, stdout
   await execute "go run ./runner.go",     { cwd: './test/src/tests' }, defer code, stderr, stdout
   done();
 
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -187,21 +187,9 @@ namespace AutoRest.Go
         /// </summary>
         /// <param name="arguments"></param>
         /// <returns></returns>
-        public static string EmitAsArguments(this IList<string> arguments)
+        public static string EmitAsArguments(this IEnumerable<string> arguments)
         {
-            var sb = new StringBuilder();
-
-            if (arguments.Count() > 0)
-            {
-                sb.Append(arguments[0]);
-
-                for (var i = 1; i < arguments.Count(); i++)
-                {
-                    sb.AppendLine(",");
-                    sb.Append(arguments[i]);
-                }
-            }
-            return sb.ToString();
+            return String.Join(",\n", arguments);
         }
 
 
--- a/src/Model/MethodGo.cs
+++ b/src/Model/MethodGo.cs
@@ -82,6 +82,8 @@ namespace AutoRest.Go.Model
             // Registering Azure resource providers should only happen with Azure resource manager REST APIs
             // This depends on go-autorest here:
             // https://github.com/Azure/go-autorest/blob/c0eb859387e57a164bf64171da307e2ef8168b58/autorest/azure/rp.go#L30
+            // As registering needs the Azure subscription ID, we take it from the operation path, on the
+            // assumption that ARM APIs should include the subsriptin ID right after `subscriptions`
             RegisterRP = cmg.APIType.EqualsIgnoreCase("arm") && Url.Split("/").Any(p => p.EqualsIgnoreCase("subscriptions"));
         }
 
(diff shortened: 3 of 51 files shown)


olydis:
~~~ Haskell
~~~

marstr:
Before I dive in, it seems like this is at odds with some of @jhendrixMSFT changes. Mind if I wait to review until @jhendrixMSFT has weighed in?

jhendrixMSFT:
@mcardosos I don't think this was properly merged with my changes from last night (looks like the v1 directory was lost).

mcardosos:
~~~
> linux-x64 restart
~~~

mcardosos:
> win32-x64 restart

mcardosos:
PTAL
@jhendrixMSFT @marstr

.gulp/gulpfile.iced
```diff
@@ -24,7 +24,6 @@ task 'init', "" ,(done)->
 task 'test', "", ['regenerate'], (done) ->
   process.env.GOPATH = "#{basefolder}/test"
   await execute "glide install",          { cwd: './test/src/tests' }, defer code, stderr, stdout
-  await execute "go fmt ./generated/...", { cwd: './test/src/tests' }, defer code, stderr, stdout
```
jhendrixMSFT: Why was this pushed to regeneration?  I'm not pushing back just curious.

src/Model/MethodGo.cs
```diff
@@ -76,6 +78,8 @@ internal void Transform(CodeModelGo cmg)
             {
                 Description += lroDescription;
             }
+
+            RegisterRP = cmg.APIType.EqualsIgnoreCase("arm") && Url.Split("/").Any(p => p.EqualsIgnoreCase("subscriptions"));
```
jhendrixMSFT: Add a comment explaining this.

.gulp/gulpfile.iced
mcardosos: Because when generated code changes, the changes should be checked in. It should be already formatted code, but right now only generating does not format. I dont want to run tests just to format code.

src/Model/MethodGo.cs
mcardosos: Added

olydis:
~~~ Haskell
~~~

src/Model/MethodGo.cs
```diff
@@ -76,6 +78,13 @@ internal void Transform(CodeModelGo cmg)
             {
                 Description += lroDescription;
             }
+
+            // Registering Azure resource providers should only happen with Azure resource manager REST APIs
+            // This depends on go-autorest here:
+            // https://github.com/Azure/go-autorest/blob/c0eb859387e57a164bf64171da307e2ef8168b58/autorest/azure/rp.go#L30
+            // As registering needs the Azure subscription ID, we take it from the operation path, on the
+            // assumption that ARM APIs should include the subsriptin ID right after `subscriptions`
```
marstr: Typo: `subsriptin`

src/Model/MethodGo.cs
```diff
@@ -390,11 +399,32 @@ public string HTTPMethodDecorator
             }
         }
 
+        public List<string> SendDecorators
```
marstr: Looks like you're just following suit, which is reasonable, but mind if we make this an `IEnumerable<string>` instead of a `List<string>`?

src/Model/MethodGo.cs
```diff
@@ -33,6 +33,8 @@ public class MethodGo : Method
         public bool IsCustomBaseUri
             => CodeModel.Extensions.ContainsKey(SwaggerExtensions.ParameterizedHostExtension);
 
+        public bool RegisterRP;
```
marstr: Mind either renaming this, or adding a comment making clear that this controls whether or not a `RegisterRPSenderDecorator` will be added? If it were renamed, maybe something like `DecorateUsingRegisterRP`? I know that's not a very concise name, so I would totally understand just commenting.

src/Model/MethodGo.cs
mcardosos: Fixed

src/Model/MethodGo.cs
mcardosos: Fixed

src/Model/MethodGo.cs
mcardosos: Fixed

src/Extensions.cs
```diff
@@ -187,21 +187,9 @@ public static string TrimPackageName(this string value, string packageName)
...
-                sb.Append(arguments[0]);
-
-                for (var i = 1; i < arguments.Count(); i++)
-                {
-                    sb.AppendLine(",");
-                    sb.Append(arguments[i]);
-                }
-            }
-            return sb.ToString();
+            return String.Join(",\n", arguments);
```
marstr: This is awesome!!!!!

mcardosos: Success!!!

PR #33, opened 2017-10-18
Make tests run faster.

jhendrixMSFT:
Some tests send failures to test retries.  The default retry delay is 30
seconds which makes total test time way too long (the default 30 second
delay should be investigated separately).  Set the retry duration to one
second so tests execute faster.


olydis:
~~~ Haskell
~~~

PR #37, opened 2017-10-20
Initial commit for the v3 generator.

jhendrixMSFT:
This updates the generator to use the new pipeline model introduced in
the new storage SDK.  It works for the storage swagger but is not yet
ready to be used for ARM as there is missing and/or broken codegen when
compared to the v2 generator.
A handful of tests have been updated but a lot have been commented out.


src/CodeGeneratorGo.cs
```diff
@@ -79,21 +71,55 @@ public override string ImplementationFileExtension
                     methodGroup.Name += "group";
                 }
                 ReservedFiles.Add(methodGroup.Name);
-                var methodGroupTemplate = new MethodGroupTemplate
-                {
-                    Model = methodGroup
-                };
-                await Write(methodGroupTemplate, FormatFileName(methodGroup.Name).ToLowerInvariant());
+                var methodGroupTemplate = new MethodGroupTemplate { Model = methodGroup };
+                await Write(methodGroupTemplate, FormatFileName(methodGroup.Name));
```
mcardosos: To lower is still a good idea, to avoid git issues

mcardosos: Oh, it is other place, nvm

jhendrixMSFT: You are right, will fix.

jhendrixMSFT: Hah so it is.  :)

src/CodeGeneratorGo.cs
```diff
@@ -79,21 +71,55 @@ public override string ImplementationFileExtension
...
+                new Tuple<ITemplate, string>(new ResponseError() { Model = codeModel }, "response_error"),
+                new Tuple<ITemplate, string>(new Validation() { Model = codeModel }, "validation")
+            };
+            return list;
         }
 
         private string FormatFileName(string fileName)
         {
-            return $"{fileName}{ImplementationFileExtension}";
+            var prefix = Settings.Instance.Host?.GetValue<string>("file-prefix").Result;
```
mcardosos: >prefix [](start = 16, length = 6)

Okok, it is this the one that would need to lower

src/CodeGeneratorGo.cs
```diff
@@ -79,21 +71,55 @@ public override string ImplementationFileExtension
...
+            {
+                var marshallingTemplate = new Marshalling { Model = codeModel };
+                await Write(marshallingTemplate, FormatFileName("marshalling"));
+            }
+        }
+
+        /// <summary>
+        /// Returns a collection of templates to emit.
+        /// </summary>
+        /// <param name="codeModel">The code moel that the template will consume.</param>
```
mcardosos: typo

src/CodeGeneratorGo.cs
```diff
@@ -79,21 +71,55 @@ public override string ImplementationFileExtension
...
+        /// <returns>A collection of templates, may be empty.</returns>
+        private IEnumerable<Tuple<ITemplate, string>> GetFixedTemplates(CodeModelGo codeModel)
+        {
+            var list = new List<Tuple<ITemplate, string>>
+            {
+                new Tuple<ITemplate, string>(new ResponderPolicy() { Model = codeModel }, "responder_policy"),
+                new Tuple<ITemplate, string>(new ResponseError() { Model = codeModel }, "response_error"),
+                new Tuple<ITemplate, string>(new Validation() { Model = codeModel }, "validation")
```
mcardosos: You might need to add these to the ReservedFiles too

src/CodeNamerGo.cs
```diff
@@ -326,9 +336,7 @@ public override string GetMethodGroupName(string name)
                 return name;
             }
 
-            // we use the base implementation here as it uses a case-insensitive comparison.
-            // this is a bit of a hacky work-around for some naming changes introduced in core...
-            return EnsureNameCase(RemoveInvalidCharacters(PascalCase(base.GetEscapedReservedName(name, "Group"))));
+            return EnsureNameCase(RemoveInvalidCharacters(PascalCase(name)));
```
mcardosos: >name [](start = 69, length = 4)

Success!!!

src/Model/CompositeTypeGo.cs
```diff
@@ -206,56 +188,85 @@ public string AddHTTPResponse()
...
                     // important, i.e. we don't want to flatten primary types like dictionaries.
-                    indented.AppendFormat("*{0} {1}\n", property.ModelType.Name, property.JsonTag());
+                    indented.AppendFormat("*{0} {1}\n", property.ModelType.Name, property.Tag());
                     property.Extensions[SwaggerExtensions.FlattenOriginalTypeName] = Name;
                 }
+                else if (!string.IsNullOrEmpty(NextLink) && property.Name.EqualsIgnoreCase(NextLink))
+                {
+                    // use custom type instead of *string
+                    indented.Append($"{NextLink} Marker `xml:\"{NextLink}\"`");
+                }
```
mcardosos: Could there be a case in which this is generated for an API that uses json?

src/Model/CompositeTypeGo.cs
```diff
@@ -264,7 +275,19 @@ public string Fields()
                     {
                         property.Name = NextLink;
                     }
-                    indented.AppendFormat("{0} *{1} {2}\n", property.Name, property.ModelType.Name, property.JsonTag());
+
+                    // to avoid any breaking changes in the v1 template always emit as a pointer type
```
mcardosos: >v1 [](start = 60, length = 2)

there are not v1 templates anymore

src/Templates/ServiceClientTemplate.cshtml
```diff
@@ -1,125 +1,66 @@
...
-@foreach (var p in Model.Properties)
-{
-    if (!p.SerializedName.IsApiVersion())
-    {
-            @:@(string.Format((p.IsRequired || p.ModelType.CanBeEmpty()
-                                                  ? "{0} {1}" : "{0} *{1}"),
-                                                  p.Name, p.ModelType.Name))
-    }
-}
+    url url.URL
```
mcardosos: >url url.UR [](start = 4, length = 10)

If this url is replacing the BaseURI, and there is no way to change it, sovereign cloud users will have a bad time

src/Templates/MethodTemplate.cshtml
```diff
@@ -24,272 +26,126 @@
...
-        err = @(Model.AutorestError("Failure sending request", "resp"))
-        return
-    }
-@EmptyLine
-    result, err = client.@(Model.ResponderMethodName)(resp)
+	req, err := client.@(opIdCamelCase)Preparer(@(Model.HelperInvocationParameters(false, false)))
+	if err != nil {
+		return nil, err
+	}
+	resp, err := client.Pipeline().Do(ctx, responderPolicyFactory{responder: client.@(opIdCamelCase)Responder}, req)
```
mcardosos: >ctx [](start = 35, length = 3)

Context might or not be generated?

src/Templates/ModelsTemplate.cshtml
```diff
@@ -1,18 +1,15 @@
...
+@using System.Linq
 
 @inherits AutoRest.Core.Template<AutoRest.Go.Model.CodeModelGo>
 @{
-    var enums = Model.EnumTypes.Cast<EnumTypeGo>().ToList();
-    enums.Sort(delegate(EnumTypeGo lhs, EnumTypeGo rhs) { return lhs.Name.FixedValue.CompareTo(rhs.Name); });
-
-    var modelTypes = Model.ModelTypes.Cast<CompositeTypeGo>().ToList();
-    modelTypes.Sort(delegate (CompositeTypeGo x, CompositeTypeGo y) { return x.Name.Value.CompareTo(y.Name.Value); });
+    var imports = Model.ModelImports;
```
mcardosos: >imports [](start = 8, length = 7)

This is not used

mcardosos:
Question, how much would the SDK break with this changes?

src/CodeGeneratorGo.cs
jhendrixMSFT: Will fix.

src/CodeGeneratorGo.cs
jhendrixMSFT: Will fix.

src/CodeGeneratorGo.cs
jhendrixMSFT: Will fix.

src/Model/CompositeTypeGo.cs
jhendrixMSFT: Good catch will fix.

src/Model/CompositeTypeGo.cs
jhendrixMSFT: Will fix.

src/Templates/ServiceClientTemplate.cshtml
jhendrixMSFT: This will be addresses when I get the v3 generator working for ARM.

src/Templates/MethodTemplate.cshtml
jhendrixMSFT: Context will always be the first param to all APIs.

src/Templates/ModelsTemplate.cshtml
jhendrixMSFT: Good catch will fix.

jhendrixMSFT:
@mcardosos For ARM it's likely that the SDK wouldn't even work as I haven't built/tested it.  There will be a separate push to make the v3 generator work for ARM.

olydis:
> win32-x64 restart

PR #42, opened 2017-10-31
Do not generate GroupClients!!!

mcardosos:
Lets get rid of this weird client naming. Also, using the go specific code namer everywhere.
When regenerating the tests, there is now an interesting side effect with models, I am not sure why. Maybe @fearthecowboy knows what is happening there :) This side effect does not introduce changes in the SDK, only in the tests.


marstr:
@mcardosos and I chatted about this offline today. We'd like to hold off a little while longer until we feel good about the name we select.

vladbarosan:
linux-x64 restart

marstr:
Alrighty, did we resolve the question of whether this applies to only anonymous Operation Groups or when it is when there would be stuttering?

mcardosos:
Both cases, if it is stuttering, in Go's case it is the same as if it was anonymous.

marstr:
And have we decided whether or not there are any instances of anonymous groups in our Azure Swaggers?

mcardosos:
There are operations without groups

mcardosos:
Added issue to swagger repo
https://github.com/Azure/azure-rest-api-specs/issues/1964

marstr:
Anonymous operations are today categorized as a "management" client, which solves the particular problem. That means this PR isn't impacted by that issue. It does mean, however, that we can't handle an operation group named "management". We should open an issue to consider how we want to handle this, so that we can solve this problem before someone supplies a swagger that forces us to. :)

PR #39, opened 2017-10-27
Refactor codegen for long-running operations.

jhendrixMSFT:
LROs will now return a future allowing the caller to get status on the
operation and control polling.  The Go routines have also been removed.

PR #43, opened 2017-10-31
Use context pattern

vladbarosan:


src/Templates/MethodTemplate.cshtml
```diff
@@ -162,9 +161,7 @@ func (client @(Model.Owner)) @(Model.PreparerMethodName)(@(Model.MethodParameter
 </text>
 }
 
-    @(Model.IsLongRunningOperation()
-        ? "return preparer.Prepare(&http.Request{Cancel: cancel})"
-        : "return preparer.Prepare(&http.Request{})")
+    return preparer.Prepare((&http.Request{}).WithContext(ctx))
 }
```
mcardosos: I would add a check for nil contexts

vladbarosan: withContext will Panic if the context is `nil`. Is there any point in having an extra check and presumably to return an error ?

mcardosos: I just wonder if it is a good idea to set the context to `context.Background()` if the user is passing a nil context, just for it not to panic.

vladbarosan:
Not sure why the CI is failing today. seems to be for some breaking changes ? those where yesterday also but I don't remember failing. Only thing I did today was the resolve conflicts after @jhendrixMSFT 's changes

vladbarosan:
Ok it seems that this PR #39 made the 2 params in the generated test code to be on the same line. Not sure why is that since previously it was on separate lines and for me also it puts it on the same line. @jhendrixMSFT any ideas why that change ? also do we want to revert or stay the same ?

olydis:
> win32-x64 restart

vladbarosan:
> win32-x64 restart

PR #44, opened 2017-10-31
Revert "Refactor codegen for long-running operations. (#39)"

marstr:
This reverts commit efe21ddebd63e15e5589f4b25a03ff9eb9168c9b.

This change should only be in the `breaking` branch, not `master` yet.

PR #45, opened 2017-10-31
Adding parameter `user-agent`

marstr:
Adding this parameter will allow for the person generating a package to
have a User Agent other than `Azure-SDK-for-Go/{Version}
arm-{Namespace}/{ApiVersion}`.

However, by default the behavior will still the same to prevent breaking
other tools.

### DIFF
--- a/src/Model/CodeModelGo.cs
+++ b/src/Model/CodeModelGo.cs
@@ -23,18 +23,21 @@ namespace AutoRest.Go.Model
 
         public string UserAgent
         {
-            get {
+            get
+            {
                 if (SpecifiedUserAgent == null) {
                     return DefaultUserAgent;
                 }
                 return SpecifiedUserAgent;
             }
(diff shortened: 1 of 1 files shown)


olydis:
~~~ Haskell
~~~

marstr:
I'm not certain, but I believe CI is failing because of conflicts that will be resolved once #44 is merged.

olydis:
~~~ Haskell
~~~

src/Model/CodeModelGo.cs
```diff
@@ -23,17 +23,32 @@ public class CodeModelGo : CodeModel
 
         public string UserAgent
         {
-            get
-            {
+            get {
```
mcardosos: `{` nit

marstr: Fixed.

PR #47, opened 2017-11-02
Make future.Result() work for all LROs.

jhendrixMSFT:
LROs that use the Azure-AsyncOperation header require a final GET after
the LRO has completed.  The generated Result() method has been updated
to make the GET call depending on the type of polling operation.


jhendrixMSFT:
Example: https://github.com/jhendrixMSFT/azure-sdk-for-go/commit/4d74982a1906513fb1d1c172a04a8b582af6789f

PR #49, opened 2017-11-03
Tiny import bug

mcardosos:
Shows when generating from the `visualstudio` swagger.


jhendrixMSFT:
How did this not fail CI?

jhendrixMSFT:
Ah this wasn't caught because the LRO tests aren't being run.  :(  Can you please update https://github.com/Azure/autorest.go/blob/master/test/src/tests/runner.go to include them in this PR?

mcardosos:
@jhendrixMSFT There is a test that is not passing :(

mcardosos:
Fixed test with this...
https://github.com/Azure/autorest.testserver/blob/c2e517f71d6a18a8ae5ec90aa2224a2354090bb8/routes/lros.js#L807

PR #50, opened 2017-11-06
Do not gen empty const block

mcardosos:
PTAL
@marstr @jhendrixMSFT @vladbarosan


olydis:
~~~ Haskell
~~~

PR #51, opened 2017-11-07
Discriminator

mcardosos:
Fixing tiny little bugs regarding the discriminator extension:

- Enum values are no longer duplicated on the SDK.
- Arrays of polymorphic structs are correctly unmarshalled.
- Json package is imported anytime there are polymorphic structs.
- Also deleted loggers because they are noisy and, in my opinion, they don't provide the most interesting info.

PTAL @jhendrixMSFT @marstr  @vladbarosan

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -327,7 +327,7 @@ namespace AutoRest.Go
         public static bool ShouldBeSyntheticType(this IModelType type)
         {            
             return (type is PrimaryType || type is SequenceType || type is DictionaryType || type is EnumType || 
-                (type is CompositeTypeGo && (type as CompositeTypeGo).IsPolymorphicResponse()));
+                (type is CompositeType && (type as CompositeTypeGo).IsPolymorphicResponse()));
         }
 
         /// <summary>
--- a/src/TransformerGo.cs
+++ b/src/TransformerGo.cs
@@ -79,7 +79,7 @@ namespace AutoRest.Go
             {
                 if (mt.IsPolymorphic)
                 {
-                    var values  = new List<EnumValue>();
+                    var values = new List<EnumValue>();
                     foreach (var dt in (mt as CompositeTypeGo).DerivedTypes)
                     {
                         var ev = new EnumValue();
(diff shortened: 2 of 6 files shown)


src/Extensions.cs
```diff
@@ -327,7 +327,7 @@ public static void AddImports(this IModelType type, HashSet<string> imports)
         public static bool ShouldBeSyntheticType(this IModelType type)
         {            
             return (type is PrimaryType || type is SequenceType || type is DictionaryType || type is EnumType || 
-                (type is CompositeTypeGo && (type as CompositeTypeGo).IsPolymorphicResponse()));
+                (type is CompositeType && (type as CompositeTypeGo).IsPolymorphicResponse()));
```
jhendrixMSFT: How do we end up with an object that isn't a CompositeTypeGo?  That seems like a bug.

mcardosos: Well, this func is called on CompositeType when it has not been casted to a CompositeTypeGo.

src/TransformerGo.cs
```diff
@@ -146,13 +146,27 @@ private void TransformEnumTypes(CodeModelGo cmg)
             {
                 if (em.Values.Where(v => topLevelNames.Contains(v.Name) || CodeNamerGo.Instance.UserDefinedNames.Contains(v.Name)).Any())
                 {
-                    em.HasUniqueNames = false;
+                    foreach (var v in em.Values)
+                    {
+                        v.Name = em.Name + v.Name;
```
jhendrixMSFT: Presumably this introduces some breaking changes yes?

mcardosos: No, I am just moving this lines from the `EnumTypeGo` file.

jhendrixMSFT: Ah I see now, I like this better.  :)

src/TransformerGo.cs
```diff
@@ -146,13 +146,27 @@ private void TransformEnumTypes(CodeModelGo cmg)
...
+                    }
                 }
                 else
                 {
                     topLevelNames.UnionWith(em.Values.Select(ev => ev.Name));
                 }
             }
+
+            var modelList = new List<EnumValue>();
```
jhendrixMSFT: Seems like modelList just evaporates after the following foreach block, why is it needed?

mcardosos: It is needed in the `GetUnique` func

jhendrixMSFT: Ugh I missed it, makes sense.

PR #52, opened 2017-11-11
Rename dataplane clients

mcardosos:
Detail noticed by @devigned  !


marstr:
@devigned, what are your opinions on the name "DataplaneClient" in lieu of "ManagementClient" where the latter is inappropriate?

devigned:
@marstr is `Client` inappropriate?

marstr:
@mcardosos, what are your thoughts?

mcardosos:
There are some cases in which the client for certain operation groups will just be named `Client`, so I would prefer not to use that name. What do you think about `ServiceClient`?

devigned:
@mcardosos can you give an example of where this occurs?

mcardosos:
[Example](https://github.com/Azure/azure-sdk-for-go/blob/master/arm/mediaservices/mediaservice.go#L28)

marstr:
Keep in mind, whatever you pick could conflict with names of operation group. Have you checked to see if there are any Operation Groups named Service?

devigned:
Perhaps, `Group` should be appended to a group if it is named `Client`. Seems like a misleading group name.

mcardosos:
No, a client named `GroupClient` is misleading and [customers have disliked it](https://github.com/Azure/azure-sdk-for-go/issues/672). Operation groups is a swagger concept that should not leak to the SDK. `OperationsClient` is less misleading in this case.

mcardosos:
Yes, there are `ServiceClient`s somewhere in the SDK :(

marstr:
> @mcardosos says:
Operation groups is a swagger concept that should not leak to the SDK.

I'm not sure that one can disentangle operation groups from what we generate. At the moment that's the entire boundary we use for picking what to associate with each client.

mcardosos:
Let me choose other words.
Currently in the compute package, we have `VirtualMachinesClient`, a client which's operations have to do with `virtual machines`.
If we had a `GroupClient`, its operations would have to do with a `group`. A group of what? In my  opinion, it does not make sense from the SDK perspective.

devigned:
With `http` you do something along these lines.
```go
client := &http.Client{
}
```
What you are trying to avoid is `ClientClient`, which would be ugly.

How about `HttpClient` rather than `ManagementClient` or `DataplaneClient`? 

If someone makes an operation group called `HttpClient`, then shame on them and us.

mcardosos:
It could be, and works in all cases

mcardosos:
What do you think about `BaseClient`?

mcardosos:
Agreed offline on `BaseClient`

PR #72, opened 2018-01-03
Emit correct code for optional body param stream.

jhendrixMSFT:
Don't assume that an optional body param is a JSON object as it could be
a binary stream.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/671

PR #73, opened 2018-01-05
Don't treat constants as parameters.

jhendrixMSFT:
Per swagger spec, required enums with one value are constants thus they
shouldn't be paramaters, so just emit the value directly in the code.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/905.

jhendrixMSFT:
Ran CI locally and it passed so merging.

PR #74, opened 2018-01-05
Fix linter issues

vladbarosan:
- Fix improper comment generation
- Fix enum stuttering 
- Fix future type stuttering

### DIFF
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -26,15 +26,15 @@ namespace AutoRest.Go.Model
 
         private CompositeTypeGo _rootType;
 
-        public CompositeTypeGo() : this("")
+        public CompositeTypeGo()
         {
         }
 
         public CompositeTypeGo(string name) : base(name)
         {
-            if (string.IsNullOrWhiteSpace(this.Documentation))
+            if (string.IsNullOrWhiteSpace(Documentation))
             {
-                this.Documentation = "...";
+                Documentation = "...";
             }
         }
 
@@ -45,9 +45,9 @@ namespace AutoRest.Go.Model
                 throw new ArgumentException("{0} is not a valid type for SyntheticType", wrappedType.ToString());
             }
 
-            if (string.IsNullOrWhiteSpace(this.Documentation))
+            if (string.IsNullOrWhiteSpace(Documentation))
             {
-                this.Documentation = "...";
+                Documentation = "...";
             }
 
             // gosdk: Ensure the generated name does not collide with existing type names
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -28,6 +28,10 @@ namespace AutoRest.Go.Model
 
         public CompositeTypeGo()
         {
+            if (string.IsNullOrEmpty(this.Documentation))
+            {
+                this.Documentation = "...";
+            }
         }
 
         public CompositeTypeGo(string name) : base(name)
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -26,16 +26,16 @@ namespace AutoRest.Go.Model
 
         private CompositeTypeGo _rootType;
 
-        public CompositeTypeGo()
+        public CompositeTypeGo() : this("")
         {
-            if (string.IsNullOrEmpty(this.Documentation))
-            {
-                this.Documentation = "...";
-            }
         }
 
         public CompositeTypeGo(string name) : base(name)
         {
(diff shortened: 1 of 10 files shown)


vladbarosan:
addresses https://github.com/Azure/azure-sdk-for-go/issues/941

src/Templates/ModelTemplate.cshtml
```diff
@@ -322,69 +315,67 @@ else
...
+    return !page.@(modelPageType.ResultFieldName).IsEmpty()
     }
 
     // Response returns the raw server response from the last page request.
-    func (page @Model.Name) Response() @pageType.ContentType.Name {
-    return page.@(pageType.ResultFieldName)
-    }
+    func (page @Model.Name) Response() @modelPageType.ContentType.Name {
+    return page.@modelPageType.ResultFieldName
+        }
```
jhendrixMSFT: Nit, indentation.

src/Model/CompositeTypeGo.cs
```diff
@@ -26,12 +26,16 @@ public class CompositeTypeGo : CompositeType
 
         private CompositeTypeGo _rootType;
 
-        public CompositeTypeGo()
+        public CompositeTypeGo() : this("")
```
jhendrixMSFT: Does this have any unintended side-effects?  Probably not but would be good to regenerate the entire SDK as part of validating this PR (perhaps you already did that).

src/Model/PropertyGo.cs
```diff
@@ -11,7 +11,20 @@ public class PropertyGo : Property
     {
         public PropertyGo()
         {
-
+            Name.OnGet += v =>
```
jhendrixMSFT: How does this fix anything, is it due to some ordering issue during transformation?

src/Model/CompositeTypeGo.cs
vladbarosan: yes I regenerated all is good

src/Model/PropertyGo.cs
vladbarosan: @jhendrixMSFT short answer yes. We are doing some transformation on the composite type transformer where we set the property name to be the type name for flattened types ( which will be embedded types for us ).  When generating the whole sdk hat produced issues because later we were changing the type name. 

I removed that part here and move this check here.

src/TransformerGo.cs
```diff
@@ -297,32 +284,32 @@ private void FixStutteringTypeNames(CodeModelGo cmg)
...
                         {
-                            (exported as CompositeType).Name.FixedValue = CodeNamerGo.AttachTypeName(name, cmg.Namespace, nameInUse, "Type");
+                            compositeType.Name.Value = CodeNamerGo.AttachTypeName(name, cmg.Namespace, nameInUse, "Type");
                         }
-                        else if (exported is Method)
+                        else if (exported is Method methodType)
                         {
-                            (exported as Method).Name.FixedValue = CodeNamerGo.AttachTypeName(name, cmg.Namespace, nameInUse, "Method");
+                            methodType.Name.Value = CodeNamerGo.AttachTypeName(name, cmg.Namespace, nameInUse, "Method");
                         }
```
mcardosos: Have you tried regenerating the SDK? the last time I touched this the generation changed a little

vladbarosan: Yes , see first comment, all looks good.

test/src/tests/generated/lro/models.go
```diff
@@ -2650,7 +2650,7 @@ func (sp *SubProduct) UnmarshalJSON(body []byte) error {
 type SubProductProperties struct {
 	ProvisioningState *string `json:"provisioningState,omitempty"`
 	// ProvisioningStateValues - Possible values include: 'ProvisioningStateValues1Succeeded', 'ProvisioningStateValues1Failed', 'ProvisioningStateValues1Canceled', 'ProvisioningStateValues1Accepted', 'ProvisioningStateValues1Creating', 'ProvisioningStateValues1Created', 'ProvisioningStateValues1Updating', 'ProvisioningStateValues1Updated', 'ProvisioningStateValues1Deleting', 'ProvisioningStateValues1Deleted', 'ProvisioningStateValues1OK'
-	ProvisioningStateValues ProvisioningStateValues `json:"provisioningStateValues,omitempty"`
+	ProvisioningStateValues ProvisioningStateValues1 `json:"provisioningStateValues,omitempty"`
```
mcardosos: >ProvisioningStateValues1 [](start = 25, length = 24)

Little side effect which is not that cool

vladbarosan: previous value was actually incorrect, this is the correct value if you look at the comment of the field

PR #98, opened 2018-02-13
Fix discriminator property and code model self dependency

vladbarosan:
- Move property name transforms out of code model construction
- Fix Discriminator Property naming


olydis:
~~~ Haskell
~~~

PR #114, opened 2018-03-28
Don't treat constants as parameters.

jhendrixMSFT:
Per swagger spec, required enums with one value are constants thus they
shouldn't be paramaters, so just emit the value directly in the code.
Change was ported from the v2 generator.


olydis:
~~~ Haskell
~~~

PR #115, opened 2018-03-29
Fix validation code for required properties.

jhendrixMSFT:
The v3 generator does not emit required properties as pointer types
however the validation codegen wasn't taking that into account (it
assumed that all struct fields were byrefs).  Removed the OR clause that
was causing the validation codegen to always assume fields are byref.


olydis:
~~~ Haskell
~~~

PR #126, opened 2018-05-07
Add Cron Job execution mode, where Specs are built.

marstr:
Fixes Azure/azure-sdk-for-go#1052

### DIFF
--- a/testSpecs.sh
+++ b/testSpecs.sh
@@ -14,8 +14,8 @@ set -ev
 
 git clone --depth 1 --branch $SPECS_REF -- https://github.com/Azure/azure-rest-api-specs.git $SPECS_LOCATION
 
-for f in $(find $SPECS_LOCATION -iname README.md | grep -v node_modules); do 
-    autorest --go $f
+for f in $(find $SPECS_LOCATION -iname README.md | grep specification); do 
+    autorest --go --multiapi $f
 done
 
 rm -rf $SPECS_LOCATION
--- a/testSpecs.sh
+++ b/testSpecs.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+SPECS_LOCATION=$(mktemp -d specs.XXXX)
+
+if test -z "$1"; then
+    SPECS_REF="master"
+else
+    SPECS_REF="$1"
+fi
+
+echo "travis_fold:start:generate_specs@$SPECS_REF"
+
+set -ev
+
+git clone --depth 1 --branch $SPECS_REF -- https://github.com/Azure/azure-rest-api-specs.git $SPECS_LOCATION
+
+for f in $(find $SPECS_LOCATION -iname README.md | grep -v node_modules); do 
+    autorest --go $f
+done
+
+rm -rf $SPECS_LOCATION
+
+echo "travis_fold:end:generate_specs@$SPECS_REF"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ go:
 matrix:
   allow_failures:
     - go: master
+  fast_finish: true
 env:
   - NODE_VERSION="8.9.4" DEP_VERSION="0.4.1"
 before_install:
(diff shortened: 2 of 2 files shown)


marstr:
Travis configuration changes were also made to satisfy the issue linked above. Those with permissions can find the updated configuration here: https://travis-ci.org/Azure/autorest.go/settings (Note, the cron-job)

Relevant documentation can be found here: https://docs.travis-ci.com/user/cron-jobs/

vladbarosan:
Gareth sent an email a while ago. Since we didn't have much to do on the generator since then that might have gotten unnoticed. @fearthecowboy should we go forward only with the VSTS jobs ?

marstr:
My understanding was that AutoRest core had moved to using VSTS. I didn't realize that the Go generator would have also moved away from Travis. Is this a hard requirement?

fearthecowboy:
The plugins were also moved, since we didn't want to self-host any of that.

Ideally, I'd recommend that you get off travis, and go to VSTS for all builds :D .. the Linux pool is 150 machines -- it's so much faster to activate.

vladbarosan:
@fearthecowboy thanks for the info, right now we didn't have a lot of activity on this repo. Going forward, while we transition, do we loose any telemetry or anything if we use PR's from forks (i.e. we don't trigger the VSTS jobs )?

fearthecowboy:
VSTS doesn't support PRs from Forks at the moment; you have to do it from branches. I've been told this is temporary, they should be bringing it back soon

marstr:
Chatting offline a while ago with @fearthecowboy, we agreed that we'll move over to VSTS for these jobs once they support running on PRs from forks. Until then, I think we should proceed with Travis. (Even if that is extremely short-lived.)

@vladbarosan, @jhendrixMSFT mind reviewing?

vladbarosan:
@marstr  I was thinking actually and :
1.  this is a cron job, it doesnt have anything to do with PR's so it's not really dependent on that.
2.  right and we already have the publishing in VSTS ( i.e. we have to make PR's from a branch if autorest is to be published ) so right now its too late.

vladbarosan:
@marstr can you post a Travis run to make sure is all good . (you can enable travis for your branch temporarily  ) ?

testSpecs.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+echo "travis_fold:start:generate_specs@$SPECS_REF"
+
+set -ev
+
+git clone --depth 1 --branch $SPECS_REF -- https://github.com/Azure/azure-rest-api-specs.git $SPECS_LOCATION
+
+for f in $(find $SPECS_LOCATION -iname README.md | grep -v node_modules); do 
+    autorest --go $f
```
vladbarosan: is autorest available as a global package ? it seems is getting pulled as a local dependency

marstr:
Merged with [master in my fork](https://github.com/marstr/autorest.go/tree/master) to run the test. You can monitor test results here:
https://travis-ci.org/marstr/autorest.go/builds

testSpecs.sh
marstr: Good catch, fixed in my local.

marstr:
Seems to be working now, @vladbarosan. As seen here: https://travis-ci.org/marstr/autorest.go/jobs/379915633

testSpecs.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "travis_fold:start:generate_specs@$SPECS_REF"
+
+set -ev
+
+git clone --depth 1 --branch $SPECS_REF -- https://github.com/Azure/azure-rest-api-specs.git $SPECS_LOCATION
+
+for f in $(find $SPECS_LOCATION -iname README.md | grep specification); do 
+    autorest --go --multiapi $f
+done
+
```
vladbarosan: can we do a go build after the generation to make sure the generated code builds ?

testSpecs.sh
```diff
@@ -0,0 +1,23 @@
...
+rm -rf $SPECS_LOCATION
```
vladbarosan: also I see some access denied in the build logs not sure if its actually building

marstr:
We're past the amount of time I want to spend on this task :p I'm going to close this PR out.

PR #127, opened 2018-05-17
Fix XML unmarshalling

jhendrixMSFT:
Avoid type name collisions for XML wrapper types.
Remove UTF8 BOM from byte slice before unmarshalling.
Remove unnecessary ResponseError interface.

### DIFF
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -75,15 +75,21 @@ namespace AutoRest.Go.Model
                 throw new ArgumentException("{0} is not a valid type for SyntheticType", wrappedType.ToString());
             }
 
-            // gosdk: Ensure the generated name does not collide with existing type names
+            CodeModel = responseToWrap.CodeModel;
             BaseType = wrappedType;
 
             if (wrappedType.XmlIsWrapped)
             {
-                Name = wrappedType.XmlName;
+                // ensure that the name won't collide with an existing type.
+                // if it does then we'll give it the standard response name.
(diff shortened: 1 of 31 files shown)


src/Model/CompositeTypeGo.cs
```diff
@@ -75,15 +75,21 @@ public CompositeTypeGo(MethodGo responseToWrap)
...
+                if (!CodeModel.ModelTypes.Where(mt => (mt.Name == wrappedType.XmlName)).Any())
+                {
+                    Name = wrappedType.XmlName;
+                }
                 XmlProperties = wrappedType.XmlProperties;
             }
-            else
+            
+            if (Name == null)
             {
```
vladbarosan: can this null Name can come only from the XmlName there ? If so maybe we can do a `wrappedType.XmlName ?? ` $"{responseToWrap.Name}Response"` there

jhendrixMSFT: The null happens if we find an existing type that has the same name as ```wrappedType.XmlName``` (see line 85).

src/Model/CompositeTypeGo.cs
```diff
@@ -203,7 +209,7 @@ public void AddImports(HashSet<string> imports)
         public string AddHTTPResponse()
         {
             return (IsResponseType || IsWrapperType) ?
-                "autorest.Response `json:\"-\"`\n" :
+                $"autorest.Response `{CodeModel.ToCodeModelGo().Encoding}:\"-\"`\n" :
                 null;
```
vladbarosan: I presume this was an error , hmm how come it didn't affect us  so far in Xml?

PR #128, opened 2018-05-24
Better field name for wrapped arrays.

jhendrixMSFT:
For the wrapper type around an array name the field 'Items'.
Added ResponseError interface back as it's needed for tests.
Fixed array tests, the ones for time types have been disabled.


src/Templates/MethodTemplate.cshtml
```diff
@@ -170,7 +170,7 @@ func (client @(Model.MethodGroup.ClientName)) @(opIdCamelCase)Responder(resp pip
     var unmarshalInto = "result";
     if (Model.ReturnValueIsWrapperType() && !Model.ReturnValueIsXmlWrapped())
```
vladbarosan: why the `&& !Model.ReturnValueIsXmlWrapped()`?

src/Model/MethodGo.cs
```diff
@@ -490,6 +490,18 @@ public bool ReturnValueIsWrapperType()
             return HasReturnValue() && ReturnValue().Body is CompositeTypeGo && ReturnValue().Body.Cast<CompositeTypeGo>().IsWrapperType;
         }
 
+        /// <summary>
+        /// Returns the name of the wrapped field.  If the return value isn't a wrapper type it returns the empty string.
+        /// </summary>
+        public string ReturnValueWrapperTypeFieldName()
```
vladbarosan: are all of these methods coming from an interface ? because I see they are all functions instead of properties, was wondering why

jhendrixMSFT: No interface.  I wrote this as a method since it calls other methods.  Generally speaking I use methods instead of properties if computing the value is non-trivial (for some definition of non-trivial).  Open to suggestions, we should probably discuss guidelines like this in one of our T/Th sync-ups.

vladbarosan: Ok, its fine. I mean if we are discussing guidlines I had a bigger proposal to get rid of all the type assertions and conditional logic we are doing on it  but not if we will ever get to that :)

PR #129, opened 2018-05-29
Update LRO codegen to use latest go-autorest features

jhendrixMSFT:
Codegen for LROs has been updated to use azure.NewFutureFromResponse
when creating a Future.
The Result method on generated future types uses Future's new internal
logic for retrieving the final result.
Added missing LRO tests and updated dependencies.

### DIFF
--- a/src/Templates/ModelTemplate.cshtml
+++ b/src/Templates/ModelTemplate.cshtml
@@ -446,10 +446,15 @@ else
     }
     @if (ftg.ResultTypeName == MethodGo.DefaultReturnType)
     {
+        // for default return types (i.e. autorest.Response)
+        // assign the raw HTTP response to the *http.Response field
     @:@(resultVar).Response = future.Response()
     }
     else
     {
+        // modeled return types embed autorest.Resposne as an anonymous field
+        // so in order to assign the raw HTTP response to the *http.Response field
+        // in it we need an extra ".Response" :(
     <text>
     sender := autorest.DecorateSender(client, autorest.DoRetryForStatusCodes(client.RetryAttempts, client.RetryDuration, autorest.StatusCodesForRetry...))
     if @(resultVar).Response.Response, err = future.GetResult(sender); err == nil && @(resultVar).Response.Response.StatusCode != http.StatusNoContent {
--- a/src/Templates/ModelTemplate.cshtml
+++ b/src/Templates/ModelTemplate.cshtml
@@ -426,14 +426,14 @@ else
     var resultVar = ftg.ResultTypeName.ToShortName();
     var resultVarTarget = resultVar;
     var futureTypeName = $"{Model.CodeModel.Namespace}.{Model.Name}";
-    if (ftg.ResultType is IteratorTypeGo)
+    if (ftg.ResultType is IteratorTypeGo itg)
     {
-        resultVarTarget = $"{resultVarTarget}.{ftg.ResultType.Cast<IteratorTypeGo>().PageField}";
+        resultVarTarget = $"{resultVarTarget}.{itg.PageField}";
     }
     <text>
     // Result returns the result of the asynchronous operation.
     // If the operation has not completed it will return an error.
-    func (future @Model.Name) Result(client @ftg.ClientTypeName) (@resultVar @ftg.ResultTypeName, err error) {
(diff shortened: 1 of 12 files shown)


src/Templates/ModelTemplate.cshtml
```diff
@@ -426,51 +426,41 @@ else
...
     }
-    var req *http.Request
-    var resp *http.Response
-    if future.PollingURL() != "" {
-        req, err = http.NewRequest(http.MethodGet, future.PollingURL(), nil)
+    @if (ftg.ResultTypeName == MethodGo.DefaultReturnType)
+    {
+    @:@(resultVar).Response = future.Response()
+    }
+    else
```
vladbarosan: is this for that case where you need to do an extra get at the end ? if so, what is `ResultTypeName` then ?

jhendrixMSFT: The default return type is ```autorest.Response``` so if that's all that's being returned then we assign the raw HTTP response to the ```Response``` field.

vladbarosan: Yes, I guess the only think im saying is that maybe its not obvious why if the response is not the default return type we need to have the extra getResult on the future.

 Nit: Maybe a comment or a helper method.

---
In reply to: [191599956](https://github.com/Azure/autorest.go/pull/129#discussion_r191599956) [](ancestors = 191599956)

test/src/tests/acceptancetests/lrogrouptest/lro_test.go
```diff
@@ -6,10 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Azure/go-autorest/autorest/to"
 	chk "gopkg.in/check.v1"
 
 	"tests/acceptancetests/utils"
-	. "tests/generated/lro"
```
vladbarosan: >.  [](start = 1, length = 2)

i thought we run go fmt part of CI and check if the pushed content is the same as the resulting one ?

test/src/tests/acceptancetests/lrogrouptest/lro_test.go
```diff
@@ -23,72 +24,638 @@ var lrosClient = getLROsClient()
...
+
+func (s *LROSuite) TestSADsPutNonRetry400(c *chk.C) {
+	_, err := lroSADSClient.PutNonRetry400(context.Background(), &lrogroup.Product{})
+	c.Assert(err, chk.NotNil)
+}
+
+// custom header client
+
+func (s *LROSuite) TestCustomHeaderPost202Retry200(c *chk.C) {
+	/*req, err := lroCustomHeaderClient.Post202Retry200Preparer(context.Background(), &lrogroup.Product{})
```
vladbarosan: is there a problem with this one ? ( and the next ones) if so maybe create an issue for them?

test/src/tests/acceptancetests/lrogrouptest/lro_test.go
jhendrixMSFT: Importing a package with a preceding "dot" means to import its content into the importing package's namespace.  Unfortunately this breaks stuff in vscode (e.g. right-clicking on a func to jump to its definition doesn't work) so I removed the "dot" from the import and reference the package content explicitly.

vladbarosan: Oh my bad, i thought it was a whitespace ( since whitespaces are rendered as dots in CodeFlow ) :)

src/Templates/ModelTemplate.cshtml
jhendrixMSFT: Added a comment.

test/src/tests/acceptancetests/lrogrouptest/lro_test.go
jhendrixMSFT: Thanks I forgot about this, will update the comment and open an issue.

jhendrixMSFT: Opened issue https://github.com/Azure/azure-sdk-for-go/issues/1922 to track adding custom header data.

vladbarosan:
Feel free to merge 👍

PR #130, opened 2018-05-31
Fix future/pageable types.

jhendrixMSFT:
The previous LRO changes were incomplete, they missed the case where
futures intersect with pageable types.
Avoid importing the net/http package when it's not needed.

### DIFF
--- a/src/Model/CodeModelGo.cs
+++ b/src/Model/CodeModelGo.cs
@@ -142,13 +142,9 @@ namespace AutoRest.Go.Model
                 {
                     imports.Add(PrimaryTypeGo.GetImportLine("github.com/Azure/go-autorest/autorest/azure"));
                     // if any of the futures return the non-default response then we need the net/http package
-                    foreach (var ftg in futureModels)
+                    if (futureModels.Any(fm => !fm.IsDefaultReturnType))
                     {
-                        if (!ftg.IsDefaultReturnType)
-                        {
-                            imports.Add(PrimaryTypeGo.GetImportLine("net/http"));
-                            break;
-                        }
+                        imports.Add(PrimaryTypeGo.GetImportLine("net/http"));
                     }
                 }
                 ModelTypes.Cast<CompositeTypeGo>()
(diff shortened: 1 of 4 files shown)


src/Model/CodeModelGo.cs
```diff
@@ -137,10 +137,19 @@ public IEnumerable<string> ModelImports
                 {
                     imports.Add(PrimaryTypeGo.GetImportLine("github.com/Azure/go-autorest/autorest"));
                 }
-                if (ModelTypes.Any(mt => mt is FutureTypeGo))
+                var futureModels = ModelTypes.Where(mt => mt is FutureTypeGo).Cast<FutureTypeGo>();
+                if (futureModels.Any())
```
vladbarosan: not sure what was wrong with the previous if, also why do we cast to FutureTypeGo

vladbarosan: oh ok nevermind we need the collection later, but do we need the cast ?

---
In reply to: [192249868](https://github.com/Azure/autorest.go/pull/130#discussion_r192249868) [](ancestors = 192249868)

jhendrixMSFT: Yes, the ```IsDefaultReturnType``` property is only on ```FutureTypeGo```.

src/Model/CodeModelGo.cs
```diff
@@ -137,10 +137,19 @@ public IEnumerable<string> ModelImports
...
                 }
-                if (ModelTypes.Any(mt => mt is FutureTypeGo))
+                var futureModels = ModelTypes.Where(mt => mt is FutureTypeGo).Cast<FutureTypeGo>();
+                if (futureModels.Any())
                 {
                     imports.Add(PrimaryTypeGo.GetImportLine("github.com/Azure/go-autorest/autorest/azure"));
-                    imports.Add(PrimaryTypeGo.GetImportLine("net/http"));
+                    // if any of the futures return the non-default response then we need the net/http package
+                    foreach (var ftg in futureModels)
+                    {
```
vladbarosan: maybe ```if(futureModels.Any( fm => !fm.IsDefaultReturnType)) {}```

PR #134, opened 2018-06-15
Emit InclusiveMaximum constraint as 64-bit type

jhendrixMSFT:
Numeric constrains that are 2^32 will overflow int when compiling on
32-bit platforms.  For the InclusiveMaximum constraint emit the constant
as an int64 to avoid the overflow.  The other max constraints are
limited to int so they were not changed.


jhendrixMSFT:
Part of the fix for https://github.com/Azure/azure-sdk-for-go/issues/1993

test/src/tests/generated/lro/models.go
```diff
@@ -1549,6 +1549,64 @@ func (future *LROsPostAsyncRetrySucceededFuture) Result(client LROsClient) (p Pr
 	return
 }
 
+// LROsPostDoubleHeadersFinalAzureHeaderGetFuture an abstraction for monitoring and retrieving the results of a
+// long-running operation.
+type LROsPostDoubleHeadersFinalAzureHeaderGetFuture struct {
```
vladbarosan: >LROsPostDoubleHeadersFinalAzureHeaderGetFuture [](start = 5, length = 46)

are these coming from new tests in the test server ?

vladbarosan: in that case is there any extra test we need to add ?

---
In reply to: [196148727](https://github.com/Azure/autorest.go/pull/134#discussion_r196148727) [](ancestors = 196148727)

PR #137, opened 2018-06-22
Add doc comment for base64url formatted properties.

jhendrixMSFT:
Current codegen emits base64url formatted types as strings so it's not
clear that the contents should be URL/base64 encoded.  Until we have
proper support emit an extra comment stating the string's format.


jhendrixMSFT:
Added to mitigate https://github.com/Azure/azure-sdk-for-go/issues/2111 util we have proper support.

PR #139, opened 2018-07-09
Fix XML tag generation.

jhendrixMSFT:
Don't use the SerializedName for polymorphic types.
Use the XML element type name for wrapped types.

### DIFF
--- a/src/Model/PropertyGo.cs
+++ b/src/Model/PropertyGo.cs
@@ -32,15 +32,7 @@ namespace AutoRest.Go.Model
             bool hasParent = false;
             if (Parent is CompositeTypeGo go && !go.IsWrapperType)
             {
-                // This extra check is necessary because of https://github.com/Azure/autorest/issues/2889
-                if (this.IsPolymorphicDiscriminator)
-                {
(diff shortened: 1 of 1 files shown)


src/Model/PropertyGo.cs
```diff
@@ -32,15 +32,7 @@ public string Tag(bool omitEmpty = true)
...
-                // This extra check is necessary because of https://github.com/Azure/autorest/issues/2889
-                if (this.IsPolymorphicDiscriminator)
-                {
-                    sb.Append(SerializedName);
-                }
-                else
-                {
-                    sb.Append(XmlName);
-                }
+                sb.Append(XmlName);
```
vladbarosan: is this not producing issues anymore ? I think for storage files sdk it was showing up.

jhendrixMSFT: I rebuilt blob, files and queues with this change and it worked as expected (i.e. no crash).

vladbarosan: ok if `go build` on them was fine also, I think its good

PR #140, opened 2018-07-12
Ensure client parameters match their definition

jhendrixMSFT:
When generating client property names in preparers ensure that the
calculated client property name matches the actual property name.

### DIFF
--- a/src/Model/ParameterGo.cs
+++ b/src/Model/ParameterGo.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace AutoRest.Go.Model
 {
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2174

src/Model/ParameterGo.cs
```diff
@@ -156,7 +156,32 @@ public string ValueForMap(bool useDefault = false)
...
+                bool found = false;
+                foreach (var clientProp in Method.CodeModel.Properties)
+                {
+                    // prefer an exact match
+                    if (clientProp.Name == propName)
+                    {
+                        found = true;
+                        break;
+                    }
+                }
```
vladbarosan: hmm, a little more inefficient but i dont think impactful:
```
                var propName = CodeNamerGo.Instance.GetPropertyName(Name.Value);

                // verify that the calculated name matches the property name
                bool found = Method.CodeModel.Properties.Any(clientProp => clientProp.Name == propName);

                // didn't find an exact match, find the closest match.  we hit this case if
                // the front-end decided to add a suffix (e.g. '1') to the client property name.
                if (!found)
                {
                    propName = Method.CodeModel.Properties.First(clientProp => System.Text.RegularExpressions.Regex.IsMatch(clientProp.Name, $"{propName}\\d")).Name;
                }
                value = $"client.{propName}";
```
also using regex to make sure we don't match by mistake another property ( i.e. "Prop1" , "PropFoo" and we are searching for "Prop")

jhendrixMSFT: Good point on the regex, I've made this change.

PR #142, opened 2018-07-19
Ensure client parameters match their definition 2

jhendrixMSFT:
The original fix missed a few cases.
Updated tests from latest test swagger.

### DIFF
--- a/src/Model/ParameterGo.cs
+++ b/src/Model/ParameterGo.cs
@@ -52,7 +52,7 @@ namespace AutoRest.Go.Model
             }
             else if (IsClientProperty)
             {
-                retval = "client." + Name.Value.Capitalize();
+                retval = GetClientPropertryName();
             }
             else
             {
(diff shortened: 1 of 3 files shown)


src/Model/ParameterGo.cs
```diff
@@ -411,7 +419,7 @@ public static string Validate(this IEnumerable<ParameterGo> parameters, HttpMeth
 
                 var name = !p.IsClientProperty
                         ? p.Name.Value
-                        : "client." + p.Name.Value.Capitalize();
+                        : p.GetClientPropertryName();
```
vladbarosan: I presume this is the fix ?

jhendrixMSFT: This is for generating validation code.  The case that's still causing compile failures is the change on line 55.

vladbarosan: looks good, thanks!

PR #143, opened 2018-07-19
Fix another case of client parameter naming

jhendrixMSFT:
Nothing is impacted by this at present, just a preventative fix while we're familiar with the issue.

### DIFF
--- a/src/Model/MethodGo.cs
+++ b/src/Model/MethodGo.cs
@@ -396,7 +396,7 @@ namespace AutoRest.Go.Model
                         }
                         else if (param.IsClientProperty)
                         {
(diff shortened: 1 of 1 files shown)


src/Model/MethodGo.cs
```diff
@@ -396,7 +396,7 @@ public IEnumerable<string> PrepareDecorators
                         }
                         else if (param.IsClientProperty)
                         {
-                            value = $"client.{param.Name.ToPascalCase()}";
```
vladbarosan: can we do a search all for `client.` to see if there are any other cases

jhendrixMSFT: Yep that's how I found this one, I don't see any others.

PR #144, opened 2018-07-24
Fix pages that contain polymorphic types

vladbarosan:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2293

### DIFF
--- a/src/Templates/ModelTemplate.cshtml
+++ b/src/Templates/ModelTemplate.cshtml
@@ -344,7 +344,6 @@ else
 @if (Model is PageTypeGo modelPageType)
 {
     var itemName = modelPageType.ItemName;
-    var itemType = modelPageType.ElementType.Name;
     <text>
     @EmptyLine
     // Next advances to the next page of values.  If there was an error making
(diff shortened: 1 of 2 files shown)


vladbarosan:
Let's not merge until i regenerate the whole repo to check of any other changes

vladbarosan:
Also ideally we should rename the iterator and page types, it doesn't seem that straightforward, will give it another try.

src/Templates/ModelTemplate.cshtml
```diff
@@ -344,7 +344,7 @@ else
 @if (Model is PageTypeGo modelPageType)
 {
     var itemName = modelPageType.ItemName;
-    var itemType = modelPageType.ElementType.Name;
+    var itemType = modelPageType.ElementType.HasInterface() ? modelPageType.ElementType.GetInterfaceName() : modelPageType.ElementType.Name.Value;
```
jhendrixMSFT: I think this should be added to the PageTypeGo class as it basically duplicates line 32 in the ctor.

vladbarosan: thought it might be confusing to have both Name and type as properties in the Page class, but I named it ReturnTypeName, should be fine.

vladbarosan:
regenerated the whole sdk and no other changes from this ( which confirms we didnt have this scenario until now )

PR #145, opened 2018-07-25
Have the iterator return the correct polymorphic value

vladbarosan:
Have the iterator Value() call return the same type as the value in the Page


vladbarosan:
somehow this wasn't included in the previous PR :( . 
output here 
https://github.com/Azure/azure-sdk-for-go/pull/2298/commits/36a094c3762a12be87f7739249400f3cc62dfbfa

PR #146, opened 2018-08-23
Fix XML tag generation.

jhendrixMSFT:
For unwrapped arrays use the element's XML name.
If a composite type defines an XML name use that for property tag.


jhendrixMSFT:
The storage folks have verified the codegen works as expected.

PR #147, opened 2018-08-23
Centralize calls to panic

jhendrixMSFT:
Added validateError() that will panic on non-nil error.
If reading a response body fails don't wrap the error so the pipeline
will recognize the error and retry as appropriate.


src/Templates/ModelsTemplate.cshtml
```diff
@@ -108,6 +108,14 @@ func joinConst(s interface{}, sep string) string {
 	}
 	return strings.Join(ss, sep)
 }
+
+@EmptyLine
+func validateError(err error) {
+	if err != nil {
+		panic(err)
```
tombuildsstuff: out of interest why panic verses propagating the error?

vladbarosan: Looking through the usage, I guess maybe it makes sense when we do that casting between the types, not 100% about the others ones.

---
In reply to: [212587493](https://github.com/Azure/autorest.go/pull/147#discussion_r212587493) [](ancestors = 212587493)

jhendrixMSFT: The storage folks are still debating on how they want to handle this.

PR #153, opened 2018-09-19
Fix discriminator enum type name collision

jhendrixMSFT:
The fix-ups for discriminated types can create a new enum type in some
cases.  To avoid a name collision perform these fix-ups after the enum
and model types have had their names updated.
Normalize the line-endings in PageTypeGo.cs.


src/TransformerGo.cs
```diff
@@ -22,6 +22,9 @@ public override CodeModelGo TransformCodeModel(CodeModel cm)
             SwaggerExtensions.ProcessGlobalParameters(cmg);
             TransformEnumTypes(cmg);
             TransformModelTypes(cmg);
+            // FixUpPolymorphicTypes can generate new enum types so to avoid
```
marstr: [nit] I'd move this comment to be above the method declaration.

PR #154, opened 2018-10-09
Add preview-chk arg

jhendrixMSFT:
This is a temporary solution for verifying that a preview input swagger
has its codegen output directory under a preview subdirectory.

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -42,8 +42,8 @@ namespace AutoRest.Go
             {
                 const string previewSubdir = "/preview/";
                 var files = await Settings.Instance.Host.GetValue<string[]>("input-file");
-                // skip validation for composite builds as we don't have a well-defined model for mixed preview/stable swaggers
-                if (files.Length == 1 && files[0].IndexOf(previewSubdir) > -1 &&
+                // only evaluate composite builds if all swaggers are preview as we don't have a well-defined model for mixed preview/stable swaggers
+                if (files.All(file => file.IndexOf(previewSubdir) > -1) &&
                     Settings.Instance.Host.GetValue<string>("output-folder").Result.IndexOf(previewSubdir) == -1)
                 {
                     throw new InvalidOperationException($"codegen for preview swagger {files[0]} must be under a preview subdirectory");
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -36,6 +36,19 @@ namespace AutoRest.Go
         /// <returns></returns>
         public override async Task Generate(CodeModel cm)
         {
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Part of the fix for https://github.com/Azure/azure-sdk-for-go/issues/2907

src/CodeGeneratorGo.cs
```diff
@@ -36,6 +36,19 @@ public override string ImplementationFileExtension
...
         public override async Task Generate(CodeModel cm)
         {
+            // if preview-chk:true is specified verify that preview swagger is output under a preview subdirectory.
+            // this is a bit of a hack until we have proper support for this in the swagger->sdk bot so it's opt-in.
+            if (Settings.Instance.Host.GetValue<bool>("preview-chk").Result)
+            {
+                const string previewSubdir = "/preview/";
+                var files = await Settings.Instance.Host.GetValue<string[]>("input-file");
+                // skip validation for composite builds as we don't have a well-defined model for mixed preview/stable swaggers
+                if (files.Length == 1 && files[0].IndexOf(previewSubdir) > -1 &&
```
vladbarosan: I would say if anything is preview than it should be in preview ? but sure we can leave it out for now

jhendrixMSFT: I tend to agree, once we as a team can decide then we can update this.

vladbarosan: >files.Length ==  [](start = 20, length = 16)

why is it == 1 ? many swaggers are broken in multiple files. or are they unified by autorest when they reach us ? (e.g. network [https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/readme.md](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/network/resource-manager/readme.md) )

vladbarosan: I guess something like if(files.All( file =>  file.Substring(0, file.LastIndexOf("/") == files[0].Substring(0, file.LastIndexOf("/") ) && ...

---
In reply to: [223891666](https://github.com/Azure/autorest.go/pull/154#discussion_r223891666) [](ancestors = 223891666)

vladbarosan: if you want just for what is the same, but actually maybe we can have simpler: if (files.All(file => file.IndexOf(previewSubdir) > -1 ) && ... ) and then if all files are in preview they should go in preview

---
In reply to: [223893324](https://github.com/Azure/autorest.go/pull/154#discussion_r223893324) [](ancestors = 223893324,223891666)

jhendrixMSFT: Fixed, and I tested the different permutations.

fearthecowboy:
## success (version: 2.1.115)

PR #155, opened 2018-10-16
Ensure failed LRO returns HTTP response

jhendrixMSFT:
Omit the early-exit from an LRO sender as it prevents the underlying
HTTP response from being returned to the caller (the LRO state machine
can handle the success/failure status codes).
Update test dependencies.

### DIFF
--- a/test/src/tests/generated/lro/lros.go
+++ b/test/src/tests/generated/lro/lros.go
@@ -64,10 +64,6 @@ func (client LROsClient) Delete202NoRetry204Sender(req *http.Request) (future LR
 	if err != nil {
 		return
 	}
-	err = autorest.Respond(resp, azure.WithErrorUnlessStatusCode(http.StatusOK, http.StatusAccepted))
-	if err != nil {
-		return
-	}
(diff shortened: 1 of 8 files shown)


test/src/tests/generated/lro/lros.go
```diff
@@ -64,10 +64,6 @@ func (client LROsClient) Delete202NoRetry204Sender(req *http.Request) (future LR
 	if err != nil {
 		return
 	}
-	err = autorest.Respond(resp, azure.WithErrorUnlessStatusCode(http.StatusOK, http.StatusAccepted))
```
vladbarosan: so the purpose of this line is just to check if there was something else than those statuses ?

jhendrixMSFT: Correct.  The status codes here come from the `responses` section in the swagger, however there are a handful of well-defined status codes for LROs as defined by ARM and we check them in the LRO state machine so having this check is redundant plus it prevents the returned future from containing the underlying HTTP response.

fearthecowboy:
## success (version: 2.1.116)

PR #156, opened 2018-10-17
Emit spans for sdk

vladbarosan:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/2902
- add traces to all sdk calls.
- add new Advancer methods `NextWithContext()` for iterator and page types.


vladbarosan:
blocked on https://github.com/Azure/go-autorest/pull/335. Ready to go after that is merged.

test/src/tests/generated/additional-properties/pets.go
```diff
@@ -31,6 +32,11 @@ func NewPetsClientWithBaseURI(baseURI string) PetsClient {
 
 // CreateAPInProperties create a Pet which contains more properties than what is defined.
 func (client PetsClient) CreateAPInProperties(ctx context.Context, createParameters PetAPInProperties) (result PetAPInProperties, err error) {
+	ctx = tracing.StartSpan(ctx, "generated/additional-properties/PetsClient.CreateAPInProperties")
+	defer func() {
+		sc := result.StatusCode
```
jhendrixMSFT: We don't need a nil check here?

vladbarosan: yes discovered this while testing, will add a nill check and use resp instead of result

test/src/tests/generated/additional-properties/pets.go
```diff
@@ -31,6 +32,14 @@ func NewPetsClientWithBaseURI(baseURI string) PetsClient {
 
 // CreateAPInProperties create a Pet which contains more properties than what is defined.
 func (client PetsClient) CreateAPInProperties(ctx context.Context, createParameters PetAPInProperties) (result PetAPInProperties, err error) {
+	ctx = tracing.StartSpan(ctx, "generated/additional-properties/PetsClient.CreateAPInProperties")
```
jhendrixMSFT: What does the generated string look like for the SDK?

vladbarosan: its going to be the fully qualified name

vladbarosan: btw not the most happy on how we determine that. pretty much we take the outputpath and do an index on it

jhendrixMSFT: I see so something like `services/compute/mgmt/2016-03-30/compute/VirtualMachinesClient.CreateOrUpdate` then?

test/src/tests/generated/additional-properties/pets.go
```diff
@@ -91,6 +100,14 @@ func (client PetsClient) CreateAPInPropertiesResponder(resp *http.Response) (res
 
 // CreateAPInPropertiesWithAPString create a Pet which contains more properties than what is defined.
 func (client PetsClient) CreateAPInPropertiesWithAPString(ctx context.Context, createParameters PetAPInPropertiesWithAPString) (result PetAPInPropertiesWithAPString, err error) {
+	ctx = tracing.StartSpan(ctx, "generated/additional-properties/PetsClient.CreateAPInPropertiesWithAPString")
```
jhendrixMSFT: Per offline discussion please update codegen to not do any of this work if `tracing.Enabled() == false`.

src/Model/IteratorTypeGo.cs
```diff
@@ -35,6 +37,11 @@ public IteratorTypeGo(PageTypeGo pageType) : base(CodeNamerGo.Instance.GetIterat
         /// </summary>
         public string IndexField => "i";
 
+        /// <summary>
+        /// The fully qualified name for the advancer method.
+        /// </summary>
+        public string AdvancerFqdn { get; private set; }
```
jhendrixMSFT: Nit, `private set` is not necessary as it's implied when setting in ctor.

test/src/tests/generated/additional-properties/pets.go
vladbarosan: yes, here is a sample, https://gist.github.com/vladbarosan/3ab6cd696e9b7413194ab78ed75225c3 . Was wondering actually, I could prefix with github.com/Azure/Azure-sdk-for-go

jhendrixMSFT: Good idea.  It might be best to emit the package import path as an unexported const then concatenate that with the fully qualified method name.

test/src/tests/generated/additional-properties/pets.go
```diff
@@ -31,6 +33,18 @@ func NewPetsClientWithBaseURI(baseURI string) PetsClient {
 
 // CreateAPInProperties create a Pet which contains more properties than what is defined.
 func (client PetsClient) CreateAPInProperties(ctx context.Context, createParameters PetAPInProperties) (result PetAPInProperties, err error) {
+	// Not necessary to perform this check as nothing will be instrumented if it is false, but
+	// adding it to avoid any potential perf issue.
+	if tracing.IsEnabled() {
+		ctx = tracing.StartSpan(ctx, fmt.Sprintf("%s/PetsClient.CreateAPInProperties", fqdn))
```
jhendrixMSFT: Nit, I would simply do `fqdn + "/PetsClient.CreateAPInProperties"`.

test/src/tests/generated/additional-properties/pets.go
```diff
@@ -31,6 +33,18 @@ func NewPetsClientWithBaseURI(baseURI string) PetsClient {
 
 // CreateAPInProperties create a Pet which contains more properties than what is defined.
 func (client PetsClient) CreateAPInProperties(ctx context.Context, createParameters PetAPInProperties) (result PetAPInProperties, err error) {
+	// Not necessary to perform this check as nothing will be instrumented if it is false, but
```
jhendrixMSFT: Hmm I'm wondering if this comment should be removed.  It adds two lines per API making the files longer and it's likely nobody will ever read it.

vladbarosan: well my reason was just not to make people think they need to do  an `isEnabled()` check. If you think is overkill I can remove it

vladbarosan: and let me know if there is anything else

vladbarosan: thought about it a little, and yea seems a little overkill, makes more sense maybe to add this note in the go-autorest functions directly, so removing it

vladbarosan:
@jhendrixMSFT  updated

PR #157, opened 2018-10-26
Create preperer when needed for reused pages across methods

vladbarosan:


src/Model/PageTypeGo.cs
```diff
@@ -133,5 +133,18 @@ public override int GetHashCode()
         {
             return Name.GetHashCode();
         }
+
+        /// <summary>
+        /// Updates the page with properties from the passed page. Throws an exception of pages are not equal.
```
jhendrixMSFT: Nit: The `of` should be `if`.

src/Model/PageTypeGo.cs
```diff
@@ -133,5 +133,18 @@ public override int GetHashCode()
...
+
+        /// <summary>
+        /// Updates the page with properties from the passed page. Throws an exception of pages are not equal.
+        /// </summary>
+        /// <param name="page"></param>
+        public void MergeWithPage(PageTypeGo page)
+        {
+            if (!this.Equals(page))
+            {
+                throw new InvalidOperationException("CreatePageableTypeForMethod requires method to be a pageable operation");
```
jhendrixMSFT: I think the `CreatePageableTypeForMethod` should be `MergeWithPage` yes?

vladbarosan: ah yea will just use `nameof()`

fearthecowboy:
## success (version: 2.1.118)

PR #158, opened 2018-10-30
Ensure single slash in package fqdn

vladbarosan:
- we could have just removed the slash but preferred to make sure its only 1 to avoid any issue coming from the config readme where it has or not the /


fearthecowboy:
## success (version: 2.1.119)

PR #159, opened 2018-11-01
Fix marshalling flattened fields

jhendrixMSFT:
Flattened fields are embedded as anonymous structs however the
marshalling codegen didn't take that into account.

### DIFF
--- a/src/Model/CompositeTypeGo.cs
+++ b/src/Model/CompositeTypeGo.cs
@@ -272,7 +272,7 @@ namespace AutoRest.Go.Model
                 }
                 if (!string.IsNullOrWhiteSpace(property.Documentation))
                 {
-                    indented.Append($"{property.Name} - {property.Documentation}".ToCommentBlock());
+                    indented.Append($"{property.FieldName} - {property.Documentation}".ToCommentBlock());
                 }
 
                 indented.AppendLine(property.Field);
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
This fixes the codegen issue we were seeing with the `storage` management package.

jhendrixMSFT:
Still need to regen the SDK to ensure no unexpected changes.

vladbarosan:
@jhendrixMSFT can you link the issue if there is one please ?

src/Model/CompositeTypeGo.cs
```diff
@@ -272,7 +272,7 @@ public virtual string Fields()
                 }
                 if (!string.IsNullOrWhiteSpace(property.Documentation))
                 {
-                    indented.Append($"{property.Name} - {property.Documentation}".ToCommentBlock());
+                    indented.Append($"{property.FieldName} - {property.Documentation}".ToCommentBlock());
```
vladbarosan: whats the difference ?

vladbarosan: oh never mind i saw in next file

jhendrixMSFT:
I forgot to open an issue.  For reference here's the PR where this was first noticed.
https://github.com/Azure/azure-sdk-for-go/pull/3167
Take a look at BlobServiceProperties.MarshalJSON() in models.go.  On line 1208 you can see the field name is incorrect.

jhendrixMSFT:
Regenerated SDK all looks good.

PR #160, opened 2018-11-05
Check object for nil before marshalling

jhendrixMSFT:
Check objects for nil before marshalling; this avoids inserting a
`"foo": null` into the JSON which some services reject.

### DIFF
--- a/src/Templates/ModelTemplate.cshtml
+++ b/src/Templates/ModelTemplate.cshtml
@@ -155,7 +155,8 @@
 
         @foreach (var property in Model.AllProperties.Where(p => !string.IsNullOrEmpty(p.SerializedName)))
         {
-            if (property.IsPointer || property.ModelType is DictionaryTypeGo)
+            // must check object for nil to avoid inserting `"foo": null` into the JSON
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Fix for https://github.com/Azure/azure-sdk-for-go/issues/2608.

fearthecowboy:
## success (version: 2.1.121)

tombuildsstuff:
@jhendrixMSFT (admittedly I’ve not tested this yet) - but won’t this change break the ability to remove previously set fields for any API that allows Patching a resource?

jhendrixMSFT:
@tombuildsstuff I was discussing this with @vladbarosan yesterday.  You are correct however the only fields that are affected are objects, example [here](https://github.com/Azure/azure-sdk-for-go/blob/master/services/resources/mgmt/2018-05-01/resources/models.go#L769).  All other fields already have this problem so now they are uniform.
One way I was thinking we can solve this is by introducing setter methods that will distinguish between a field being unset, set, or set to nil so we generate the expected JSON (I will fab up an example to make this clear).  Totally open to suggestions as well.

PR #161, opened 2018-11-09
Generate interfaces

jhendrixMSFT:
Create an interfaces.go file that contains interfaces for each operation
group.  Only the defined operations will be in the interfaces, i.e. they
will not contain preparers/senders/responders.
Generate factory methods for pageable and iterator types.


test/src/tests/generated/paging/models.go
```diff
@@ -185,6 +190,11 @@ func (page OdataProductResultPage) Values() []Product {
 	return *page.opr.Values
 }
 
+// Creates a new instance of the OdataProductResultPage type.
+func NewOdataProductResultPage(getNextPage func(context.Context, OdataProductResult) (OdataProductResult, error)) OdataProductResultPage {
```
vladbarosan: hmm, can we create a function type , instead of inlining it ?

vladbarosan: hmm, can we create a function type , instead of inlining it ?

test/src/tests/generated/body-array/interfaces.go
```diff
@@ -0,0 +1,89 @@
+package arraygroup
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+//
+// Code generated by Microsoft (R) AutoRest Code Generator.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+
+import (
```
vladbarosan: i thought we would do the interfaces in a subdir ?

vladbarosan: i thought we would do the interfaces in a subdir ?

jhendrixMSFT: Yep working on that now.  It gets a little tricky in the codegen since some param/return types are defined in the parent package.

src/Templates/InterfacesTemplate.cshtml
```diff
@@ -0,0 +1,81 @@
...
+@using System.Collections.Generic;
+@using System.Linq;
+
+@inherits AutoRest.Core.Template<AutoRest.Go.Model.CodeModelGo>
+
+@{
+    var content = new Dictionary<string, IEnumerable<MethodGo>>();
+    var imports = new HashSet<string> { PrimaryTypeGo.GetImportLine(package: "context") };
+
+    if (Model.ClientMethods.Any())
```
vladbarosan: is this if really necessary ? if there are not client methods, I guess it just wont add anything ?

src/Templates/InterfacesTemplate.cshtml
```diff
@@ -0,0 +1,81 @@
...
+    {
+        content.Add(Model.BaseClient, Model.ClientMethods.OrderBy(m => m.Name.Value));
+    }
+
+    var methodGroups = Model.MethodGroups.Where(mg => !string.IsNullOrEmpty(mg.Name));
```
vladbarosan: not sure why is this necessary ? if it is can you add a comment or maybe a method in the Model itself ?

src/Templates/InterfacesTemplate.cshtml
```diff
@@ -0,0 +1,81 @@
...
+    if (methodGroups.Any())
```
vladbarosan: same not sure if this is necessary.

vladbarosan: same not sure if this is necessary.

src/Templates/InterfacesTemplate.cshtml
```diff
@@ -0,0 +1,81 @@
...
+                method.ReturnType.Body.AddImports(imports);
+            }
+            else
+            {
+                addArImport = true;
+            }
+        }
+    }
+
+    var arImport = PrimaryTypeGo.GetImportLine(package: "github.com/Azure/go-autorest/autorest");
```
vladbarosan: so this is needed only if there are methods without a return value ? is that the only case when that import is necessary ? in that case can we extract a isAutorestImportNeeded method or something like that

vladbarosan: so this is needed only if there are methods without a return value ? is that the only case when that import is necessary ? in that case can we extract a isAutorestImportNeeded method or something like that

src/Templates/InterfacesTemplate.cshtml
```diff
@@ -0,0 +1,81 @@
...
+    if (addArImport && !imports.Contains(arImport))
```
vladbarosan: nit: since its a hashHset can we skip the second part of the if ?

vladbarosan: actually cant we just add it in the else branch instead of setting the addArImport flag

src/Templates/InterfacesTemplate.cshtml
jhendrixMSFT: I think you're correct, will remove.

src/Templates/InterfacesTemplate.cshtml
jhendrixMSFT: So `Model.MethodGroups` will contain all method groups including the ones in `Model.ClientMethods` which are just unnamed method groups.

src/Templates/InterfacesTemplate.cshtml
jhendrixMSFT: Will remove.

test/src/tests/generated/paging/models.go
jhendrixMSFT: So I didn't because this is the only place where the type is publicly exposed and I didn't want to add more stuff to the namespace.  Are you worried about the length?

vladbarosan: hmm if its the only place used, I guess its fine... only thing would be if they wan to pass the function around through their code they would need to define themselves the type or just have the param as we do

jhendrixMSFT: Good point.  I suggest we leave it as-is and if we get a complaint/ask then we make that change.

test/src/tests/generated/booleangroup/booleangroupapi/interfaces.go
```diff
@@ -0,0 +1,31 @@
...
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
+
+import (
+	"context"
+	"github.com/Azure/go-autorest/autorest"
+	"tests/generated/booleangroup"
+)
+
+// BaseClientAPI contains the set of methods on the BaseClient type.
+type BaseClientAPI interface {
```
jhendrixMSFT: Wait this is ugly and needs to be fixed.

fearthecowboy:
## success (version: 2.1.123)

PR #162, opened 2018-11-09
Normalize test directories

jhendrixMSFT:
The test package directories didn't match the package names which goes
against the best practice so this has been fixed.
No functional changes.


fearthecowboy:
## success (version: 2.1.122)

PR #163, opened 2018-11-12
Fix NullReferenceException when package-name isn't specified

jhendrixMSFT:


src/Model/CodeModelGo.cs
```diff
@@ -112,7 +112,7 @@ public string PackageFqdn
         {
             get
             {
-                var pkgName = Settings.Instance.Host?.GetValue<string>("package-name").Result.ToLowerInvariant();
+                var pkgName = Settings.Instance.Host?.GetValue<string>("package-name").Result;
```
vladbarosan: cant we do Result?.ToLowerInvariant()

fearthecowboy:
## success (version: 2.1.124)

PR #164, opened 2018-11-13
Fix interface codegen

jhendrixMSFT:
The interfaces codegen didn't add the package name for all cases,
specifically for sequence and dictionary types.
Worked around an autorest.common bug that doesn't set the CodeModel on
sequence and dictionary types.


src/Model/MethodGo.cs
```diff
@@ -125,9 +125,14 @@ public string MethodParametersSignature(bool includePkgName = false)
 
         private string ParameterTypeSig(IModelType type, bool includePkgName)
         {
-            if (includePkgName && type is CompositeTypeGo)
+            if (includePkgName)
             {
-                return $"{CodeModel.Namespace}.{type.Name}";
+                if (type.IsUserDefinedType())
+                    return $"{CodeModel.Namespace}.{type.Name}";
+                else if (type is SequenceTypeGo stg)
```
vladbarosan: why do we need the package prefix on these ? wont all the types be in the same package ?

jhendrixMSFT: no the generated interfaces are in a sub-package.

src/TransformerGo.cs
```diff
@@ -47,6 +47,14 @@ private static void TransformPropertyTypes(CodeModelGo cmg)
                     {
                         property.Name = property.ModelType.HasInterface() ? property.ModelType.GetInterfaceName() : property.ModelType.Name.Value;
                     }
+                    if (property.ModelType is SequenceTypeGo stg)
```
vladbarosan: did we figure out why those are missing ?

src/Model/MethodGo.cs
vladbarosan: oh ok, for the interfaces cuz they are in different packages.

src/TransformerGo.cs
jhendrixMSFT: yes it's a bug in autorest.common.  I discussed with @fearthecowboy and he suggested this workaround.

src/Model/MethodGo.cs
vladbarosan: ah vs code comments dont sink very well

fearthecowboy:
## success (version: 2.1.125)

PR #165, opened 2018-11-16
Simplify GetClientPropertyName()

jhendrixMSFT:
Use ClientProperty.Name instead of trying to calculate it.  The
calculation is broken for cases where names are camel-cased due to
invalid characters that were removed during naming.  I.e. user.name is
converted to UserName during initial naming; this will translate to a
parameter name of username which incorrectly camel-cases to Username
using the old algorithm.

### DIFF
--- a/src/Model/ParameterGo.cs
+++ b/src/Model/ParameterGo.cs
@@ -265,18 +265,7 @@ namespace AutoRest.Go.Model
         /// </summary>
         internal string GetClientPropertryName()
         {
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
I verified that the fix for https://github.com/Azure/azure-sdk-for-go/issues/2174 has not regressed.

fearthecowboy:
## success (version: 2.1.126)

PR #167, opened 2018-12-06
Fix interface name collisions

jhendrixMSFT:
If an interface name collides add an extra "Basic" prefix.
Fixed perf issue with common initialisms.
Updated autorest.common dependency.

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -32,6 +32,9 @@ namespace AutoRest.Go
         // contains a map from a model type to its corresponding interface name
         private static Dictionary<IModelType, string> s_interfaceNames = new Dictionary<IModelType, string>();
 
+        // contains a map from a one-word string to multiple words e.g. "FooBarBaz" => ["Foo", "Bar", "Baz"]
+        private static Dictionary<string, string[]> s_wordMap = new Dictionary<string, string[]>();
+
         /////////////////////////////////////////////////////////////////////////////////////////
         //
         // General Extensions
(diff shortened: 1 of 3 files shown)


src/Extensions.cs
```diff
@@ -370,9 +380,19 @@ public static bool HasInterface(this IModelType type)
         /// <returns></returns>
         public static string GetInterfaceName(this IModelType type, bool includePkgName = false)
         {
+            // this function is called in a *LOT* of places so is perf sensitive
```
vladbarosan: when you say this function you mean  `GetInterfaceName` i presume

jhendrixMSFT: Correct.

src/Extensions.cs
```diff
@@ -370,9 +380,19 @@ public static bool HasInterface(this IModelType type)
         /// <returns></returns>
         public static string GetInterfaceName(this IModelType type, bool includePkgName = false)
         {
+            // this function is called in a *LOT* of places so is perf sensitive
+            if (!s_interfaceNames.ContainsKey(type))
+            {
+                var baseName = $"Basic{type.Name}";
```
vladbarosan: nit: interfaceName instead of baseName ?

jhendrixMSFT: Will fix.  I also have another pending fix for perf, will include in that.

src/Extensions.cs
```diff
@@ -370,9 +380,19 @@ public static bool HasInterface(this IModelType type)
         /// <returns></returns>
         public static string GetInterfaceName(this IModelType type, bool includePkgName = false)
         {
+            // this function is called in a *LOT* of places so is perf sensitive
+            if (!s_interfaceNames.ContainsKey(type))
+            {
+                var baseName = $"Basic{type.Name}";
+                if (type.CodeModel.AllModelTypes.Any(mt => mt.Name == baseName))
```
vladbarosan: Nit: in case there is already a BasicBasic*, etc, maybe a while here ?

fearthecowboy:
## success (version: 2.1.127)

PR #169, opened 2019-04-15
Don't send read-only properties over the wire

jhendrixMSFT:
Upcoming changes to some services will reject API calls that send values
for read-only properties.  Set all read-only properties to nil in the
preparer and omit them from custom marshalling.  Emit a comment
indicating that a field is read-only.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest/issues/2545

fearthecowboy:
## success (version: 2.1.129)

PR #171, opened 2019-04-22
Don't set enum properties to nil

jhendrixMSFT:
The change in 0bf5bf was incorrect for enum properties, set them to
empty string as they are not pointer types.

### DIFF
--- a/src/Templates/MethodTemplate.cshtml
+++ b/src/Templates/MethodTemplate.cshtml
@@ -190,9 +190,10 @@ func (client @(Model.Owner)) @(Model.Name)(@Model.MethodParametersSignature()) (
             @EmptyLine
         </text>
     }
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
Validating entire SDK will build with this change, will update when done.

vladbarosan:
so `omitempty` will omit the empty strings as they are the nil value for string I guess ?

jhendrixMSFT:
Yep, try the following
```go
type fooThing struct {
	ID     *string `json:",omitempty"`
	Length *int    `json:",omitempty"`
	Type   string  `json:",omitempty"`
}

func main() {
	ft := fooThing{}
	b, err := json.Marshal(&ft)
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))
}
```
It prints `{}`

fearthecowboy:
## success (version: 2.1.131)

PR #172, opened 2019-05-03
Support generation of go.mod files

jhendrixMSFT:
Disabled by default.  By specifying --gomod-root arg a module name
will be calculated based on the root value and the output directory.

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -120,13 +120,22 @@ namespace AutoRest.Go
                 }
                 // module name is everything to the right of the start of the module root
                 var gomodTemplate = new GoModTemplate { Model = new GoMod(normalized.Substring(i)) };
-                await Write(gomodTemplate, "go.mod");
+                await Write(gomodTemplate, $"{StagingDir()}go.mod");
             }
         }
 
         private string FormatFileName(string fileName)
         {
-            return $"{fileName}{ImplementationFileExtension}";
+            return $"{StagingDir()}{fileName}{ImplementationFileExtension}";
+        }
+
+        private string StagingDir()
+        {
+            if (!Settings.Instance.Host.GetValue<bool>("stage").Result)
+            {
+                return "";
+            }
+            return "stage/";
         }
     }
 }
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -107,6 +107,21 @@ namespace AutoRest.Go
             // Version
             var versionTemplate = new VersionTemplate { Model = codeModel };
             await Write(versionTemplate, FormatFileName("version"));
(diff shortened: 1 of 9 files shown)


src/Model/GoMod.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace AutoRest.Go.Model
+{
+    public class GoMod
+    {
+        public GoMod(string goMod)
+        {
+            Module = goMod;
+        }
+
+        public string Module { get; }
```
vladbarosan: I guess it doesnt really matter for now, but should this be readonly also ?

vladbarosan: I guess it doesnt really matter for now, but should this be readonly also ?

src/CodeGeneratorGo.cs
```diff
@@ -107,11 +107,35 @@ public override async Task Generate(CodeModel cm)
...
+                // module name is everything to the right of the start of the module root
+                var gomodTemplate = new GoModTemplate { Model = new GoMod(normalized.Substring(i)) };
+                await Write(gomodTemplate, $"{StagingDir()}go.mod");
+            }
         }
 
         private string FormatFileName(string fileName)
         {
-            return $"{fileName}{ImplementationFileExtension}";
+            return $"{StagingDir()}{fileName}{ImplementationFileExtension}";
```
vladbarosan: nit: arrow function ?

src/Model/GoMod.cs
jhendrixMSFT: It is read-only by virtue of no setter, or am I misunderstanding?

src/CodeGeneratorGo.cs
```diff
@@ -107,11 +107,35 @@ public override async Task Generate(CodeModel cm)
             // Version
             var versionTemplate = new VersionTemplate { Model = codeModel };
             await Write(versionTemplate, FormatFileName("version"));
+
+            // go.mod file, opt-in by specifying the gomod-root arg
+            var modRoot = Settings.Instance.Host.GetValue<string>("gomod-root").Result;
```
vladbarosan: the only thing it sucks here a little, that this needs to be added to all generation configs. and if authors screw it up what happens ? 
Should we have a stricter check on what the gomod-root can be ?

jhendrixMSFT: Since the value is constant we can add this in the swagger-to-sdk config file in the repo so we don't have to specify in each config file.

src/Model/GoMod.cs
vladbarosan: yea its fine, the github extension published the comment although I thought it didnt, even twice  :/

fearthecowboy:
## success (version: 2.1.132)

PR #174, opened 2019-07-08
Update senders to use autorest.GetSendDecorators

jhendrixMSFT:
Senders codegen now uses GetSendDecorators with a default chain.
Updated dependencies and fixed LRO test.


fearthecowboy:
## success (version: 2.1.134)

PR #177, opened 2019-08-20
Misc codegen fixes

jhendrixMSFT:
For optional string parameters with constraints, add the Empty
constraint allowing it to be empty.
Dereference pointer types so the actual value is converted to string.

### DIFF
--- a/src/Templates/MethodTemplate.cshtml
+++ b/src/Templates/MethodTemplate.cshtml
@@ -227,11 +227,17 @@ func (client @(Model.Owner)) @(Model.Name)(@Model.MethodParametersSignature()) (
 
     @foreach (var p in Model.OptionalHeaderParameters)
     {
+        var deref = string.Empty;
+        if (p.ModelType.IsNumericType())
+        {
+            // optional numeric values are passed as pointers
+            deref = "*";
+        }
         <text>
             if @(p.GetEmptyCheck(p.GetParameterName(), false)) {
             preparer = autorest.DecoratePreparer(preparer,
-            @(string.Format("autorest.WithHeader(\"{0}\",autorest.String({1}))",
-                                      p.SerializedName, p.GetParameterName())))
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/5502

src/Templates/MethodTemplate.cshtml
```diff
@@ -227,11 +227,17 @@ func (client @(Model.Owner)) @(Model.Name)(@Model.MethodParametersSignature()) (
...
+            // optional numeric values are passed as pointers
+            deref = "*";
+        }
         <text>
             if @(p.GetEmptyCheck(p.GetParameterName(), false)) {
             preparer = autorest.DecoratePreparer(preparer,
-            @(string.Format("autorest.WithHeader(\"{0}\",autorest.String({1}))",
-                                      p.SerializedName, p.GetParameterName())))
+            @(string.Format("autorest.WithHeader(\"{0}\",autorest.String({1}{2}))",
+                                      p.SerializedName, deref, p.GetParameterName())))
```
ArcturusZhang: I have a question here. I suppose that `autorest.String({1}{2})` here is used for dereference a variable with name of `p.GetParameterName()` and pass it to `autorest.String` method. My question is why this deref symbol `*` comes after the variable name?

ArcturusZhang: Sorry, I made a mistake. This is right and the asterisk is before the variable name.

fearthecowboy:
## success (version: 2.1.136)

ArcturusZhang:
Hi @jhendrixMSFT , should I update the version info in go-SDK repo config file to point this updated extension? If so, to which version?

jhendrixMSFT:
The version is `2.1.136`, see the comment above yours.  I already updated to this version in the `latest` branch.

PR #178, opened 2019-09-05
Add support for collection format 'multi'

jhendrixMSFT:


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/309

src/Extensions.cs
```diff
@@ -266,6 +266,7 @@ public static string GetSeparator(this CollectionFormat format)
             switch (format)
             {
                 case CollectionFormat.Csv:
+                case CollectionFormat.Multi:
                     return ",";
```
ArcturusZhang: I understand this will suppress the exception from throwing, but according to the example given by swagger doc
```
multi - corresponds to multiple parameter instances instead of multiple values for a single instance foo=bar&foo=baz. This is valid only for parameters in "query" or "formData".
```
does the multi format use comma as a separator? The doc does not make this very clear

jhendrixMSFT: Agreed it's not very clear.  The implementation generates code that passes the tests so I have some confidence.  @fearthecowboy can you please clarify the behavior here?

fearthecowboy: O_O 

`multi` means that given an array of values for `foo` : `['bar','bin','baz']` that should encode as  `foo=bar&foo=bin&foo=baz` in the query or formdata.

Should not be commas. 

I'm thinking the test framework is probably too lax for picking up the values and permits commas. 

The good news is that next week I should be starting the new compliance test stuff using wiremock; hopefully we'll have a bit better set of tests for stuff in the not-too-distant-future.

jhendrixMSFT: Thanks for clarifying; that's what I originally thought, unfortunately the tests don't reflect the reality at all :(.  I will update my change accordingly.  I can update the tests as well however it will likely cause failures for other generators if they implement it in this broken way.

jhendrixMSFT: Should be working now, no changes required to the test server required.

fearthecowboy:
## success (version: 2.1.137)

PR #184, opened 2019-12-20
Don't omit local parameters named api-version

jhendrixMSFT:
The logic for embedding client property api-version as a constant query
parameter value was inadvertently removing api-version when specified as
a method parameter leading to incorrect codegen.

Pin test server version and update tests.


src/Model/ParameterGo.cs
```diff
@@ -61,11 +55,9 @@ public string GetParameterName()
             return retval;
         }
 
-        public override bool IsClientProperty => base.IsClientProperty == true && !IsAPIVersion;
-
         public virtual bool IsAPIVersion => SerializedName.IsApiVersion();
 
-        public virtual bool IsMethodArgument => !IsClientProperty && !IsAPIVersion && !IsConstant;
```
jhendrixMSFT: This is the meat of the fix, the rest is removing needless clutter.

jhendrixMSFT:
CC @catalinaperalta

fearthecowboy:
## success (version: 2.1.139)

jhendrixMSFT:
@ArcturusZhang could you try out this new build and if it looks ok update the latest branch?  I built a handful of services and it looked ok.

paulomarquesc:
Thanks guys for the quick repsone on this.

ArcturusZhang:
> @ArcturusZhang could you try out this new build and if it looks ok update the latest branch? I built a handful of services and it looked ok.

Sure

PR #185, opened 2020-01-06
Include ListComplete method in mock interfaces

jhendrixMSFT:
Update comment for client constructors.


jhendrixMSFT:
Fixes:
https://github.com/Azure/azure-sdk-for-go/issues/6559
https://github.com/Azure/azure-sdk-for-go/issues/6605

src/Templates/ServiceClientTemplate.cshtml
```diff
@@ -67,7 +67,7 @@ else
 {
 <text>
 @EmptyLine
-@WrapComment("// ", string.Format("NewWithBaseURI creates an instance of the {0} client.", Model.BaseClient))
+@WrapComment("// ", string.Format("NewWithBaseURI creates an instance of the {0} client using a custom endpoint.  Use this when interacting with an Azure cloud that uses a non-standard base URI (sovereign clouds, Azure stack).", Model.BaseClient))
```
catalinaperalta: Just checking, we don't want the name of the client in this method name? (like: New{0}WithBaseURI)

jhendrixMSFT: In track 1 we have these base clients that don't have a specific name, see https://github.com/Azure/azure-sdk-for-go/blob/master/services/redis/mgmt/2018-03-01/redis/client.go#L45 for an example.

catalinaperalta: I see, thanks for the info!

jhendrixMSFT:
CC @ArcturusZhang

fearthecowboy:
## success (version: 2.1.140)

PR #195, opened 2020-01-14
Adding ByteGroup acceptance code

catalinaperalta:
Adding code for ByteGroup acceptance testing.


test/acceptance/bytegroup/byte.go
```diff
@@ -0,0 +1,159 @@
+package bytegroup
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jhendrixMSFT: This comment block should go at the beginning of the file.

test/acceptance/bytegroup/go.mod
```diff
@@ -0,0 +1,11 @@
+module bytegroup
```
jhendrixMSFT: Since we won't actually be releasing any of these packages I think we should make test One Big Module, i.e. move this to test/go.mod so we don't have to deal with a bunch of go.mod/go.sum files.  What do you think?

test/acceptance/bytegroup/go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+go 1.13
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v0.1.0
+	github.com/Azure/go-autorest/autorest v0.9.3
```
jhendrixMSFT: I'm guessing there's no real dependency on go-autorest?  You can run `go mod tidy` to clean this stuff up.

catalinaperalta: Done

test/acceptance/bytegroup/byte.go
catalinaperalta: Done

test/acceptance/bytegroup/go.mod
catalinaperalta: Yes that makes sense, I think we could call it something like generatortests?

jhendrixMSFT: Sounds good to me.

catalinaperalta: OK, then we're all set. Updated the module name and moved the files. 👍

PR #203, opened 2020-01-17
Fix error models to match what's sent over the wire

jhendrixMSFT:


test/autorest/generated/bytegroup/internal/bytegroup/models.go
```diff
@@ -7,7 +7,7 @@ import "github.com/Azure/azure-sdk-for-go/sdk/azcore"
 
 // Error ...
 type Error struct {
-	Status  int32
+	Status  string
```
catalinaperalta: What is Status going to mean? Like Status OK?

jhendrixMSFT: It's pretty generic.  Here's what I saw when I first encountered this.
```json
{"status":"Error","message":"Endpoint not found"}
```
I've got a PR out to fix the swagger for this service, see https://github.com/Azure/autorest.testserver/pull/127

catalinaperalta: I see, that makes sense then.

Issue #190, opened 2020-01-13
Write test runners for "hello world" tests


jhendrixMSFT:
Once reference implementations are committed, write a test runner for each test that runs against the local test server.

catalinaperalta:
string_getMbcs and string_putMbcs tests added in this PR: #201
basic_getValid and basic_putValid tests added in this PR: #204

Issue #189, opened 2020-01-13
Create reference implementation for "hello world" tests


jhendrixMSFT:
Create a hand-written reference implementation for the set of "hello world" tests.

- basic_getValid
- basic_putValid
- string_getMbcs
- string_putMbcs

catalinaperalta:
basic_getValid and basic_putValid added in this PR: #204

PR #204, opened 2020-01-17
Track2 complexgroup

catalinaperalta:
This PR adds code related to the reference samples for the following testserver endpoints:
basic_getValid
basic_putValid

Related to issues: #189 and #190


test/autorest/generated/complexgroup/internal/complexgroup/service.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// Service ..
+type Service struct{}
+
+// GetValidCreateRequest creates the GetValid request.
+func (Service) GetValidCreateRequest(u url.URL) (*azcore.Request, error) {
+	u.Path = path.Join(u.Path, "/complex/basic/valid")
```
jhendrixMSFT: The fact that this doesn't need an `api-version` query param is a bug in the [response file](https://github.com/Azure/autorest.testserver/blob/master/mappings/vanilla/complex_basic_valid-62466d5b-943d-47d3-a192-520d96d5d6c3.json#L5).  Can you please confirm?

test/autorest/generated/complexgroup/internal/complexgroup/models.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (e Error) Error() string {
+	return e.Message
+}
+
+// Basic ..
+type Basic struct {
+	ID    int    `json:"id,omitempty"`
+	Name  string `json:"name,omitempty"`
+	Color string `json:"color,omitempty"`
```
jhendrixMSFT: Color should be an enum.

test/autorest/generated/complexgroup/internal/complexgroup/models.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// GetValidResponse ...
+type GetValidResponse struct {
+	StatusCode int
+	Value      Basic
```
jhendrixMSFT: The field name should be the same as the model name, so `Basic`.  We use `Value` only for operations that return scalar types.
Also, should be pointer-to-type.

test/autorest/generated/complexgroup/internal/complexgroup/service.go
catalinaperalta: Yes, GetValid is currently able to respond without the api-version query parameter. PutValid on the other hand requires api-version in order to accept the request.

test/autorest/generated/complexgroup/internal/complexgroup/models.go
```diff
@@ -0,0 +1,42 @@
...
+	return err
+}
+
+func (e Error) Error() string {
+	return e.Message
+}
+
+// Basic ..
+type Basic struct {
+	ID    int    `json:"id,omitempty"`
```
jhendrixMSFT: Fields need to be pointer-to-type, with the exception of maps and slices.  Same applies to `Error` as well (I missed this earlier).

test/autorest/generated/complexgroup/internal/complexgroup/models.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package complexgroup
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+// Error ...
+type Error struct {
```
jhendrixMSFT: We should add the correct JSON tags to the fields on this type (another thing I missed).

test/autorest/generated/complexgroup/internal/complexgroup/models.go
catalinaperalta: You mean we need to have something like this?

// Basic ..
type Basic struct {
	ID    *int       `json:"id,omitempty"`
	Name  *string    `json:"name,omitempty"`
	Color *ColorType `json:"color,omitempty"`
}

test/autorest/generated/complexgroup/internal/complexgroup/models.go
catalinaperalta: Done

test/autorest/generated/complexgroup/internal/complexgroup/models.go
catalinaperalta: Good to know. I've updated it!

test/autorest/generated/complexgroup/internal/complexgroup/models.go
catalinaperalta: Done. I added JSON tags to the error struct

test/autorest/generated/complexgroup/internal/complexgroup/models.go
catalinaperalta: For now I'm assuming that you do mean what I put above and have pushed the changes. So also done?

test/autorest/generated/complexgroup/internal/complexgroup/models.go
```diff
@@ -0,0 +1,61 @@
...
+	// ColorMagenta ...
+	ColorMagenta ColorType = "Magenta"
+	// ColorCyan ...
+	ColorCyan ColorType = "cyan"
+	// ColorBlack ...
+	ColorBlack ColorType = "blacK"
+)
+
+// ColorValues ...
+func ColorValues() []ColorType {
```
jhendrixMSFT: PossibleColorValues()

catalinaperalta: Done

PR #205, opened 2020-01-22
Adding Custom Base URL 

catalinaperalta:
This PR adds the samples for the custom base URL swagger and its tests.

### DIFF
--- a/test/autorest/generated/custombaseurlgroup/complex.go
+++ b/test/autorest/generated/custombaseurlgroup/complex.go
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package custombaseurlgroup
+
+import (
+	"context"
+	azinternal "generatortests/autorest/generated/custombaseurlgroup/internal/custombaseurlgroup"
+	"net/url"
+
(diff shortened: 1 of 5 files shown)


test/autorest/generated/custombaseurlgroup/complex.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"context"
+	azinternal "generatortests/autorest/generated/custombaseurlgroup/internal/custombaseurlgroup"
+	"net/url"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// DefaultEndpoint is the default endpoint used for the CustomBaseURLClient service.
+const DefaultEndpoint = "http://localhost:3000"
```
jhendrixMSFT: This should be omitted.  The custom-baseUrl swagger uses the `x-ms-parameterized-host` extension which means the swagger doesn't define a base URL and expects the caller to provide it.  I realize this doesn't completely map with what we're doing in track 2 since we always require this parameter now.
This got me thinking though, that for swaggers that define a host we should emit a comment in the constructor that says "Use DefaultEndpoint for blah blah" or something.

catalinaperalta: Ah, you're right. Yes that makes sense to me. I removed it.

PR #209, opened 2020-01-23
Adding to Body Complex reference samples endpoints

catalinaperalta:
Adding full Basic model endpoint references (and renamed it from complex client to basic client) and initial primitive model endpoint references.


test/autorest/complexgroup/complexgroup_test.go
```diff
@@ -11,8 +11,16 @@ import (
 	"testing"
 )
 
-func getComplexClient(t *testing.T) *complexgroup.ComplexClient {
-	client, err := complexgroup.NewComplexClient(complexgroup.DefaultEndpoint, nil)
+func getBasicClient(t *testing.T) *complexgroup.BasicClient {
```
catalinaperalta: The change here is simply renaming all of the ComplexClient references to BasicClient since ComplexClient was too broad.

test/autorest/complexgroup/complexgroup_test.go
```diff
@@ -67,3 +75,85 @@ func TestPutValid(t *testing.T) {
 	}
 	deepEqualOrFatal(t, result, expected)
 }
+
+// TODO check this
+func TestGetInvalid(t *testing.T) {
```
catalinaperalta: Wondering what to do here, since the test specifically sends a string instead of an integer and obviously that cannot be unmarshalled into the Basic model

test/autorest/complexgroup/complexgroup_test.go
```diff
@@ -67,3 +75,85 @@ func TestPutValid(t *testing.T) {
...
+	}
+	expected := &complexgroup.GetEmptyResponse{
+		StatusCode: http.StatusOK,
+		Basic:      &complexgroup.Basic{},
+	}
+	deepEqualOrFatal(t, result, expected)
+}
+
+// TODO nil or null?
+func TestGetNull(t *testing.T) {
```
catalinaperalta: If the endpoint specifies Null will we rename everything to nil?

test/autorest/generated/complexgroup/basic.go
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package complexgroup
```
catalinaperalta: This file didnt change, I only renamed it since it was previously called complex.go which was too broad for the complex package since there are several different models within it.

test/autorest/complexgroup/complexgroup_test.go
jhendrixMSFT: This client name comes from the swagger, specifically from the `operationId`.  E.g. for `basic_getValid` the client name would be `BasicClient` with a method `GetValid`.

I didn't look at the operations that have been implemented, but you might want to double-check that the client names match the swagger definition.

test/autorest/complexgroup/complexgroup_test.go
jhendrixMSFT: So unmarshalling should fail and an error is returned right?  If that's the case then ensure that the response object is nil and the error is non-nil.  At some point we'll need to ensure that the returned errors are actionable.

test/autorest/complexgroup/complexgroup_test.go
jhendrixMSFT: Should come from the `operationId` in the swagger.

test/autorest/generated/complexgroup/basic.go
jhendrixMSFT: In track 1 we named the file to match the client name minus the `Client` suffix; let's do the same thing in track 2.

test/autorest/complexgroup/complexgroup_test.go
catalinaperalta: Yes no worries. I already noticed that and that's the reason for the changes in this PR. So everything that was previously ComplexClient is now BasicClient and I also have a PrimitiveClient for operations like: primitive_getLong. Both are part of the complexgroup package

test/autorest/complexgroup/complexgroup_test.go
catalinaperalta: Ah perfect, I'll just add the check for response object

test/autorest/complexgroup/complexgroup_test.go
catalinaperalta: Ok great, then I'm leaving it as GetNull since that's what the swagger specifies

test/autorest/generated/complexgroup/basic.go
catalinaperalta: Perfect that change is also part of this PR

test/autorest/complexgroup/complexgroup_test.go
catalinaperalta: Done

test/autorest/generated/complexgroup/primitive.go
```diff
@@ -0,0 +1,96 @@
...
+
+// PrimitiveClient is the test Infrastructure for AutoRest Swagger
+type PrimitiveClient struct {
+	s azinternal.Service
+	u *url.URL
+	p azcore.Pipeline
+}
+
+// PrimitiveClientOptions ...
+type PrimitiveClientOptions struct {
```
jhendrixMSFT: I wonder if we should have a nested `StandardClientOptions` or something that all client option structs take.  Nothing to change here just thinking out loud, we should discuss.

test/autorest/generated/complexgroup/primitive.go
```diff
@@ -0,0 +1,96 @@
...
+	s, err := client.s.GetIntHandleResponse(resp)
+	if err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+// PutInt ...
+func (client *PrimitiveClient) PutInt(ctx context.Context, complexBody IntWrapper) (*PutIntResponse, error) {
+	req, err := client.s.PutIntCreateRequest(*client.u, &complexBody)
```
jhendrixMSFT: Is there a reason `complexBody` is passed by reference here?  I think it should be passed to the creator function in the same way it's passed to the caller.

test/autorest/generated/complexgroup/primitive.go
catalinaperalta: Yes I was thinking the same, there are a few areas where we could probably reduce repeated code. I'll try to make a note of all of the places I think we could do this and we can discuss later.

test/autorest/generated/complexgroup/primitive.go
catalinaperalta: Right, I fixed it. Thanks!

PR #215, opened 2020-01-28
Adding capitalizeAcronyms function in namer.ts

catalinaperalta:


src/namer/namer.ts
```diff
@@ -45,19 +45,23 @@ async function process(session: Session<CodeModel>) {
...
     details.name = pascalCase(details.name);
+    details.name = capitalizeAcronyms(details.name)
     for (const prop of values(obj.properties)) {
       const details = <Language>prop.language.go;
       details.name = pascalCase(details.name);
+      details.name = capitalizeAcronyms(details.name)
     }
   }
 
   // pascal-case operation groups and their operations
```
catalinaperalta: IF we agree that this is the best place to call capitalizeAcronyms then I'll update the comments

src/namer/namer.ts
```diff
@@ -79,3 +84,50 @@ function cloneLanguageInfo(graph: any) {
     }
   }
 }
+
+let acronyms = [
```
jhendrixMSFT: `const` instead of `let`

src/namer/namer.ts
```diff
@@ -79,3 +84,50 @@ function cloneLanguageInfo(graph: any) {
...
+  "Ui",
+  "Uid",
+  "Uuid",
+  "Uri",
+  "Url",
+  "Utf8",
+  "Vm",
+  "Xml",
+  "Xsrf",
+  "Xss"]
```
jhendrixMSFT: I think the trailing `]` goes on the next line.

src/namer/namer.ts
```diff
@@ -79,3 +84,50 @@ function cloneLanguageInfo(graph: any) {
     }
   }
 }
+
+let acronyms = [
+  "Acl",
```
jhendrixMSFT: Also I believe strings are typically delimited with `'` characters.

src/namer/namer.ts
```diff
@@ -79,3 +84,50 @@ function cloneLanguageInfo(graph: any) {
...
+  "Url",
+  "Utf8",
+  "Vm",
+  "Xml",
+  "Xsrf",
+  "Xss"]
+
+function capitalizeAcronyms(name: string): string {
+  for (const word of acronyms) {
+    name = name.replace(word, word.toUpperCase())
```
jhendrixMSFT: I don't think this will do a case-insensitive comparison.  If we didn't do that in track 1 then this is probably ok although we might need to revisit later.

src/namer/namer.ts
jhendrixMSFT: Yep I think this works great.

src/namer/namer.ts
catalinaperalta: Yes, that's right. This wont do that. But I didn't want to implement that just yet, since I'm thinking through scenarios where it could cause problems. 
i.e. (a very unlikely example) curl would end up being cURL if I lowercase everything.

jhendrixMSFT: Good point, that would look bad (in fact I want to say I've seen this in track 1 codegen).

PR #216, opened 2020-01-28
initial protocol operations generation

jhendrixMSFT:


src/generator/models.ts
```diff
@@ -5,183 +5,129 @@
...
 import { values } from '@azure-tools/linq';
+import { ContentPreamble, ImportManager, SortAscending } from './helpers'
 
 // Creates the content in models.go
 export async function generateModels(session: Session<CodeModel>): Promise<string> {
-  const headerText = comment(await session.getValue("header-text", "MISSING LICENSE HEADER"), "// ");
-  const namespace = await session.getValue('namespace');
-  let text = `${headerText}\n\n`
-  text += `package ${namespace}\n\n`
+  let text = await ContentPreamble(session);
```
catalinaperalta: When should we use let vs const

jhendrixMSFT: Default should be to use `const`.  Only time you want to use `let` is if you plan to reassign some value to the variable.

PR #219, opened 2020-01-29
Escape reserved words and add mappings file

catalinaperalta:
This PR adds a method to escape reserved words with a pad string and also separated the common acronyms and reserved words mappings into their own files


src/namer/mappings.ts
```diff
@@ -0,0 +1,112 @@
+export const CommonAcronyms = [
```
jhendrixMSFT: File is missing license/copyright header.

src/namer/mappings.ts
```diff
@@ -0,0 +1,112 @@
...
+      'sync', 'atomic',
+      'syscall',
+      'testing', 'iotest', 'quick',
+      'text', 'scanner', 'tabwriter', 'template', 'parse',
+      'time',
+      'unicode', 'utf16', 'utf8',
+      'unsafe',
+
+      // Other reserved names and packages (defined by the base libraries this code uses)
+      'autorest', 'client', 'date', 'err', 'req', 'resp', 'result', 'sender', 'to', 'validation', 'm', 'v', 'k', 'objectMap'
```
jhendrixMSFT: The entries specific to go-autorest can be removed.

PR #220, opened 2020-01-29
add code generation for models in convenience layer

jhendrixMSFT:


src/generator/convenience/models.ts
```diff
@@ -0,0 +1,74 @@
...
+import { ChoiceValue, CodeModel } from '@azure-tools/codemodel';
+import { values } from '@azure-tools/linq';
+import { ContentPreamble, HasDescription, SortAscending } from '../common/helpers';
+
+const internalPackage = 'azinternal';
+
+export async function generateModels(session: Session<CodeModel>): Promise<string> {
+  let text = await ContentPreamble(session);
+
+  const module = await session.getValue('module-path');
```
catalinaperalta: Just wondering, is the module-path where we're going to store the current path to the module? Where do we set/update that?
Like in this example: 
`import azinternal "generatortests/autorest/generated/custombaseurlgroup/internal/custombaseurlgroup"`
Would module path be `generatortests/autorest/generated/custombaseurlgroup`?

jhendrixMSFT: Yes exactly, it's a required command-line arg now for the `go-convenience` phase.

PR #224, opened 2020-01-30
Refactor clients into one client

jhendrixMSFT:


test/autorest/generated/complexgroup/client.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// ClientOptions ...
+type ClientOptions struct {
+	// HTTPClient sets the transport for making HTTP requests.
+	HTTPClient azcore.Transport
+
+	// LogOptions configures the built-in request logging policy behavior.
+	LogOptions azcore.RequestLogOptions
```
JeffreyRichter: I don't think we need this struct any more.

test/autorest/generated/complexgroup/client.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Telemetry configures the built-in telemetry policy behavior.
+	Telemetry azcore.TelemetryOptions
+}
+
+// DefaultClientOptions creates a ClientOptions type initialized with default values.
+func DefaultClientOptions() ClientOptions {
+	return ClientOptions{
+		HTTPClient: azcore.DefaultHTTPClientTransport(),
+		Retry:      azcore.DefaultRetryOptions(),
```
JeffreyRichter: You should also init telemetry options here

test/autorest/generated/complexgroup/client.go
```diff
@@ -0,0 +1,83 @@
...
+func NewClient(endpoint string, options *ClientOptions) (*Client, error) {
+	if options == nil {
+		o := DefaultClientOptions()
+		options = &o
+	}
+	p := azcore.NewPipeline(options.HTTPClient,
+		azcore.NewTelemetryPolicy(options.Telemetry),
+		azcore.NewUniqueRequestIDPolicy(),
+		azcore.NewRetryPolicy(&options.Retry),
+		azcore.NewRequestLogPolicy(options.LogOptions))
```
JeffreyRichter: I think we can kill log options (but not the policy)

test/autorest/generated/complexgroup/client.go
jhendrixMSFT: I can add init methods for these in `azcore` however they use zero-init semantics to indicate "don't do anything".

JeffreyRichter:
I think the pattern should be consistent throughout the code base. The .go file that defines the policy also defines the options struct and the method to initialize it's fields (even if the fields default to zero values).

The .go file that defines the client structure also defines the client's options structure composed of the desired policy option structures and also defines a method to get the default client option structure. Internally, this method MUST call the individual policy functions that initialize reach policy option structure.

This pattern MUST be followed so that versioning works correctly. A policy might add a new field whose default value is not the zero value and, if this pattern is followed, this will work correctly.


  *   Jeffrey Richter (watch Architecting Distributed Cloud Apps<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fazure.microsoft.com%2Ftraining%2Fdistributed-cloud-apps%2F&data=02%7C01%7Cjeffreyr%40microsoft.com%7C82df4ff9e5ae46a5095908d7345e957b%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637035453166233741&sdata=4GnkpFUKxlZXGXFr6oQo5GB%2FFNIbDXIZDXUij0q4AEE%3D&reserved=0>)

________________________________

PR #225, opened 2020-01-31
Fixed array type transform assignment

catalinaperalta:
Basing array transforms on element type instead of simply assigning the name in the language.go property.

### DIFF
--- a/src/transform/transform.ts
+++ b/src/transform/transform.ts
@@ -42,32 +42,13 @@ async function process(session: Session<CodeModel>) {
 function schemaTypeToGoType(schema: Schema): string {
   switch (schema.type) {
     case SchemaType.Array:
-      const arraySchema = <ArraySchema>schema
+      const arraySchema = <ArraySchema>schema;
       switch (arraySchema.elementType.type) {
-        case SchemaType.Boolean:
-          return '[]bool';
-        case SchemaType.Date:
-          return '[]time.Time'
-        case SchemaType.DateTime:
-          return '[]time.Time';
-        case SchemaType.Duration:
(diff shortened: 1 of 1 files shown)


src/transform/transform.ts
```diff
@@ -42,7 +42,33 @@ async function process(session: Session<CodeModel>) {
...
+          return '[]time.Duration';
+        case SchemaType.Integer:
+          if ((<NumberSchema>schema).precision === 32) {
+            return '[]int32';
+          }
+          return '[]int64';
+        case SchemaType.Number:
+          if ((<NumberSchema>schema).precision === 32) {
+            return '[]float32';
+          }
```
catalinaperalta: These cases might not be very common, but it would help in the future if an array of these types exists in a swagger

src/transform/transform.ts
```diff
@@ -42,7 +42,33 @@ async function process(session: Session<CodeModel>) {
 function schemaTypeToGoType(schema: Schema): string {
   switch (schema.type) {
     case SchemaType.Array:
-      return `[]${(<ArraySchema>schema).elementType.language.go!.name}`;
+      const arraySchema = <ArraySchema>schema
```
jhendrixMSFT: Missing `;`

src/transform/transform.ts
```diff
@@ -42,7 +42,33 @@ async function process(session: Session<CodeModel>) {
...
+      switch (arraySchema.elementType.type) {
```
jhendrixMSFT: Instead of repeating this switch/case I think you could recursively call `schemaTypeToGoType()` passing in the schema of the element type.

PR #226, opened 2020-01-31
Adding transformations for swagger data types

catalinaperalta:
This PR includes:
- Transforms for array types in the swagger
- Transforms for dictionary types in the swagger
- Adding the operation group name to response types

### DIFF
--- a/src/namer/namer.ts
+++ b/src/namer/namer.ts
@@ -1,6 +1,6 @@
-/*  --------------------------------------------------------------------------------------------- *
- *  Copyright (c) Microsoft Corporation. All rights reserved.                                     *
- *  Licensed under the MIT License. See License.txt in the project root for license information.  *
+/*  ---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
  *  --------------------------------------------------------------------------------------------  */
 
 import { serialize, pascalCase } from '@azure-tools/codegen'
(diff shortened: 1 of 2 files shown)


src/namer/namer.ts
```diff
@@ -78,6 +78,7 @@ async function process(session: Session<CodeModel>) {
   // pascal-case and capitzalize acronym operation groups and their operations
   for (const group of values(model.operationGroups)) {
     const details = <Language>group.language.go;
+    const opGroupName = capitalizeAcronyms(pascalCase(group.$key));
```
jhendrixMSFT: What is `group.$key`?

src/namer/namer.ts
```diff
@@ -1,13 +1,13 @@
-/*---------------------------------------------------------------------------------------------
+/*  ---------------------------------------------------------------------------------------------
```
jhendrixMSFT: Are all these formatting changes from a tool?

src/namer/namer.ts
catalinaperalta: $key contains the operationGroup identifier

image

src/namer/namer.ts
catalinaperalta: I was playing around with the linter. I'll take it away for now. ESLint rules said that there should be a space or tab after /*

catalinaperalta: It's not all formatted completely according to the linter rules

catalinaperalta: image

Actually I'm thinking of leaving it unless you have any objections

jhendrixMSFT: I don't however if this is what the linter wants we should enable it and fix up all the source files to conform (can happen in a separate PR).

catalinaperalta: Agreed, there are definitely a few more things that the linter has brought up. I'll add those changes to another PR.

PR #227, opened 2020-01-31
Move protocol method sig info into the code model

jhendrixMSFT:
The method info generated by the protocol layer is now placed into the
code model so the convenience layer can build on it.
Updated the ImportManager to accept package aliases.

### DIFF
--- a/src/generator/common/helpers.ts
+++ b/src/generator/common/helpers.ts
@@ -5,25 +5,28 @@
 
 import { Session } from '@azure-tools/autorest-extension-base';
 import { comment } from '@azure-tools/codegen'
-import { CodeModel, Language } from '@azure-tools/codemodel';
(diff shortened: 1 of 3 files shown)


src/generator/common/helpers.ts
```diff
@@ -5,41 +5,56 @@
...
   length(): number {
     return this.imports.length
   }
 
+  // returns the import list as Go source code
   text(): string {
-    this.imports.sort(SortAscending);
+    if (this.imports.length === 1) {
+      const first = this.imports[0];
+      return `import ${this.alias(first)}"${first.imp}"\n\n`;
```
catalinaperalta: is this for the azinternal package?

jhendrixMSFT: Yeah needed to support package aliases for it.

PR #229, opened 2020-02-01
updated via code generation

jhendrixMSFT:


test/autorest/generated/bytegroup/internal/bytegroup/models.go
```diff
@@ -1,53 +1,64 @@
...
+		return err
+	}
+	return err
 }
 
-// PutNonASCIIResponse ...
-type PutNonASCIIResponse struct {
-	StatusCode int
+func (e Error) Error() string {
+	return "TODO"
```
jhendrixMSFT: We need to discuss what to return here.

test/autorest/generated/bytegroup/byte.go
```diff
@@ -1,158 +1,112 @@
...
 		return nil, err
 	}
-	return ba, nil
+	return result, nil
 }
 
-// GetInvalid get invalid byte value ':::SWAGGER::::'
-func (client *ByteClient) GetInvalid(ctx context.Context) (*GetInvalidResponse, error) {
-	req, err := client.s.GetInvalidCreateRequest(*client.u)
+func (client *byteOperations) GetInvalid(ctx context.Context) (*ByteGetInvalidResponse, error) {
```
catalinaperalta: I think the comments aren't being added.

jhendrixMSFT: I was only emitting them on the method definitions in the interface since that's what will show up via intellisense.  I can add them here too if you think it's useful (I was on the fence about it).

test/autorest/generated/bytegroup/client.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// NewClientWithPipeline creates an instance of the Client type with the specified endpoint and pipeline.
+func NewClientWithPipeline(endpoint string, p azcore.Pipeline) (*Client, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	c := &Client{u: u, p: p}
+	c.byteOperations = &byteOperations{Client: c}
```
catalinaperalta: I'm just going to admit that this part looks a bit strange to me, because we're assigning a byteOperations instance that references the client that it's going to be placed on. I know it works and I don't think it needs to be changed...but maybe we should add a comment as to why we do this? I don't think it's an issue if we don't just wanted to put it out there. (Also don't know if it's worth it since this is all generated code.)

test/autorest/generated/bytegroup/internal/bytegroup/models.go
catalinaperalta: Agreed, I think the message should include information relevant to the specific operation group. We can discuss it on Tuesday if you want.

test/autorest/generated/bytegroup/byte.go
jhendrixMSFT: Hmm I think the linter will likely complain about them missing, I'll add them.

catalinaperalta: Hmm, good point. My thought process was: if I'm a user and (have decided to ignore intellisense) I'm looking through the bytegroup code to see what is happening in GetInvalid and (for some reason) I dont understand what's happening there then I'd have to scroll back up to the method definition...which is fine...but possibly annoying? So I'm a little more for having the comments here as well.

test/autorest/generated/bytegroup/internal/bytegroup/models.go
jhendrixMSFT: We can include the group+method name, however there should be detailed error info in the fields that was returned from the service that should be included.  Maybe we just include all the fields?  At any rate we can discuss next week.

test/autorest/generated/bytegroup/client.go
jhendrixMSFT: Agreed it looks a little wonky.  It's so the operation groups have access to the URL and pipeline and this was the simplest way to pass it.

catalinaperalta: Yep, it makes sense, but maybe for someone reading the code who doesn't know...it would be confusing? Honestly, it's not a huge issue and I mean I think people can infer the reason from the operations.

PR #233, opened 2020-02-05
Rush regenerate script and generated operation groups

catalinaperalta:
This PR adds a script used with the command "rush regenerate" that will generate swaggers from the autorest.testserver. 

Also adding several operation groups that we successfully generated.


test/autorest/generated/custombaseurlgroup/internal/custombaseurlgroup/paths.go
```diff
@@ -11,19 +13,18 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
-// PathsClient ...
-type PathsClient struct{}
+type PathsOperations struct{}
 
 // GetEmptyCreateRequest creates the GetEmpty request.
-func (PathsClient) GetEmptyCreateRequest(u url.URL) (*azcore.Request, error) {
+func (PathsOperations) GetEmptyCreateRequest(u url.URL, accountName string) (*azcore.Request, error) {
```
jhendrixMSFT: Note that `accountName` doesn't actually get used.  Could you please open a bug to track this.

catalinaperalta: Yes I noticed the same thing. Will do 👍

test/autorest/generated/morecustombaseurigroup/internal/morecustombaseurigroup/paths.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"net/http"
+	"net/url"
+	"path"
+)
+
+type PathsOperations struct{}
+
+// GetEmptyCreateRequest creates the GetEmpty request.
+func (PathsOperations) GetEmptyCreateRequest(u url.URL, vault string, secret string, keyName string, keyVersion string) (*azcore.Request, error) {
```
jhendrixMSFT: More unused params.

catalinaperalta: So I believe vault and secret are unused...but keyName and keyVersion should be used I believe. And we're somehow missing SubscriptionID. 
I found this: https://github.com/Azure/autorest.testserver/blob/master/mappings/vanilla/customuri_test12_key1-a0b11188-1f63-4c4f-acd0-729a1f91d6ac.json

PR #237, opened 2020-02-07
Prototype for custom marshalling/unmarshalling of time

jhendrixMSFT:


jhendrixMSFT:
See http://choly.ca/post/go-json-marshalling/ for the details.

PR #238, opened 2020-02-07
Added transforms for any and unixtime types

catalinaperalta:


src/transform/transform.ts
```diff
@@ -72,6 +73,7 @@ function schemaTypeToGoType(schema: Schema): string {
         return 'float32';
       }
       return 'float64';
+    case SchemaType.Any:
```
jhendrixMSFT: I'm thinking that `Any` should probably be a `interface{}` type.  Thoughts?

catalinaperalta: Ah yes! That's much better than string. I agree. I'll update it.

jhendrixMSFT: We can start with that and always change it later as needed.

PR #239, opened 2020-02-07
Adding json tags to structs and adding group name to namespaces

catalinaperalta:


src/generator/protocol/models.ts
```diff
@@ -74,7 +74,7 @@ class StructDef {
         // for constants we use the underlying type name
         typeName = (<ConstantSchema>prop.schema).valueType.language.go!.name;
       }
-      text += `\t${prop.language.go!.name} ${this.byref(prop)}${typeName}\n`;
+      text += `\t${prop.language.go!.name} ${this.byref(prop)}${typeName} \`json:"${camelCase(prop.serializedName)},omitempty"\`\n`;
```
jhendrixMSFT: You should be able to omit the call to `camelCase()` here, the value should contain the exact name as it appears on the wire (if not then I think there's a bug somewhere else).

catalinaperalta: The serializedName does come in the right format, I added it for our own models. But you're right, it would cause more problems to keep it. Removing it!

PR #246, opened 2020-02-11
Adding array acceptance tests

catalinaperalta:


test/autorest/complexgroup/array_test.go
```diff
@@ -0,0 +1,85 @@
...
+	expected := &complexgroup.ArrayGetValidResponse{
+		StatusCode:   http.StatusOK,
+		ArrayWrapper: &val,
+	}
+	deepEqualOrFatal(t, result, expected)
+}
+
+// func TestArrayPutEmpty(t *testing.T) {
+// 	client := getArrayOperations(t)
+// 	result, err := client.PutEmpty(context.Background(), complexgroup.ArrayWrapper{Array: []string{}})
```
catalinaperalta: The endpoint is expecting to receive a [] but instead we're sending a {} with the above implementaiton. Any ideas on how this should be sent over?

test/autorest/complexgroup/array_test.go
```diff
@@ -0,0 +1,85 @@
...
+// 	}
+// 	expected := &complexgroup.ArrayPutEmptyResponse{
+// 		StatusCode: http.StatusOK,
+// 	}
+// 	deepEqualOrFatal(t, result, expected)
+// }
+
+// func TestArrayPutValid(t *testing.T) {
+// 	client := getArrayOperations(t)
+// 	result, err := client.PutValid(context.Background(), complexgroup.ArrayWrapper{Array: []string{"1, 2, 3, 4", "", "", "&S#$(*Y", "The quick brown fox jumps over the lazy dog"}})
```
catalinaperalta: The endpoint is expecting a null string value but sending a "" doesn't seem to work. Again, any ideas on how to send a null string value?

catalinaperalta: Also, so you dont have to look it up, in track1 we send it over like this: 
`array := []string{"1, 2, 3, 4", "", "", "&S#$(*Y", "The quick brown fox jumps over the lazy dog"}`

test/autorest/complexgroup/array_test.go
catalinaperalta: Track1 implementation: ArrayWrapper{Array: &[]string{}}

test/autorest/complexgroup/array_test.go
catalinaperalta: Added an issue to update []string to []*string here: #248. Leaving this commented until we add that.

test/autorest/complexgroup/array_test.go
catalinaperalta: And also added an issue to discuss the omitempty issue with json tags here: #247

PR #254, opened 2020-02-12
Adding complexgroup dictionary tests

catalinaperalta:


test/autorest/complexgroup/dictionary_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestDictionaryGetNull(t *testing.T) {
+	client := getDictionaryOperations(t)
+	result, err := client.GetNull(context.Background())
+	if err != nil {
+		t.Fatalf("GetNull: %v", err)
+	}
+	expected := &complexgroup.DictionaryGetNullResponse{
+		StatusCode:        http.StatusOK,
+		DictionaryWrapper: &complexgroup.DictionaryWrapper{},
```
jhendrixMSFT: This looks identical to `GetNotProvided()`, is this due to more `omitempty` shenanigans?

catalinaperalta: Actually, I would say this is more of a testserver shenanigan, the endpoint isn't expecting a true null it's expecting a null field in DictionaryWrapper. 

So I looked into what GetNotProvided returns versus GetNull. 
GetNull is expecting: {"defaultProgram":null}
GetNotProvided is expecting: {}

I also just removed `omitempty` locally and nothing changed, both tests still run successfully. I think this could still be related to how we're unmarshalling though.

jhendrixMSFT: Ah I see, thanks for the explanation.  With this info then I would expect this test to fail since its dictionary is empty.  If you remove the `omitempty` from the `DefaultProgram` field of `DictionaryWrapper` does it work as expected?

catalinaperalta: Disregard the above comment

catalinaperalta: The endpoints are expecting this: 
```
 else if (req.params.scenario === 'empty') {
      coverage['getComplexDictionaryEmpty']++;
      res.status(200).type('json').end('{"defaultProgram":{}}');
    } else if (req.params.scenario === 'null') {
      coverage['getComplexDictionaryNull']++;
      res.status(200).type('json').end('{"defaultProgram":null}');
```

catalinaperalta: So both of them are pretty similar, I think they get unmarshalled into the same format which is why the tests pass, regardless of the `omitempty` tag. Also, I'm going to update the code to remove the `val` var in GetNotProvided since it's unnecessary.

jhendrixMSFT: It seems similar to the issue we were looking at with arrays yesterday and the whole "empty array vs nil array" and is that meaningful.  For now I think just open an issue to track this and we'll look at it later.

catalinaperalta: Done. Issue is here: #255

PR #258, opened 2020-02-13
Adding tests for ReadOnlyProperty and Inheritance in complexgroup

catalinaperalta:
Adding tests for ReadOnlyProperty and Inheritance in complexgroup

### DIFF
--- a/test/autorest/complexgroup/readonlyproperty_test.go
+++ b/test/autorest/complexgroup/readonlyproperty_test.go
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package complexgrouptest
+
+import (
+	"context"
+	"generatortests/autorest/generated/complexgroup"
+	"net/http"
+	"testing"
+)
+
+func getReadonlypropertyOperations(t *testing.T) complexgroup.ReadonlypropertyOperations {
+	client, err := complexgroup.NewDefaultClient(nil)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


test/autorest/complexgroup/readonlyproperty_test.go
```diff
@@ -0,0 +1,45 @@
...
+	expected := &complexgroup.ReadonlypropertyGetValidResponse{
+		StatusCode:  http.StatusOK,
+		ReadonlyObj: &complexgroup.ReadonlyObj{ID: &id, Size: &size},
+	}
+	deepEqualOrFatal(t, result, expected)
+}
+
+func TestReadonlypropertyPutValid(t *testing.T) {
+	client := getReadonlypropertyOperations(t)
+	result, err := client.PutValid(context.Background(), complexgroup.ReadonlyObj{})
```
jhendrixMSFT: So here the test should specify `ID` and `Size` (I assume these are both RO) and it should still pass.  Probably won't right now due to lack of custom marshalling to omit RO properties.  If you can confirm, probably want to open an issue and comment out the test.

catalinaperalta: Ah I see, thanks for pointing that out. I didn't understand what was happening with this test. Yes, we aren't doing the custom marshalling to omit the RO properties so they were being sent over to the endpoint which was rejecting them. I'll update the test and comment it out.

jhendrixMSFT: Might want to open an issue too.

catalinaperalta: Done. Here's the issue: #259

PR #261, opened 2020-02-14
Use client.Send for sending requests

jhendrixMSFT:
Introduced in v13.4.0 of go-autorest, client.Send allows configuring
SendDecorators per client.


jhendrixMSFT:
NOTE: this is using a monkey-patched version of go-autorest since v13.4.0 hasn't been released yet.  Will update this PR with official v13.4.0 bits when available.

fearthecowboy:
## success (version: 2.1.142)

PR #262, opened 2020-02-14
Updating more custom base uri test to include missing param and tests

catalinaperalta:
Adding more custom uri test and modified/updated the corresponding generated code to add missing params.

### DIFF
--- a/test/autorest/generated/morecustombaseurigroup/internal/morecustombaseurigroup/paths.go
+++ b/test/autorest/generated/morecustombaseurigroup/internal/morecustombaseurigroup/paths.go
@@ -6,18 +6,21 @@
 package morecustombaseurigroup
 
 import (
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"net/http"
 	"net/url"
 	"path"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 )
 
 type PathsOperations struct{}
 
 // GetEmptyCreateRequest creates the GetEmpty request.
-func (PathsOperations) GetEmptyCreateRequest(u url.URL, vault string, secret string, keyName string, keyVersion string) (*azcore.Request, error) {
-	u.Path = path.Join(u.Path, "/customuri/{subscriptionId}/{keyName}")
-	return azcore.NewRequest(http.MethodGet, u), nil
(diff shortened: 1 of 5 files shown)


test/autorest/generated/morecustombaseurigroup/internal/morecustombaseurigroup/paths.go
```diff
@@ -6,18 +6,21 @@
...
 )
 
 type PathsOperations struct{}
 
 // GetEmptyCreateRequest creates the GetEmpty request.
-func (PathsOperations) GetEmptyCreateRequest(u url.URL, vault string, secret string, keyName string, keyVersion string) (*azcore.Request, error) {
-	u.Path = path.Join(u.Path, "/customuri/{subscriptionId}/{keyName}")
-	return azcore.NewRequest(http.MethodGet, u), nil
+func (PathsOperations) GetEmptyCreateRequest(u url.URL, vault string, secret string, keyName string, keyVersion string, subscriptionID string) (*azcore.Request, error) {
+	u.Path = path.Join(u.Path, "/customuri", subscriptionID, keyName)
```
jhendrixMSFT: I like this solution.  Is there a bug tracking fixing this in the generator?

catalinaperalta: I added it as part of my general list of bugs that I've found while doing the tests but I'll also add it's own specific one. 

Also for some reason both the DNS and subscriptionID param were not added. I wonder if it has to do with how they're referenced in the swagger.

catalinaperalta: Here's the issue: #263

jhendrixMSFT: For the missing parameters, my guess is they're defined as `client` which in track 1 meant they were parameters on the client constructor.  For the internal package we'll need to add them in the regular list of params.  And for the convenience layer we'll likely need to add them as constructor params.

catalinaperalta: Ah probably. I'm going to add your last two sentences as comments in the issue.

PR #264, opened 2020-02-15
initial support for query parameters

jhendrixMSFT:
Note that the path* operation groups have been omitted from the package as the generated code doesn't compile.
Not all of the tests have been implemented, and a few being added are skipped due to failures that need investigation.


jhendrixMSFT:
Issue tracking query params https://github.com/Azure/autorest.go/issues/265

src/generator/protocol/operations.ts
```diff
@@ -83,6 +83,70 @@ class methodSig implements MethodSig {
...
+      return `"${constSchema.value.value}"`;
+    case SchemaType.Date:
+    case SchemaType.DateTime:
+      return `${param.language.go!.name}.String()`;
+    case SchemaType.Integer:
+      imports.add('strconv');
+      const intSchema = <NumberSchema>param.schema;
+      let intParam = param.language.go!.name;
+      if (intSchema.precision === 32) {
+        intParam = `int64(${intParam})`;
```
catalinaperalta: Just checking, why is this int64 and not int32?

src/generator/protocol/operations.ts
```diff
@@ -83,6 +83,70 @@ class methodSig implements MethodSig {
...
+      }
+      return `strconv.FormatInt(${intParam}, 10)`;
+    case SchemaType.Number:
+      imports.add('strconv');
+      const numberSchema = <NumberSchema>param.schema;
+      let floatParam = param.language.go!.name;
+      if (numberSchema.precision === 32) {
+        floatParam = `float64(${floatParam})`;
```
catalinaperalta: Same question as the int question. Why float64 and not float32?

src/generator/protocol/operations.ts
jhendrixMSFT: `strconv.FormatInt()` takes an `int64` as its first parameter so we have to up-cast in this case (same is true for `FormatFloat` below).

catalinaperalta: Ah I see, thanks for the explanation!

test/autorest/urlgroup/queries_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func TestArrayStringCsvValid(t *testing.T) {
+	client := getQueriesClient(t)
+	result, err := client.ArrayStringCsvValid(context.Background(), []string{
+		"ArrayQuery1", "begin!*'();:@ &=+$,/?#[]end", "", "",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if result.StatusCode != http.StatusOK {
```
catalinaperalta: This is just a question, why did you switch from using the deep equals func to just checking the status code? It's the same result for the successful status codes, so it doesnt really matter, but just wondering why you preferred this.

jhendrixMSFT: It was shorter to write since there is no response body to compare, just the status code.

PR #266, opened 2020-02-18
Move generated enum content to an enums.go file

jhendrixMSFT:
Also includes the `ToPtr()` method codegen.


test/autorest/generated/stringgroup/internal/stringgroup/enum.go
```diff
@@ -115,4 +115,3 @@ func (EnumOperations) PutReferencedConstantHandleResponse(resp *azcore.Response)
 	}
 	return &EnumPutReferencedConstantResponse{StatusCode: resp.StatusCode}, nil
 }
```
catalinaperalta: The linter is going to complain about this isnt it? (The missing endline)

test/autorest/generated/stringgroup/internal/stringgroup/string.go
```diff
@@ -228,4 +228,3 @@ func (StringOperations) PutWhitespaceHandleResponse(resp *azcore.Response) (*Str
 	}
 	return &StringPutWhitespaceResponse{StatusCode: resp.StatusCode}, nil
 }
```
catalinaperalta: Same endline problem

test/autorest/generated/stringgroup/models.go
```diff
@@ -79,4 +67,3 @@ type StringPutNullResponse = azinternal.StringPutNullResponse
 
 // StringPutWhitespaceResponse contains the response from method String.PutWhitespace.
 type StringPutWhitespaceResponse = azinternal.StringPutWhitespaceResponse
-
```
catalinaperalta: Endline missing.

test/autorest/generated/stringgroup/internal/stringgroup/enum.go
jhendrixMSFT: There were two `\n` characters at the end, the extra one was removed.

catalinaperalta: Do we want to leave it like that?

jhendrixMSFT: It's just an artifact of the codegen.  While we could fix it in the code generator to not emit the extra `\n', since we're going to run `gofmt` anyways I think it's easier to let the formatting tool fix it.

catalinaperalta: Ok sounds good 👍

PR #267, opened 2020-02-18
Initial work for adding headers to requests

catalinaperalta:
Adding custom header generation and one initial test. 
This also adds generated headergroup files. 

NOTE: this does not consider special formats for dates and time


src/generator/protocol/operations.ts
```diff
@@ -100,22 +100,34 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
+      text += `\tif err != nil {\n`;
+      text += `\t\treturn nil, err\n`;
+      text += `\t}\n`;
+    }
+    // add specific request headers and import the fmt package only if there are headers to add
+    if (headerParamCount > 0) {
+      imports.add("fmt");
+      const headerParam = values(op.request.parameters).where((each: Parameter) => { return each.protocol.http!.in === 'header'; });
+      headerParam.forEach(header => {
+        text += `\treq.Header.Set("${header.language.go!.name}", fmt.Sprintf("%v", ${header.language.go!.name}))\n`;
```
jhendrixMSFT: I think you can use `formatQueryParamValue()` (and rename it) instead of the `Sprintf` here since we already know the parameter type (plus it will be necessary in cases where we need to format the time in a specific way).

catalinaperalta: Ok will do, but does formatQueryParamValue already exist? I couldn't find it doing a quick search right now, aside from formatConstantValue

jhendrixMSFT: Yeah it was merged in as part of my query params work.

catalinaperalta: Ah cool, will do 👍

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -0,0 +1,488 @@
...
+	"net/url"
+	"path"
+	"strconv"
+	"time"
+)
+
+type HeaderOperations struct{}
+
+// CustomRequestIDCreateRequest creates the CustomRequestID request.
+func (HeaderOperations) CustomRequestIDCreateRequest(u url.URL) (*azcore.Request, error) {
```
jhendrixMSFT: Is this missing a param?

catalinaperalta: That's the one with the special params that aren't specified in the swagger. Wasn't sure when to add to those cases.

jhendrixMSFT: Ah ok.  We can work up a design for that case a little later, can you please open an issue to track it?

catalinaperalta: Yep, here it is: #268

PR #269, opened 2020-02-19
Adding header tests 

catalinaperalta:
Adding headergroup tests and manual updates to generated headergroup code to make some endpoints work. 

Adding pending bug fixes in this issue: #270

### DIFF
--- a/test/autorest/headergroup/headergroup_test.go
+++ b/test/autorest/headergroup/headergroup_test.go
@@ -252,3 +252,313 @@ func TestHeaderParamLong(t *testing.T) {
 	}
 	deepEqualOrFatal(t, result, expected)
 }
+
+func TestHeaderParamProtectedKey(t *testing.T) {
+	client := getHeaderClient(t)
+	result, err := client.ParamProtectedKey(context.Background(), "text/html")
+	if err != nil {
+		t.Fatalf("ParamProtectedKey: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


test/autorest/headergroup/headergroup_test.go
```diff
@@ -39,12 +39,526 @@ func deepEqualOrFatal(t *testing.T, result interface{}, expected interface{}) {
...
+	deepEqualOrFatal(t, result, expected)
+}
+
+func TestHeaderResponseDatetime(t *testing.T) {
+	client := getHeaderClient(t)
+	result, err := client.ResponseDatetime(context.Background(), "valid")
+	if err != nil {
+		t.Fatalf("ResponseDatetime: %v", err)
+	}
+	expected := &headergroup.HeaderResponseDatetimeResponse{
```
jhendrixMSFT: The models for operations with header responses should have them baked in.  E.g. based on the swagger, `HeaderResponseDatetimeResponse` should have a field named `Value` that's a `time.Time`.

jhendrixMSFT: Maybe this is already tracked in the work item you opened the other day.

catalinaperalta: Yes I agree it should have them baked it, I dont know why it's not happening here. But it is listed the #270

Issue #263, opened 2020-02-14
Fix morecustombaseurigroup by adding missing params and fixing URL


catalinaperalta:
Need to make sure subscriptionID is added to the generated code and that subscriptionID and keyName are added to the URL path, along with setting the keyVersion query param.

catalinaperalta:
For the parameters defined in the morecustombaseurigroup, depending on if they are defined as client then we'll do the following: For the internal package we'll need to add them in the regular list of params. And for the convenience layer we'll likely need to add them as constructor params.

catalinaperalta:
Resolved.

PR #271, opened 2020-02-19
Removing noByRef func so that all types are pointers

catalinaperalta:


catalinaperalta:
urlgroup fixes needed are here: #272

catalinaperalta:
morecustombaseurigroup issue for tracking fixes is here: #263

Issue #270, opened 2020-02-19
Add support for duration format


catalinaperalta:
~1. ParamByte needs to use base64.STDEncoding.EncodeToString instead of just string cast~
~2. ParamExistingKey User-Agent header has been parsed and it shouldnt be, however even when the header is parsed correctly the request still fails.~
~3. ParamEnum doesn't pass in the null scenario~
4. All the time, date and duration TESTS and generated code PARSING need to be looked at, they are not being sent in the correct format. 
~5. ParamProtectedKey the contentType header is not being generated correctly since it was renamed.~
6. ParamString null scenario is broken
~7. All the response endpoints dont marshal the value returned into the response structs~

catalinaperalta:
Bug 1, 2, 5 are resolved in this PR: #273

jhendrixMSFT:
Repurposing this issue to track adding support for duration format (it was the only thing remaining for headergroup and is applicable to models as well).

catalinaperalta:
As per offline discussion, durations are being left as strings.

PR #273, opened 2020-02-20
Fixing headergroup bugs

catalinaperalta:
This PR updates the way that byte arrays are parsed to strings in the generated code and also fixes the value that's passed in as the header key when setting specific request headers, the generator now uses the original value passed in through the swagger. 

Also updated the corresponding headergroup tests and adding generated code. 

The issue tracking pending fixes is: #270

### DIFF
--- a/src/generator/protocol/operations.ts
+++ b/src/generator/protocol/operations.ts
@@ -309,17 +309,17 @@ function createProtocolResponse(client: string, op: Operation, imports: ImportMa
 
   const resp = op.responses![0];
   let respObj = `${resp.language.go!.name}{RawResponse: resp.Response}`;
-  let headResp = <HeaderResponse>{}
+  let headResp = <HeaderResponse>{};
   // check if the response is expecting information from headers
   if (resp.protocol.http!.headers) {
     for (const header of values(resp.protocol.http!.headers)) {
       const head = <LanguageHeader>header;
-      headResp = formatHeaderResponseValue(head, imports)
+      headResp = formatHeaderResponseValue(head, imports);
       // reassign respObj to include the value returned from the headers
       respObj = `${resp.language.go!.name}${headResp.respObj}`;
       // add the code necessary to process data returned in a header
       if (headResp.body) {
-        text += headResp.body
+        text += headResp.body;
       }
     }
   }
--- a/src/generator/protocol/operations.ts
+++ b/src/generator/protocol/operations.ts
@@ -155,7 +155,10 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
 }
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(header: LanguageHeader, imports: ImportManager): HeaderResponse {
+function formatHeaderResponseValue(header: LanguageHeader, imports: ImportManager, respObj: string): HeaderResponse {
+  if (respObj[respObj.length-1] == '}') {
+    respObj = respObj.substring(0, respObj.length-1);
+  }
   let headerText = <HeaderResponse>{};
   let text = ``;
   switch (header.schema.type) {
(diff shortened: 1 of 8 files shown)


test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -313,7 +318,11 @@ func (HeaderOperations) ResponseDateHandleResponse(resp *azcore.Response) (*Head
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	return &HeaderResponseDateResponse{StatusCode: resp.StatusCode}, nil
+	val, err := time.Parse(time.RFC3339, resp.Header.Get("value"))
```
catalinaperalta: This isn't working properly yet. The issue is tracked in #270. I'm taking a look at it.

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -329,7 +338,11 @@ func (HeaderOperations) ResponseDatetimeHandleResponse(resp *azcore.Response) (*
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	return &HeaderResponseDatetimeResponse{StatusCode: resp.StatusCode}, nil
+	val, err := time.Parse(time.RFC3339, resp.Header.Get("value"))
```
catalinaperalta: Same parse issue

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -345,7 +358,11 @@ func (HeaderOperations) ResponseDatetimeRFC1123HandleResponse(resp *azcore.Respo
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	return &HeaderResponseDatetimeRFC1123Response{StatusCode: resp.StatusCode}, nil
+	val, err := time.Parse(time.RFC3339, resp.Header.Get("value"))
```
catalinaperalta: Same parse issue.

catalinaperalta: It compiles it just doesnt know how to respond to data returned from the endpoint.

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -377,7 +398,11 @@ func (HeaderOperations) ResponseDurationHandleResponse(resp *azcore.Response) (*
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	return &HeaderResponseDurationResponse{StatusCode: resp.StatusCode}, nil
+	val, err := time.ParseDuration(resp.Header.Get("value"))
```
catalinaperalta: Same parse issue.

src/generator/protocol/operations.ts
```diff
@@ -196,15 +285,15 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
     // add specific request headers
     const headerParam = values(op.request.parameters).where((each: Parameter) => { return each.protocol.http!.in === 'header'; });
     headerParam.forEach(header => {
-      text += `\treq.Header.Set("${header.language.go!.name}", ${formatParamValue(header, imports)})\n`;
+      text += `\treq.Header.Set("${header.language.default.name}", ${formatParamValue(header, imports)})\n`;
```
jhendrixMSFT: Might want to add a comment why the default name is used here.

test/autorest/generated/headergroup/internal/headergroup/header.go
jhendrixMSFT: Try `time.RFC1123Z`, or possibly the variant without the Z suffix.

src/generator/protocol/operations.ts
```diff
@@ -148,6 +154,89 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
...
+  let text = ``;
+  switch (header.schema.type) {
+    case SchemaType.Boolean:
+      imports.add('strconv');
+      text = `\tval, err := strconv.ParseBool(resp.Header.Get("${header.header}"))\n`;
+      text += `\tif err != nil {\n`;
+      text += `\t\treturn nil, err\n`;
+      text += `\t}\n`;
+      headerText.body = text;
+      headerText.respObj = `{RawResponse: resp.Response, ${header.name}: &val}`;
```
jhendrixMSFT: The "RawResponse: resp.Response" part is declared by the caller.  I suggest making this part a parameter to this function so it's only defined once by the caller.

test/autorest/generated/headergroup/internal/headergroup/header.go
catalinaperalta: Yup I just figured out the fix, but I think I'll leave it for the next PR where I fix all the time/date related issues.

src/generator/protocol/operations.ts
catalinaperalta: Done

src/generator/protocol/operations.ts
catalinaperalta: Ok added it and passed it into the function as a param

test/autorest/generated/headergroup/internal/headergroup/models.go
```diff
@@ -39,175 +41,189 @@ func (e Error) Error() string {
...
 }
 
 // HeaderResponseExistingKeyResponse contains the response from method Header.ResponseExistingKey.
 type HeaderResponseExistingKeyResponse struct {
-	// StatusCode contains the HTTP status code.
-	StatusCode int
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+	UserAgent *string
 }
```
jhendrixMSFT: You will also want to add any description text as a comment above the header value.  If the description is empty we can put some more generic text like "UserAgent contains the value from the User-Agent header response." or something like that.

jhendrixMSFT: And include a new-line between entries.

catalinaperalta: Ok done.

PR #274, opened 2020-02-20
Replace StatusCode with RawResponse

jhendrixMSFT:
Response types now include the *http.Response instead of just the status
code, making response headers etc available.
Added helpers package to consolidate test helpers.


src/generator/protocol/operations.ts
```diff
@@ -219,7 +219,7 @@ function createProtocolResponse(client: string, op: Operation): string {
   text += '\t}\n';
 
   const resp = op.responses![0];
-  const respObj = `${resp.language.go!.name}{StatusCode: resp.StatusCode}`;
+  const respObj = `${resp.language.go!.name}{RawResponse: resp.Response}`;
```
catalinaperalta: Just double checking since most fields will be pointer to type this doesnt need to be &resp.Response?

jhendrixMSFT: The `http.Response` wrapped in `azcore.Response` is already a pointer (same goes for requests).

catalinaperalta: Ah you're right, thanks for the info!

PR #277, opened 2020-02-24
Group optional parameters into their own Options struct

jhendrixMSFT:
For operations with optional parameters (i.e. not marked as required:
true in the swagger) create a type <OperationGroup><Operation>Options
that contains all of the optional args.  This type will be passed by
pointer as the last method argument.

### DIFF
--- a/src/generator/convenience/client.ts
+++ b/src/generator/convenience/client.ts
@@ -7,7 +7,7 @@ import { Session } from '@azure-tools/autorest-extension-base';
 import { camelCase } from '@azure-tools/codegen';
 import { CodeModel, Parameter } from '@azure-tools/codemodel';
 import { values } from '@azure-tools/linq';
-import { ContentPreamble, ImportManager, ParamInfo } from '../common/helpers';
+import { ContentPreamble, formatParamInfoTypeName, ImportManager, ParamInfo, sortParamInfoByRequired } from '../common/helpers';
 
 // generates content for client.go
 export async function generateClient(session: Session<CodeModel>): Promise<string> {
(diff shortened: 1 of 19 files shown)


src/generator/convenience/client.ts
```diff
@@ -7,7 +7,7 @@ import { Session } from '@azure-tools/autorest-extension-base';
 import { camelCase } from '@azure-tools/codegen';
 import { CodeModel, Parameter } from '@azure-tools/codemodel';
 import { values } from '@azure-tools/linq';
-import { ContentPreamble, ImportManager, ParamInfo } from '../common/helpers';
+import { ContentPreamble, formatParamInfoTypeName, ImportManager, ParamInfo, sortParamInfoByRequired } from '../common/helpers';
```
catalinaperalta: So this is a format convention question for me to keep in mind. Are we leaving functions uncapitalized? and interfaces/classes etc capitalized?

jhendrixMSFT: I asked Garrett about this, his response was.
> type/interface/class names are PascalCased
variables, methods, properties, functions are camelCased

So I've been trying this (sorry I forgot to bring this up earlier).

src/generator/protocol/operations.ts
```diff
@@ -127,40 +137,44 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
     case SchemaType.DateTime:
     case SchemaType.Duration:
     case SchemaType.UnixTime:
-      return `${param.language.go!.name}.String()`;
+      if (param.required !== true) {
+        // remove the dereference (bit of a hack but this is the only case where it's necessary)
+        paramName = paramName.substr(1);
```
catalinaperalta: If we want it to be a little less hacky we could add a check to see if the dereference * exists on the string and then perform the substr func. Not something I would insist on, but I did something similar when parsing the response structs to remove the trailing }.

src/generator/convenience/client.ts
catalinaperalta: Ah ok good to know! Thanks for the info

src/generator/protocol/operations.ts
catalinaperalta: I was thinking about it...I think it might be safer to add the check. Just in case we were ever to pass something that doesnt include the deref

jhendrixMSFT: Good point, added.

PR #278, opened 2020-02-26
Adding date and datetime parsing functionality to requests and responses

catalinaperalta:
This PR also includes a bug fix for when the description is not included in the Language object passed into HasDescription.

### DIFF
--- a/src/generator/protocol/operations.ts
+++ b/src/generator/protocol/operations.ts
@@ -93,9 +93,9 @@ export interface HeaderFormat extends Schema {
 }
 
 function formatParamValue(param: Parameter, imports: ImportManager): string {
-  const dateFormat = "2006-01-02";
-  const datetimeFormat = "time.RFC3339";
-  const datetimeRFC1123Format = "time.RFC1123";
+  const dateFormat = '2006-01-02';
+  const datetimeFormat = 'time.RFC3339';
+  const datetimeRFC1123Format = 'time.RFC1123';
 
   let separator = ',';
   switch (param.protocol.http?.style) {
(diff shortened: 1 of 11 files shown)


src/generator/protocol/operations.ts
```diff
@@ -140,7 +148,15 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
       // cannot use formatConstantValue() since all values are treated as strings
       return `"${constSchema.value.value}"`;
     case SchemaType.Date:
+      return `${paramName}.Format("${dateFormat}")`;
     case SchemaType.DateTime:
+      if (paramName[0] == '*') {
+        paramName = paramName.substr(1);
+      }
+      if ((<HeaderFormat>param.schema).format === 'date-time-rfc1123') {
```
jhendrixMSFT: Minor issue, there should be an existing definition for `date-time-rfc1123` in autorest.core someplace.

src/generator/protocol/operations.ts
```diff
@@ -171,8 +187,11 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
 
 // use this to generate the code that will help process values returned in response headers
 function formatHeaderResponseValue(header: LanguageHeader, imports: ImportManager, respObj: string): HeaderResponse {
-  if (respObj[respObj.length-1] == '}') {
-    respObj = respObj.substring(0, respObj.length-1);
+  const dateFormat = '2006-01-02';
+  const datetimeFormat = 'time.RFC3339';
+  const datetimeRFC1123Format = 'time.RFC1123';
```
jhendrixMSFT: Looks like the same constants are defined earlier, you could probably move them to the file-level and share them.

src/generator/protocol/operations.ts
catalinaperalta: I did try to find it in the schema types before, but I'll take a look elsewhere for it then.

catalinaperalta: So this is what I've found so far in reference to the alternate datetime format: 
```
/** a schema that represents a DateTime value */
export interface DateTimeSchema extends PrimitiveSchema {
    /** the schema type  */
    type: SchemaType.DateTime;
    /** date-time format  */
    format: 'date-time-rfc1123' | 'date-time';
}
```

catalinaperalta: Doesnt seem like it has its own specific interface, I also did a search in the node_modules folder to find any other reference to it and it seems that that is the only one.

PR #279, opened 2020-02-27
Added filegroup

jhendrixMSFT:
Disable auto-downloading of the HTTP response body for operations that
return a binary stream.
Simplified adding request headers.
Fix formatting of struct fields without comments.
Deleted a duplicate `additionalproperties` group.


test/autorest/generated/filegroup/internal/filegroup/models.go
```diff
@@ -0,0 +1,58 @@
...
+package filegroup
+
+import (
+	"fmt"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"net/http"
+)
+
+type Error struct {
+	Message *string `json:"message,omitempty"`
```
catalinaperalta: Shouldnt these have comments?

jhendrixMSFT: Yes however they're missing from the swagger :(

PR #282, opened 2020-02-28
modelerfour clean-up

jhendrixMSFT:
Recent changes in modelerfour split request parameters into two groups,
one at the operation level and one at the request level; this is to
support operations with multiple requests.  Aggregate parameter groups
into a single array when enumerating request parameters.
Fixed a few minor bugs in formatParamValue().
Skip constants when generating an optional params struct.
Added transformations for enum types.
Fixed the regeneration script to format all generated content, and
commented out test groups that are NYI.


rushScripts/regeneration.js
```diff
@@ -5,70 +5,75 @@ var exec = require('child_process').exec;
...
-    'numbergroup':['body-number.json'],
-    'optionalgroup':['required-optional.json'],
-    'paginggroup':['paging.json'],
-    'reportgroup':['report.json'],
-    'stringgroup':['body-string.json'],
-    'urlgroup':['url.json'],
-    'urlmultigroup':['url-multi-collectionFormat.json'],
-    'validationgroup':['validation.json'],
-    'xmlgroup':['xml-service.json'],
+    'additionalpropertiesgroup': 'additionalProperties.json',
```
catalinaperalta: I'm thinking we should also comment this one out for now. We don't really have tests for it yet and it's not as stable. (We haven't gotten any errors because there isn't test coverage for it yet.)

test/autorest/generated/stringgroup/enums.go
```diff
@@ -10,12 +10,11 @@ import azinternal "generatortests/autorest/generated/stringgroup/internal/string
 type Colors = azinternal.Colors
 
 const (
-	ColorsRedColor = azinternal.ColorsRedColor
-	ColorsGreenColor = azinternal.ColorsGreenColor
-	ColorsBlueColor = azinternal.ColorsBlueColor
+	ColorsRedcolor   = azinternal.ColorsRedcolor
```
jhendrixMSFT: Unclear what caused this change, regression in the modeler?

jhendrixMSFT: It's due to the following line in the namer.
```ts
details.name = `${enm.language.go?.name}${capitalizeAcronyms(pascalCase(details.name.toLowerCase()))}`;
```
Not sure why this is showing up now but I think we should remove the call to `toLowerCase()`.  What do you think?

test/autorest/generated/additionalpropertiesgroup/internal/additionalpropertiesgroup/models.go
```diff
@@ -42,79 +43,84 @@ func (e Error) Error() string {
...
-	Name *string `json:"name,omitempty"`
-	Status *bool `json:"status,omitempty"`
+	ID     *int32  `json:"id,omitempty"`
+	Name   *string `json:"name,omitempty"`
+	Status *bool   `json:"status,omitempty"`
 }
 
 // PetsCreateApInPropertiesResponse contains the response from method Pets.CreateApInProperties.
 type PetsCreateApInPropertiesResponse struct {
 	PetApInProperties *PetApInProperties
```
catalinaperalta: Should this have comment over it?

test/autorest/generated/stringgroup/enums.go
catalinaperalta: Yup agreed! Will you remove it? Or should I make a note of it?

jhendrixMSFT: I just pushed the fix.

catalinaperalta: Now I'm just not sure it all of the groups need to be regenerated. I saw the the generated output for urlgroup still has color as lower case.

test/autorest/generated/additionalpropertiesgroup/internal/additionalpropertiesgroup/models.go
jhendrixMSFT: So there is a description defined in the swagger.
```json
          "200": {
            "description": "A Pet which contains more properties than what is defined.",
            "schema": {
              "$ref": "#/definitions/PetAPInProperties"
            }
          },
```
However I don't see this in the model.  I'll follow up about that.

PR #283, opened 2020-02-29
Add support for XML marshalling/unmarshalling

jhendrixMSFT:


src/generator/protocol/operations.ts
```diff
@@ -353,6 +353,29 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
+        if (isArraySchema(bodyParam!.schema)) {
+          fieldName = pascalCase(bodyParam!.language.go!.name);
+          let tag = bodyParam!.schema.elementType.language.go!.name;
+          if (bodyParam!.schema.elementType.serialization?.xml?.name) {
+            tag = bodyParam!.schema.elementType.serialization.xml.name;
+          }
+          text += `\t\t${fieldName} *${bodyParam!.schema.language.go!.name} \`xml:"${tag}"\`\n`;
+        } else {
+          // embed as anonymous field
+          text += `\t\t*${bodyParam!.schema.language.go!.name}\n`;
```
catalinaperalta: Just trying to understand this a little more, what cases do we embed this anonymously?

jhendrixMSFT: If it's an object then it's embedded anonymously (there are only two cases, arrays or objects).

test/autorest/generated/urlgroup/internal/urlgroup/paths.go
```diff
@@ -96,7 +96,7 @@ func (PathsOperations) ByteNullCreateRequest(u url.URL, bytePath []byte) (*azcor
 
 // ByteNullHandleResponse handles the ByteNull response.
 func (PathsOperations) ByteNullHandleResponse(resp *azcore.Response) (*PathsByteNullResponse, error) {
-	if !resp.HasStatusCode(http.StatusOK) {
+	if !resp.HasStatusCode(http.StatusBadRequest) {
```
catalinaperalta: We're expecting null scenarios to return a BadRequest status?

src/generator/protocol/operations.ts
catalinaperalta: Ah ok, good to know. Thanks!

test/autorest/generated/urlgroup/internal/urlgroup/paths.go
catalinaperalta: I know this comes from the swagger, but I'm wondering what changed. Also, shouldnt it be without the ! for bad requests? That would be an error wouldnt it?

jhendrixMSFT: The way it's worded in the test, the operation returns invalid data that should cause unmarshalling to fail so the API call returns an error.  So if you actually get back the 400 and no error something went wrong elsewhere (it's kinda weird).

jhendrixMSFT: As for what changed, this was fixed via `formatStatusCodes()`.  Before that we had a hard-coded `http.StatusOK`.

catalinaperalta: Hmmm it is very strange. But I guess this is the expected output then. Doesnt sound like real world scenarios should work like this though.

PR #287, opened 2020-03-05
Initial support for time formats

jhendrixMSFT:
This covers scalar time request and response.  Support for time within
structs will be added in a separate change.
Emit internal time helper types when a request and/or response contains
a time type.
Switched header format checks to use DateTimeSchema.


.vscode/settings.json
```diff
@@ -64,5 +64,11 @@
...
       "directory": "./src",
     },
-  ]
+  ],
+  "go.useLanguageServer": true,
+  "go.languageServerExperimentalFeatures": {
+    "format": true,
+    "diagnostics": true,
+    "documentLink": true
+  }
```
catalinaperalta: Just wondering....shouldn't this file be in our gitignore?

jhendrixMSFT: Don't we use this to enable common eslint features?

catalinaperalta: Hmmm...well yes we could. I havent pushed my recent changes to this file though. But I could do it as well. So then we'll keep this in the repo until we merge track2 into master? And then delete it before then? Or would be keep it?

jhendrixMSFT: Well, if this is how we standardize on what eslint features to require then I think we'll have to keep it.

catalinaperalta: Yes that's true. Ok so we're keeping it 👍

PR #289, opened 2020-03-09
Add multiple response functionality

catalinaperalta:
This PR adds code changes that will implement multiple responses for code generation, including related fixes.
Changes implemented:
- Fixing header variable names for parsing in response code
- Adding multiple response parameters and headers to responses
- Adding multiple response functionality for requests
- Adding missing naming changes in the namer, related to headers
- Adding a fix to sealed-choice transforms for sealed choices that are not all of the same schema
- Adding considerations for Schema Types: Binary and URI
- Adding multiple response related fields to corresponding structs


src/generator/common/helpers.ts
```diff
@@ -135,15 +138,14 @@ export function formatParamInfoTypeName(param: ParamInfo): string {
...
-  if (op.requests!.length > 1) {
-    throw console.error('multiple requests NYI');
-  }
   let params = new Array<Parameter>();
   if (op.parameters) {
     params = params.concat(op.parameters);
   }
-  if (op.requests![0].parameters) {
-    params = params.concat(op.requests![0].parameters);
+  for (const req of values(op.requests)) {
```
jhendrixMSFT: I don't think this is what we want to do.  E.g. multiple requests means that the endpoint could take a JSON object or a file stream.  Some languages will convert this into two separate APIs (`PutWidgetWithBody()` and `PutWidgetWithFileStream()`); others will fold the distinct params into a struct (`PutWidget(payload PutWidgetPayload)`), then will send whatever field is set in the payaload.  I think we need to discuss this more in depth to decide which approach we want to take.

src/generator/common/helpers.ts
```diff
@@ -241,11 +240,20 @@ export function getEnums(schemas: Schemas): EnumEntry[] {
     enums.push(entry);
   }
   for (const choice of values(schemas.sealedChoices)) {
-    const entry = new EnumEntry(choice.language.go!.name, choice.choiceType.language.go!.name, choice.language.go!.possibleValuesFunc, choice.choices);
-    if (HasDescription(choice.language.go!)) {
-      entry.desc = choice.language.go!.description;
+    // TODO check what to do in the case of having a sealedChoice that doesnt have the choiceType field
```
jhendrixMSFT: Since our enums are always strings we could hard-code the type and skip checking `choice.choiceType`.  Do you know what enum(s) in the blob storage swagger are missing this?

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -298,11 +298,11 @@ func (HeaderOperations) ResponseBoolHandleResponse(resp *azcore.Response) (*Head
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	val, err := strconv.ParseBool(resp.Header.Get("value"))
+	Value, err := strconv.ParseBool(resp.Header.Get("value"))
```
jhendrixMSFT: `Value` should be camel-cased.

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -319,8 +319,8 @@ func (HeaderOperations) ResponseByteHandleResponse(resp *azcore.Response) (*Head
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	val := []byte(resp.Header.Get("value"))
-	return &HeaderResponseByteResponse{RawResponse: resp.Response, Value: &val}, nil
+	Value := []byte(resp.Header.Get("value"))
```
jhendrixMSFT: Isn't this a base-64 encoded string that needs to be decoded?  Can be fixed separately.

test/autorest/generated/headergroup/internal/headergroup/header.go
```diff
@@ -358,11 +358,11 @@ func (HeaderOperations) ResponseDatetimeHandleResponse(resp *azcore.Response) (*
 	if !resp.HasStatusCode(http.StatusOK) {
 		return nil, newError(resp)
 	}
-	val, err := time.Parse(time.RFC3339, resp.Header.Get("value"))
+	Value, err := time.Parse(time.RFC3339, resp.Header.Get("value"))
```
jhendrixMSFT: I think this should be `timeRFC3339Nano`, can be fixed separately.

src/generator/protocol/operations.ts
```diff
@@ -307,16 +314,17 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
   let text = `${comment(name, '// ')} creates the ${info.name} request.\n`;
   text += `func (${client}) ${name}(${generateParamsSig(sig.protocolSigs.requestMethod.params, true)}) (${sig.protocolSigs.requestMethod.returns.join(', ')}) {\n`;
   text += `\turlPath := "${op.requests![0].protocol.http!.path}"\n`;
-  if (values(aggregateParameters(op)).any((each: Parameter) => { return each.protocol.http!.in === 'path' })) {
-    // replace path parameters
+  const inPathParams = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined; }).where((each: Parameter) => { return each.protocol.http!.in === 'path'; });
+  // replace path parameters
+  for (const pp of values(inPathParams)) {
+    // TODO this needs to be fixed, this is an ugly solution to guarantee that string and url imports are only added when an in path parameter is actually added
```
jhendrixMSFT: I assume this TODO is about repeated calls to `imports.add()`?  You could move this out of the `for` loop, under something like `if (inPathParams.length >0) then add imports`.

src/generator/protocol/operations.ts
```diff
@@ -307,16 +314,17 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
     imports.add('strings');
     imports.add('net/url');
-    for (const pp of values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http!.in === 'path' })) {
-      text += `\turlPath = strings.ReplaceAll(urlPath, "{${pp.language.go!.serializedName}}", url.PathEscape(${formatParamValue(pp, imports)}))\n`;
-    }
+    text += `\turlPath = strings.ReplaceAll(urlPath, "{${pp.language.go!.serializedName}}", url.PathEscape(${formatParamValue(pp, imports)}))\n`;
   }
   text += `\tu.Path = path.Join(u.Path, urlPath)\n`;
-  if (values(aggregateParameters(op)).any((each: Parameter) => { return each.protocol.http!.in === 'query' })) {
+  // const inQueryParams = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http!.in === 'query'; });
```
jhendrixMSFT: Did this not work?  Probably better to get the list of query params once.

src/generator/common/helpers.ts
catalinaperalta: Yes, you're right. I'll remove that and leave it as it was before for now.

catalinaperalta: Actually I'll throw an error if we receive multiple responses so we know

src/generator/common/helpers.ts
catalinaperalta: There's a ContentType enum and it only has one choice.

test/autorest/generated/headergroup/internal/headergroup/header.go
catalinaperalta: Value is capitalized because we're using the name we've parsed for the fields (exported). I could camel case it in place only when we're parsing values. But the value stored in language.go.name is capitalized since it will be used as part of the structs.

src/generator/protocol/operations.ts
catalinaperalta: Yes it's due to the repeated imports and I previously had exactly what you're suggesting but there were failures and edge cases that were causing errors. So I'm digging into this as part of the pending paths fix.

catalinaperalta: For instance in some cases inPathParams.length >0 BUT when we want to iterate over the inPathParams values there's nothing to iterate over, therefore the imports are added but unused.

src/generator/protocol/operations.ts
catalinaperalta: Nope, this is what I reverted to the previous version. It seems to exclude query params even when they should be considered. It's a similar issue to the inPathParams

src/generator/protocol/models.ts
```diff
@@ -18,20 +18,30 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
...
       if (op.responses) {
-        // check if the response has http headers that it will expect information from. 
-        if (op.responses![0].protocol.http!.headers) {
-          for (const header of values(op.responses![0].protocol.http!.headers)) {
-            const head = <LanguageHeader>header;
-            // convert each header to a property and append it to the response properties list
-            if (!HasDescription(head)) {
-              head.description = `${head.name} contains the information returned from the ${head.name} header response.`
+        let firstResp = op.responses![0];
+        let headerArray = new Map<string, LanguageHeader>();
```
jhendrixMSFT: Might be good to add a comment explaining this merging of response headers into one response type.

src/generator/common/helpers.ts
catalinaperalta: Removed

test/autorest/generated/headergroup/internal/headergroup/header.go
catalinaperalta: Done, they're camel-cased now and int32 and float32 vars end with 32 instead of 0

test/autorest/generated/headergroup/internal/headergroup/header.go
catalinaperalta: Done

src/generator/protocol/operations.ts
catalinaperalta: This is done for swaggers that use http.in = path. This does not consider x-ms-paths yet. But did leave the imports outside of the loop. There's an issue here #296  tracking the x-ms-paths functionality.

src/generator/common/helpers.ts
catalinaperalta: Removed it from the sealed-choice transform and helper func that would add it as an enum. They still show up on the model however.

src/generator/protocol/models.ts
catalinaperalta: Done

test/autorest/generated/headergroup/internal/headergroup/header.go
catalinaperalta: Adding a separate issue to track this fix: #298

src/generator/common/helpers.ts
```diff
@@ -135,13 +138,13 @@ export function formatParamInfoTypeName(param: ParamInfo): string {
...
 // aggregates the Parameter in op.parameters and the first request
 export function aggregateParameters(op: Operation): Array<Parameter> {
-  if (op.requests!.length > 1) {
-    throw console.error('multiple requests NYI');
-  }
   let params = new Array<Parameter>();
   if (op.parameters) {
     params = params.concat(op.parameters);
   }
+  if (op.requests!.length > 1) {
```
jhendrixMSFT: Nit: move this block back to line 138 so diff goes away.

catalinaperalta: You mean line 141? outside of that it would be outside of the functions

catalinaperalta: Moved it

src/namer/namer.ts
```diff
@@ -97,7 +97,7 @@ async function process(session: Session<CodeModel>) {
         paramDetails.name = getEscapedReservedName(camelCase(paramDetails.name), 'Parameter');
         // this is a bit of a weird case and might be due to invalid swagger in the test
         // server.  how can you have an optional parameter that's also a constant?
-        if (param.required !== true && param.schema.type !== SchemaType.Constant) {
+        if (param.required !== true && param.schema.type !== SchemaType.Constant && !(param.schema.type === SchemaType.SealedChoice && ((<SealedChoiceSchema>param.schema).choices !== undefined && (<SealedChoiceSchema>param.schema).choices.length === 1))) {
```
jhendrixMSFT: Nit: I think you can get rid of the `((<SealedChoiceSchema>param.schema).choices !== undefined` part as it should never be undefined.

src/transform/transform.ts
```diff
@@ -146,12 +154,23 @@ function recursiveAddMarshallingFormat(schema: Schema, marshallingFormat: 'json'
 function processOperationRequests(session: Session<CodeModel>) {
   for (const group of values(session.model.operationGroups)) {
     for (const op of values(group.operations)) {
+      if (op.requests) {
```
jhendrixMSFT: Nit: Is this check really necessary?

src/namer/namer.ts
catalinaperalta: Yes you're right it's unnecessary. Removing it.

src/transform/transform.ts
catalinaperalta: Without this check it says that op.requests is possibly undefined

catalinaperalta: I'll add op.requests!.length for the first check and remove the if

jhendrixMSFT: Huh that's interesting, I wonder when it would ever be undefined (how can you have an operation with no requests....).  If it ends up being problematic can leave it as-is.

catalinaperalta: I removed it for now, I agree that it's unlikely that it would ever return undefined.

PR #293, opened 2020-03-09
Custom marshalling/unmarshalling for types with time.Time

jhendrixMSFT:
Moved wrapper type for marshalling into custom marshaller method so it
works with custom time marshalling.
Removed ToTime() methods on custom time types (uses casts now).
Use default marshallers for timeRFC3339 type.


src/generator/protocol/models.ts
```diff
@@ -188,6 +190,99 @@ class StructDef {
...
+    }
+    for (const prop of values(this.Properties)) {
+      if (prop.schema.type !== SchemaType.DateTime) {
+        continue;
+      }
+      text += `\t${receiver}.${prop.language.go!.name} = (*time.Time)(aux.${prop.language.go!.name})\n`;
+    }
+    text += '\treturn nil\n';
+    text += '}\n\n';
+    return text
```
catalinaperalta: ;

test/autorest/generated/xmlgroup/internal/xmlgroup/models.go
```diff
@@ -390,14 +511,15 @@ type XMLGetEmptyWrappedListsResponse struct {
 // XMLGetHeadersResponse contains the response from method XML.GetHeaders.
 type XMLGetHeadersResponse struct {
 	// CustomHeader contains the information returned from the CustomHeader header response.
-	CustomHeader *string `xml:"CustomHeader"`
+	CustomHeader *string
```
catalinaperalta: is it right that this doesn't have the XML tag anymore?

jhendrixMSFT: Yes, this value comes from an HTTP response header so no tag is required.

src/generator/protocol/models.ts
jhendrixMSFT: Fixed when rebasing on track2.

PR #305, opened 2020-03-12
Fix optional header values

jhendrixMSFT:
Check for nil before dereferencing optional header values.


jhendrixMSFT:
Fixes part of https://github.com/Azure/autorest.go/issues/270

PR #307, opened 2020-03-13
Adding function to remove prefixes from names

catalinaperalta:


src/namer/namer.ts
```diff
@@ -187,3 +187,20 @@ function getEscapedReservedName(name: string, appendValue: string): string {
...
+  }
+
+  for (var i = 0; i < prefix.length; i++) {
+    if (prefix[i] != name[i]) {
+      return name
+    }
+  }
+
+  return name.slice(prefix.length);
+}
```
jhendrixMSFT: Nit, missing new-line.

PR #308, opened 2020-03-16
Merge convenience layer into protocol layer

jhendrixMSFT:
I merged the code with as few changes as possible, however there is some opportunity for further refinement (e.g. no longer needing to pass global client params).  Will address in a separate PR.


src/generator/generator.ts
```diff
@@ -7,40 +7,45 @@ import { serialize } from '@azure-tools/codegen';
...
-
-    const operations = await generateOperations(session);
     for (const op of values(operations)) {
       host.WriteFile(`${op.name.toLowerCase()}.go`, op.content, undefined, 'source-file-go');
     }
 
     const enums = await generateEnums(session);
     if (enums.length > 0) {
-      host.WriteFile('enums.go', enums, undefined, 'source-file-go');
+      host.WriteFile(`enums.go`, enums, undefined, 'source-file-go');
```
catalinaperalta: Why did you change to these types of quotes? According to me the correct thing is to use ' ' if it's a simple string as above.

src/generator/generator.ts
```diff
@@ -7,40 +7,45 @@ import { serialize } from '@azure-tools/codegen';
...
     }
 
     const models = await generateModels(session);
-    host.WriteFile('models.go', models, undefined, 'source-file-go');
+    host.WriteFile(`models.go`, models, undefined, 'source-file-go');
```
catalinaperalta: Same as my previous comment

src/generator/generator.ts
jhendrixMSFT: It was from copy/paste from the other layer, will fix.

PR #315, opened 2020-03-23
Pagers implementation

jhendrixMSFT:
For pageable operations, generate all pager types in a pagers.go file.


src/generator/pagers.ts
```diff
@@ -0,0 +1,96 @@
...
+	// Err returns the last error encountered while paging.
+	Err() error
+}
+
+type ${responderType} func(*azcore.Response) (*${responseType}, error)
+
+type ${advanceType} func(*${responseType}) (*azcore.Request, error)
+
+type ${pagerType} struct {
+	cli *${pager.client}
```
catalinaperalta: Minor observation: is cli the best name for this field? Would cl or ct be better? Just because cli actually has another meaning strongly associated with it.

src/generator/pagers.ts
```diff
@@ -0,0 +1,96 @@
...
+
+	// Err returns the last error encountered while paging.
+	Err() error
+}
+
+type ${responderType} func(*azcore.Response) (*${responseType}, error)
+
+type ${advanceType} func(*${responseType}) (*azcore.Request, error)
+
+type ${pagerType} struct {
```
catalinaperalta: I think it would be good if each of the fields on this type has  a comment related to its purpose

src/generator/pagers.ts
jhendrixMSFT: Fixed.

src/generator/pagers.ts
jhendrixMSFT: Fixed.

PR #316, opened 2020-03-23
Publish AutoRest test server coverage report in CI

jhendrixMSFT:
Added test/autorest/covereport/main.go to generate the scenario coverage
report; this is published via the added step in ci.yml.
Updated generator version to v4.0.0 as v3.x.y is already in use.


jhendrixMSFT:
@daviwil is there anything else I need here to get this to work?  I'm not seeing any results on the AutoRest dashboard.

jhendrixMSFT:
Hmm so the publish command fails, it isn't resolving the value for `$(storage-coverage-pass)`.  So there are two issues here; the step doesn't fail when the publish command does, and why doesn't the var $(storage-coverage-pass) resolve.

jhendrixMSFT:
@daviwil do you have any suggestions on how to fix the value for `$(storage-coverage-pass)`?  I must be missing something.

jhendrixMSFT:
Secrets don't resolve on PR builds from forks so this is expected.

PR #318, opened 2020-03-24
Adding options prefix for body params

catalinaperalta:


test/autorest/stringgroup/string_test.go
```diff
@@ -120,21 +120,21 @@ func TestStringGetWhitespace(t *testing.T) {
...
 
-// func TestStringPutNull(t *testing.T) {
-// 	client := getStringClient(t)
-// 	result, err := client.PutNull(context.Background())
-// 	if err != nil {
-// 		t.Fatalf("PutNull: %v", err)
-// 	}
-// 	helpers.VerifyStatusCode(t, result, http.StatusOK)
-// }
+func TestStringPutNull(t *testing.T) {
```
jhendrixMSFT: Just FYI I enabled this test.

src/generator/operations.ts
```diff
@@ -427,10 +427,15 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
   const mediaType = getMediaType(op.requests![0].protocol);
   if (mediaType === 'JSON' || mediaType === 'XML') {
     const bodyParam = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http!.in === 'body'; }).first();
+    // adding this variable to control whether a 'options.' needs to be added before optional body parameters
+    let setOptionsPrefix = false;
```
jhendrixMSFT: I think this var can be removed, then down on line 478 say `if (!bodyParam!.required)`

jhendrixMSFT: Oh hmm maybe not, since it's not set to `true` in certain cases.

src/generator/operations.ts
```diff
@@ -427,10 +427,15 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
     // default to the body param name
     let body = bodyParam!.language.go!.name;
     if (bodyParam!.schema.type === SchemaType.Constant) {
       // if the value is constant, embed it directly
+      if (!bodyParam!.required) {
```
jhendrixMSFT: Nit: Can be replaced with `setOptionsPrefix = !bodyParam!.required` to eliminate nested `if` block.

src/generator/operations.ts
```diff
@@ -451,16 +456,39 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
-      text += `\taux := ${bodyParam!.schema.language.go!.internalTimeType}(${body})\n`;
+      if (!bodyParam!.required) {
+        text += `\taux := ${bodyParam!.schema.language.go!.internalTimeType}(options.${pascalCase(body)})\n`;
+      } else {
+        text += `\taux := ${bodyParam!.schema.language.go!.internalTimeType}(${body})\n`;
+      }
       body = 'aux';
+    } else {
+      // final check for each body param to know if it is optional
+      if (!bodyParam!.required) {
```
jhendrixMSFT: Nit: Can be replaced with `setOptionsPrefix = !bodyParam!.required` to eliminate nested `if` block.

PR #319, opened 2020-03-24
Remove XMS prefix from header response fields

jhendrixMSFT:
Added Crc to list of TLAs for capitalization.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -204,10 +204,10 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
 }
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(header: string, schema: Schema, imports: ImportManager, respObj: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string): string {
   let text = '';
   let needsErrorCheck = true;
-  const name = camelCase(header);
+  const name = camelCase(propName);
   switch (schema.type) {
     case SchemaType.Boolean:
       imports.add('strconv');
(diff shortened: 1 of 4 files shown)


src/namer/namer.ts
```diff
@@ -79,7 +79,6 @@ async function process(session: Session<CodeModel>) {
   // pascal-case and capitzalize acronym operation groups and their operations
   for (const group of values(model.operationGroups)) {
     const details = <Language>group.language.go;
-    const opGroupName = capitalizeAcronyms(pascalCase(group.$key));
```
catalinaperalta: Why dont we want to have this naming fix anymore?

jhendrixMSFT: It's dead code, but you bring up a good point that maybe it shouldn't be.  Let me take a look.

src/generator/operations.ts
```diff
@@ -204,10 +204,10 @@ function formatParamValue(param: Parameter, imports: ImportManager): string {
 }
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(header: string, schema: Schema, imports: ImportManager, respObj: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string): string {
```
catalinaperalta: What's the difference between propName and header?

jhendrixMSFT: `propName` is the name of the property as it appears in the struct.  `header` is the actual header, e.g. `x-ms-foo`

catalinaperalta: Ah I see. Thanks for the explanation!

src/namer/namer.ts
jhendrixMSFT: The reference to it was removed in https://github.com/Azure/autorest.go/pull/313.  Apparently I forgot to remove the definition then.  :(

PR #320, opened 2020-03-24
Workaround for non-required constants

jhendrixMSFT:
Omit non-required constants from codegen for now.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -397,6 +397,7 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
         text += `\tquery.Set("${qp.language.go!.serializedName}", ${formatParamValue(qp, imports)})\n`;
       } else if (qp.schema.type === SchemaType.Constant) {
         // omit this query param. TODO once non-required constants are fixed
+        console.warn(`omitting non-required constant query parameter ${qp.language.go!.serializedName}`);
       } else if (qp.implementation === ImplementationLocation.Client) {
         // global optional param
         text += `\tif ${qp.language.go!.name} != nil {\n`;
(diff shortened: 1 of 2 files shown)


src/generator/operations.ts
```diff
@@ -395,6 +395,8 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
     for (const qp of values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http!.in === 'query'; })) {
       if (qp.required === true) {
         text += `\tquery.Set("${qp.language.go!.serializedName}", ${formatParamValue(qp, imports)})\n`;
+      } else if (qp.schema.type === SchemaType.Constant) {
+        // omit this query param. TODO once non-required constants are fixed
```
catalinaperalta: Should we print something to the console so we know this isn't being implemented yet?

src/generator/operations.ts
```diff
@@ -418,6 +420,8 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
   headerParam.forEach(header => {
     if (header.required) {
       text += `\treq.Header.Set("${header.language.go!.serializedName}", ${formatParamValue(header, imports)})\n`;
+    } else if (header.schema.type === SchemaType.Constant) {
+      // omit this header. TODO once non-required constants are fixed
```
catalinaperalta: Same here

src/generator/operations.ts
jhendrixMSFT: Added

src/generator/operations.ts
jhendrixMSFT: Added

PR #321, opened 2020-03-24
Adding url path parsing depending on path items

catalinaperalta:
Adding url path based on whether path items are present and fixing err variable assignment on body params.


test/storage/2019-07-07/azblob/service.go
```diff
@@ -57,7 +56,7 @@ func (client *serviceOperations) GetAccountInfo(ctx context.Context) (*ServiceGe
 }
 
 // getAccountInfoCreateRequest creates the GetAccountInfo request.
-func (client *serviceOperations) getAccountInfoCreateRequest(urlParameter string) (*azcore.Request, error) {
+func (client *serviceOperations) getAccountInfoCreateRequest() (*azcore.Request, error) {
 	urlPath := "/"
```
jhendrixMSFT: Would be good to remove this, avoiding the call to `client.u.Parse()` thus removing creating a new `url.URL` (I'm fine with deferring it to a separate PR if that works better).

src/generator/helpers.ts
```diff
@@ -164,13 +167,17 @@ export function generateParameterInfo(op: Operation): ParamInfo[] {
     }
     // include client and method params
     const global = param.implementation === ImplementationLocation.Client;
-    params.push(new paramInfo(param.language.go!.name, param.schema.language.go!.name, global, param.required === true));
+    let isHost = false;
+    if (global) {
+      isHost = (param.extensions!['x-ms-priority'] === 0 && param.extensions!['x-in'] === 'path');
```
jhendrixMSFT: Is the `!` going to be problematic?  `extensions` can legitimately be undefined.

catalinaperalta: I can add an if to check if it's undefined before

catalinaperalta: if (global && param.extensions !== undefined)

jhendrixMSFT: If you use `?` instead it gives you the same effect.

catalinaperalta: ok 👍

PR #323, opened 2020-03-25
Fix publishing of autorest test server coverage

jhendrixMSFT:
Specify the generator version when publishing results as the auto
detection logic for package.json doesn't work when using rush for
package management.  This required updating to the latest autorest
test server which introduced some swagger changes.

### DIFF
--- a/test/autorest/stringgroup/string_test.go
+++ b/test/autorest/stringgroup/string_test.go
@@ -120,7 +120,7 @@ func TestStringGetWhitespace(t *testing.T) {
 
 func TestStringPutEmpty(t *testing.T) {
 	client := getStringClient(t)
-	result, err := client.PutEmpty(context.Background(), "")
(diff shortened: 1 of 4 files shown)


test/autorest/stringgroup/string_test.go
```diff
@@ -120,7 +120,7 @@ func TestStringGetWhitespace(t *testing.T) {
 
 func TestStringPutEmpty(t *testing.T) {
 	client := getStringClient(t)
-	result, err := client.PutEmpty(context.Background(), "")
+	result, err := client.PutEmpty(context.Background())
```
catalinaperalta: These changes are due to a change in the swagger not a change in our codegen, right?

jhendrixMSFT: Correct, it's the change that was discussed in Tuesday's scrum.

PR #336, opened 2020-04-02
Add tests for arraygroup

jhendrixMSFT:
Fixed marshalling and unmarshalling of arrays of time types.
Improved generation of array and dictionary type names.


test/storage/2019-07-07/azblob/models.go
```diff
@@ -2353,7 +2353,7 @@ type ContainerGetPropertiesResponse struct {
 // An Azure Storage container
 type ContainerItem struct {
 	// Dictionary of <string>
-	Metadata *map[string]*string `xml:"String"`
+	Metadata *map[string]*string `xml:"string"`
```
catalinaperalta, line 2356: Just want to highlight that this changed. Is it still correct?

jhendrixMSFT: No, but it's ok as the codegen for metadata is wrong.  See the example of what we did for track 1.8.  We'll probably want to do something similar for track 2.
https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/zz_generated_models.go#L33

Issue #337, opened 2020-04-03
Check padding when encoding/decoding to base64


catalinaperalta:

jhendrixMSFT:
This is a dupe of https://github.com/Azure/autorest.go/issues/250

PR #338, opened 2020-04-03
Adding urlgroup paths tests

catalinaperalta:
Three tests are not working currently: 
- base64URL: #337
- StringUnicode
- UnixTimeURL

### DIFF
--- a/test/autorest/urlgroup/paths_test.go
+++ b/test/autorest/urlgroup/paths_test.go
@@ -1,0 +1,275 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package urlgrouptest
+
+import (
+	"context"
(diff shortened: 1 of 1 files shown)


test/autorest/urlgroup/paths_test.go
```diff
@@ -0,0 +1,275 @@
...
+func TestPathsStringURLNonEncoded(t *testing.T) {
+	client := getPathsOperations(t)
+	result, err := client.StringURLNonEncoded(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	helpers.VerifyStatusCode(t, result, http.StatusOK)
+}
+
+// TODO verify how to work with this type of string
```
jhendrixMSFT, line 253: I would uncomment these out and use `t.Skip` so that we can grep for all disabled tests.

catalinaperalta: Done

Issue #339, opened 2020-04-03
Integergroup: Error unmarshalling unix time to time.Time


catalinaperalta:

catalinaperalta:
This also includes marshalling time.Time to unix times or would include switching unix time params to int64

PR #341, opened 2020-04-03
Adding integergroup and integer tests

catalinaperalta:
Tests that are being skipped:
- Unix time related tests: #339 
- Max64 and Min64 tests the values that the testserver expects exceed the limits of Go's int64: #342


test/autorest/generated/integergroup/models.go
```diff
@@ -0,0 +1,63 @@
...
+		msg = "missing error info"
+	}
+	return msg
+}
+
+// Int32Response is the response envelope for operations that return a int32 type.
+type Int32Response struct {
+	// RawResponse contains the underlying HTTP response.
+	RawResponse *http.Response
+	Value       *int32
```
jhendrixMSFT, line 46: Note that this codegen is incorrect.  Since the type isn't annotated with `x-nullable` this shouldn't be pointer-to-type and thus fail when receiving a `null` JSON payload.

jhendrixMSFT: It will require an update to the swagger and test server.

test/autorest/integergroup/integergroup_test.go
```diff
@@ -0,0 +1,174 @@
...
+	client := getIntegerOperations(t)
+	result, err := client.GetNullUnixTime(context.Background())
+	if err != nil {
+		t.Fatalf("GetNullUnixTime: %v", err)
+	}
+	helpers.VerifyStatusCode(t, result.RawResponse, http.StatusOK)
+	helpers.DeepEqualOrFatal(t, result.Value, (*time.Time)(nil))
+}
+
+// TODO check if this is an appropriate check
```
jhendrixMSFT, line 65: Your overflow/underflow checks look good (unmarshalling should fail) so I think you can remove these comments (let me know if I misunderstood).

test/autorest/integergroup/integergroup_test.go
```diff
@@ -0,0 +1,174 @@
...
+		t.Fatalf("GetUnixTime: %v", err)
+	}
+	t1 := time.Unix(1460505600, 0)
+	helpers.VerifyStatusCode(t, result.RawResponse, http.StatusOK)
+	helpers.DeepEqualOrFatal(t, result.Value, t1)
+}
+
+func TestIntPutMax32(t *testing.T) {
+	client := getIntegerOperations(t)
+	result, err := client.PutMax32(context.Background(), int32(2147483647))
```
jhendrixMSFT, line 127: Use the min/max constants defined in the `math` package.

test/autorest/integergroup/integergroup_test.go
catalinaperalta: Yes I was just wondering if it was right if unmarshalling should fail. Will remove comments

test/autorest/generated/integergroup/models.go
catalinaperalta: Ok will create a PR with those changes in the testserver

PR #343, opened 2020-04-06
Enable complexgroup in its current state

jhendrixMSFT:
Updated complexgroup with latest codegen and fixed up tests.
Some things in this swagger are not fully supported (e.g. discriminated
types) so coverage for NYI items have not been added.


jhendrixMSFT:
Also updated text output for `covreport` to distinguish between general and Azure tests.

Issue #340, opened 2020-04-03
Add support for x-nullable


jhendrixMSFT:
Swagger schema specifies that nulls in payloads are forbidden unless the type is marked as `x-nullable`.  So there are two things that need to happen.

- For dictionary types, don't default to emitting the value as pointer-to-type
- For types annotated with `x-nullable`, emit as pointer-to-type (arrays, maps, etc)

jhendrixMSFT:
Given the problems with misbehaving services we won't be implementing full support for `x-nullable`.

PR #344, opened 2020-04-06
Added dictionarygroup codegen and tests

jhendrixMSFT:
By default, map values are no longer pointer-to-type as null is not
acceptable unless x-nullable: true has been set.
Some tests are commented out/skipped due to x-nullable or needed codegen
fixed (time and map of map).


src/transform/transform.ts
```diff
@@ -94,7 +94,7 @@ function schemaTypeToGoType(codeModel: CodeModel, schema: Schema, inBody: boolea
       const dictSchema = <DictionarySchema>schema;
       const dictElem = <Schema>dictSchema.elementType;
       dictElem.language.go!.name = schemaTypeToGoType(codeModel, dictElem, inBody);
-      return `map[string]*${dictElem.language.go!.name}`;
+      return `map[string]${dictElem.language.go!.name}`;
```
catalinaperalta, line 97: Doesn't this mean that we should check for the x-ms-nullable field here and add the * depending on its existence?

jhendrixMSFT: Yes, unfortunately it needs some work in autorest.

jhendrixMSFT: Here's the issue tracking support for this on our side.  https://github.com/Azure/autorest.go/issues/340

PR #346, opened 2020-04-07
Adding httpinfrastructure code and tests

catalinaperalta:


test/autorest/generated/httpinfrastructuregroup/httpredirects.go
```diff
@@ -0,0 +1,693 @@
...
+	if err != nil {
+		return nil, err
+	}
+	req := azcore.NewRequest(http.MethodGet, *u)
+	return req, nil
+}
+
+// get300HandleResponse handles the Get300 response.
+func (client *httpRedirectsOperations) get300HandleResponse(resp *azcore.Response) (*HTTPRedirectsGet300Response, error) {
+	if !resp.HasStatusCode(http.StatusOK) {
```
jhendrixMSFT, line 125: This should be `if !resp.HasStatusCode(http.StatusOK, http.StatusMultipleChoices)`

catalinaperalta: I added the status code to the list of status codes we can add in this function call, but the modeler is only returning this as a single status code 200 OK for this response. Will confirm if this is expected behavior with the autorest crew.

jhendrixMSFT: My earlier comment was from the track 1 codegen.  It could be the case that this has changed for track 2.

src/generator/operations.ts
```diff
@@ -412,9 +412,13 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
     let setOptionsPrefix = !bodyParam!.required;
     // default to the body param name
     let body = bodyParam!.language.go!.name;
+    let skipBody = false;
     if (bodyParam!.schema.type === SchemaType.Constant) {
       // if the value is constant, embed it directly
       body = formatConstantValue(<ConstantSchema>bodyParam!.schema);
+      if ((<ConstantSchema>bodyParam!.schema).valueType.type === SchemaType.Boolean) {
```
jhendrixMSFT, line 419: Why do we skip this only for bools?  Probably good to add a comment here.

src/generator/operations.ts
```diff
@@ -505,6 +509,13 @@ function isArrayOfRFC1123(schema: Schema): boolean {
 function createProtocolResponse(client: string, op: Operation, imports: ImportManager): string {
   const info = <OperationNaming>op.language.go!;
   const name = info.protocolNaming.responseMethod;
+  if (!op.responses) {
+    let text = `${comment(name, '// ')} handles the ${info.name} response.\n`;
```
jhendrixMSFT, line 513: I think this block could be moved down a bit so the first two lines aren't duplicated.

src/generator/operations.ts
catalinaperalta: Added the comment, also added an issue to improve this check: #348

PR #354, opened 2020-04-16
Change the version.go to use the `Version()` function to unify behaviour with onever and module version

ArcturusZhang:
This PR is trying to change the `version.go` in generated code from
```go
// UserAgent returns the UserAgent string to use when sending http.Requests.
func UserAgent() string {
	return "Azure-SDK-For-Go/" + version.Number + " windowsesu/2019-09-16-preview"
}

// Version returns the semantic version (see http://semver.org) of the client.
func Version() string {
	return version.Number
}
```
to
```go
// UserAgent returns the UserAgent string to use when sending http.Requests.
func UserAgent() string {
	return "Azure-SDK-For-Go/" + Version() + " windowsesu/2019-09-16-preview"
}

// Version returns the semantic version (see http://semver.org) of the client.
func Version() string {
	return version.Number
}
```
This pattern will also apply to the version file with modules:
```go
// UserAgent returns the UserAgent string to use when sending http.Requests.
func UserAgent() string {
	return "Azure-SDK-For-Go/" + Version() + " windowsesu/2019-09-16-preview"
}

// Version returns the semantic version (see http://semver.org) of the client.
func Version() string {
	return "v0.0.0"
}
```

### DIFF
--- a/src/Model/CodeModelGo.cs
+++ b/src/Model/CodeModelGo.cs
@@ -80,7 +80,7 @@ namespace AutoRest.Go.Model
         {
             get
             {
-                var verStr = $"\" + Version() + \"";
+                var verStr = "\" + Version() + \"";
                 var suffix = "";
 
                 // the API version will not be populated for composite packages that span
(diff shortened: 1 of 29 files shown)


fearthecowboy:
## success (version: 2.1.144)

PR #355, opened 2020-04-20
Add support for discriminated types

jhendrixMSFT:
Create interface types for discriminator base types with internal marker
method to be implemented by concrete types.
Generate custom marshaller that sets the discriminator automatically.
Generate custom unmarshaller for response envelopes and types that
contain discriminated types.
Generate internal polymorphic helpers for unmarshalling scalars and
arrays of polymorphic types.
Fix up struct fields and parameters to use applicable interface types.


src/transform/transform.ts
```diff
@@ -481,3 +495,66 @@ function generateResponseTypeName(schema: Schema): Language {
...
+    for (const parent of values(root.parents?.immediate)) {
+      // there can be parents that aren't part of the hierarchy.
+      // e.g. if type Foo is in a DictionaryOfFoo, then one of
+      // Foo's parents will be DictionaryOfFoo which we ignore.
+      if (isObjectSchema(parent)) {
+        root = parent;
+      }
+    }
+    // fail-safe to prevent infinite loop due to a bug somewhere...
+    if (root === obj) {
```
catalinaperalta, line 529: Is this bug pointing to a swagger that expresses the parent and the child as the same? OR what happened here? Just to know.

test/autorest/complexgroup/inheritance_test.go
```diff
@@ -31,12 +31,12 @@ func TestInheritanceGetValid(t *testing.T) {
 		Breed: to.StringPtr("persian"),
 		Color: to.StringPtr("green"),
 		Hates: &[]complexgroup.Dog{
-			complexgroup.Dog{
```
catalinaperalta, line 34: Why did this change? Again, just to know. Can we not declare the type here now?

src/transform/transform.ts
jhendrixMSFT: Nothing actually happened here, it's a preventative measure.  E.g. perhaps my loop isn't correct in some obscure case, or something changes in M4 etc, it will prevent the generator from just hanging.

test/autorest/complexgroup/inheritance_test.go
jhendrixMSFT: We can but it's redundant.

jhendrixMSFT: To clarify, array of `type Dog struct`, each element can only be a `Dog` so no need to explicitly declare each element's type.

src/transform/transform.ts
catalinaperalta: Ah I see.

test/autorest/complexgroup/inheritance_test.go
catalinaperalta: Ok good to know, will keep in mind

PR #356, opened 2020-04-21
Initial LRO implementation work

catalinaperalta:
This PR includes:
- naming polling methods according to guidelines
- implementing polling interfaces and internal structs
- implementing bare methods for internal structs that implement the polling interface
- outputting polling related code to pollers.go
- adding generated LRO testserver code
- corrected generateReturnsInfo name


test/autorest/generated/paginggroup/paging.go
```diff
@@ -272,26 +272,8 @@ func (client *pagingOperations) getMultiplePagesFragmentWithGroupingNextLinkHand
...
-			if err != nil {
-				return nil, fmt.Errorf("invalid NextLink: %w", err)
-			}
-			if u.Scheme == "" {
-				return nil, fmt.Errorf("no scheme detected in NextLink %s", *resp.ProductResult.NextLink)
-			}
-			return azcore.NewRequest(http.MethodGet, *u), nil
-		},
-	}, nil
+func (client *pagingOperations) BeginGetMultiplePagesLro(options *PagingGetMultiplePagesLroOptions) (ProductResultPager, error) {
```
jhendrixMSFT, line 275: This should return a `ProductResultPoller` (something like that).  When pagers and pollers intersect, the return type is a poller with a pager response type.

jhendrixMSFT: Just saw the TODO about this.

src/generator/operations.ts
```diff
@@ -627,6 +662,9 @@ function createInterfaceDefinition(group: OperationGroup, imports: ImportManager
       // don't generate a public API for the methods used to advance pages
       continue;
     }
+    if (isLROOperation(op)) {
+      op.language.go!.name = `Begin${op.language.go!.name}`;
```
jhendrixMSFT, line 666: This should probably go in the namer.

src/generator/operations.ts
```diff
@@ -253,18 +253,53 @@ function generateOperation(clientName: string, op: Operation, imports: ImportMan
...
-  text += `func (client *${clientName}) ${op.language.go!.name}(${params}) (${returns.join(', ')}) {\n`;
   // split param list into individual params
   const reqParams = getCreateRequestParametersSig(op).split(',');
   // slice off the parameter names from the type/type tuples
   for (let i = 0; i < reqParams.length; ++i) {
     reqParams[i] = reqParams[i].trim().split(' ')[0];
   }
+  // TODO Exception for Pageable LRO operations NYI
+  if (isLROOperation(op)) {
+    text += `func (client *${clientName}) Begin${op.language.go!.name}(${params}) (${returns.join(', ')}) {\n`;
```
jhendrixMSFT, line 269: Move adding Begin prefix to namer.

src/generator/pollers.ts
```diff
@@ -0,0 +1,72 @@
...
+
+import { Session } from '@azure-tools/autorest-extension-base';
+import { pascalCase } from '@azure-tools/codegen';
+import { CodeModel } from '@azure-tools/codemodel';
+import { values } from '@azure-tools/linq';
+import { PollerInfo } from '../common/helpers';
+import { contentPreamble, sortAscending } from './helpers';
+import { ImportManager } from './imports';
+
+// Creates the content in pagers.go
```
jhendrixMSFT, line 14: pollers.go

src/generator/operations.ts
catalinaperalta: Well....I thought about this when I was implementing it. I can do it, but I thought it would be better NOT to rename the operation to use begin. Since the original operation name is used/referred to in more places than the name with the Begin prefix which is only used in the interface and method declarations. What do you think?

jhendrixMSFT: Hmm this is a good point, it would spill into response envelope type names which is probably not what we want.  I wonder though if we should do this in the namer via something like `op.language.go!.methodPrefix` which we'd only set for LROs (although conceivably it could be used in the future for other things).  Then during interface/method generation if `methodPrefix` exists we add it.

test/autorest/generated/lrogroup/models.go
```diff
@@ -0,0 +1,851 @@
...
+// LrOSCustomHeaderPost202Retry200Options contains the optional parameters for the LrOSCustomHeader.Post202Retry200 method.
+type LrOSCustomHeaderPost202Retry200Options struct {
+	// Product to put
+	Product *Product
+}
+
+// LrOSCustomHeaderPost202Retry200Response contains the response from method LrOSCustomHeader.Post202Retry200.
+type LrOSCustomHeaderPost202Retry200Response struct {
+	// Location contains the information returned from the Location header response.
+	Location *string
```
jhendrixMSFT, line 50: Our header codegen is working great, but unfortunately this is introducing a bunch of response envelopes and thus pollers which seems unfortunate.

The following headers are implementation details for LROs.

- Azure-AsyncOperation
- Location
- Retry-After

Some swaggers include them however it's not necessary as the ARM LRO engine knows about them.  I haven't thought too deeply about this but I'm thinking that we should omit these headers from codegen which should remove all of these response envelopes and pollers.  Thoughts?

test/autorest/generated/paginggroup/paging.go
catalinaperalta: Sounds good, I'll implement it in my next PR

src/generator/operations.ts
catalinaperalta: Yes, I'll do that. What other future uses do you think it will be useful for? So that I keep that in mind when implementing this.

test/autorest/generated/lrogroup/models.go
catalinaperalta: Well, as you said, I think there could be other services that will use some/all of them. I am worried about removing them in case some service uses them and there is value in exposing them instead of hiding them away from users for cases that are not LRO.

src/generator/operations.ts
jhendrixMSFT: It could be useful if we want to add a `List` prefix to pageable methods.

test/autorest/generated/lrogroup/models.go
jhendrixMSFT: They'd use them as an implementation detail to polling, I don't think exposing them like this has any value but I could be wrong (it could be the case we do this for ARM but not data-plane).  We should discuss this in scrum.

src/generator/operations.ts
catalinaperalta: Added the methodPrefix

src/generator/operations.ts
catalinaperalta: Ok sounds good, I added a methodPrefix

test/autorest/generated/lrogroup/models.go
catalinaperalta: Hmmm well if you think that this is the only case they would be used then we should remove them. I'll add an issue.

catalinaperalta: Here's the issue #358

src/generator/operations.ts
```diff
@@ -266,7 +266,7 @@ function generateOperation(clientName: string, op: Operation, imports: ImportMan
   }
   // TODO Exception for Pageable LRO operations NYI
   if (isLROOperation(op)) {
-    text += `func (client *${clientName}) Begin${op.language.go!.name}(${params}) (${returns.join(', ')}) {\n`;
+    text += `func (client *${clientName}) ${op.language.go!.methodPrefix}${op.language.go!.name}(${params}) (${returns.join(', ')}) {\n`;
```
jhendrixMSFT, line 269: I think you can move this line back to its original place and do a check to see if `methodPrefix` has been defined then add it; this way it will work for methods other than LROs (see also change on line 666).

catalinaperalta: I'm thinking of not adding an if. If methodPrefix is nil then nothing will be appended. Thoughts?

catalinaperalta: Done. I left it with the if for now, it provides more control.

jhendrixMSFT: Even better.

src/generator/operations.ts
```diff
@@ -627,6 +665,9 @@ function createInterfaceDefinition(group: OperationGroup, imports: ImportManager
       // don't generate a public API for the methods used to advance pages
       continue;
     }
+    if (op.language.go!.methodPrefix) {
+      op.language.go!.name = `${op.language.go!.methodPrefix}${op.language.go!.name}`;
```
jhendrixMSFT, line 669: This will change the underlying name which is different than what we did elsewhere, is this safe?

catalinaperalta: I agree. I was thinking about this but didn't see it change anything else. Not safe or guaranteed however. I'll change it.

catalinaperalta: Done.

Issue #348, opened 2020-04-09
Review boolean check for skipOptions param in operations.ts


catalinaperalta:
Review whether this should also check for string types. Or see how to add a check for constants that are not enumerated types?

jhendrixMSFT:
Fixed in https://github.com/Azure/autorest.go/pull/357

PR #357, opened 2020-04-21
Add support for x-ms-parameter-grouping

jhendrixMSFT:
Create parameter group structs as specified by the extension.
The grouping of optional args into an 'options' struct was reworked to
build on this implementation.
Added paramgroupinggroup tests and fixed a few tests that were impacted.


src/generator/models.ts
```diff
@@ -432,10 +427,8 @@ function generateStruct(lang: Language, props?: Property[]): StructDef {
   return st;
 }
 
-function generateOptionalParamsStruct(optionalParam: Parameter): StructDef {
-  // for the optional params struct, the params are smuggled via GroupProperty
-  const params = (<GroupProperty>(<ObjectSchema>optionalParam.schema).properties![0]).originalParameter;
-  const st = new StructDef(optionalParam.schema.language.go!, undefined, params);
```
catalinaperalta, line 438: Are parameter groups always optional? Since I'm seeing that it's replacing this.

jhendrixMSFT: Parameter groups are specified with [x-ms-parameter-grouping](https://github.com/Azure/autorest/tree/master/docs/extensions#x-ms-parameter-grouping) which is optional.

The creation of the optional params struct was reworked to use this same mechanism, see link 258 in transform.ts.

jhendrixMSFT:
At first I was a bit annoyed about the changing of the optional params arg name, however after thinking about it a bit, it's consistent with explicitly defined param groups so I'm ok with it now (let me know if you disagree).

catalinaperalta:
I'm still thinking about it. Not sure I'm where you are about it yet. But I'm also not strongly opposed at the moment. You're saying that you don't mind it because the name corresponds to the param group name?

jhendrixMSFT:
Yes and is consistent with explicitly defined parameter groups.

jhendrixMSFT:
Fixed https://github.com/Azure/autorest.go/issues/348

Issue #295, opened 2020-03-10
Return interface{} from operations that return multiple models


catalinaperalta:
If not the same then implement distinct responses.

jhendrixMSFT:
Example `multipleResponses_get200Model201ModelDefaultError201Valid` in httpinfrastructuregroup.json.

```go
func (client *multipleResponsesOperations) get200Model201ModelDefaultError201ValidHandleResponse(resp *azcore.Response) (interface{}, error) {
	switch resp.StatusCode {
	case http.StatusOK:
		result := MyExceptionResponse{RawResponse: resp.Response}
		return &result, resp.UnmarshalAsJSON(&result.MyException)
	case http.StatusCreated:
		result := BResponse{RawResponse: resp.Response}
		return &result, resp.UnmarshalAsJSON(&result.B)
	default:
		return nil, client.get200Model201ModelDefaultError201ValidHandleError(resp)
	}
}
```

jhendrixMSFT:
Fixed

PR #359, opened 2020-04-22
Initial support for multiple responses

jhendrixMSFT:
Add support for operations that specify multiple response codes that
have the same schema (or no schema).


jhendrixMSFT:
Part of the fix for https://github.com/Azure/autorest.go/issues/295

PR #361, opened 2020-04-24
Adding polling functionality for Poll() and Done() methods

catalinaperalta:
This PR adds:
- internal polling helpers in pollers_helpers.go
- polling initialization in Begin polling methods
- functionality for Poll() and Done() methods
- updates handling of poller responses
- adding a couple of initial polling tests


src/generator/pollers.ts
```diff
@@ -49,25 +49,711 @@ type ${poller.name} struct {
...
+	imports.add('github.com/Azure/azure-sdk-for-go/sdk/azcore');
+	imports.add('io/ioutil');
+	imports.add('net/http');
+	imports.add('net/url');
+	imports.add('strings');
+	imports.add('fmt');
+	imports.add('errors');
+	text += imports.text();
+	let errorCtor = pollers[0].pollingError.language.go!.constructorName;
+	// TODO separate this into manageable chunks of text by section of functionality
```
catalinaperalta, line 146: This will be included in a future PR

test/autorest/lrogroup/lrogroup_test.go
```diff
@@ -0,0 +1,80 @@
...
+	for {
+		w, err := poller.Poll(context.Background())
+		if err != nil {
+			t.Fatal(err)
+		}
+		if poller.Done() {
+			helpers.VerifyStatusCode(t, w.RawResponse, 204)
+			break
+		}
+		time.Sleep(1 * time.Second)
```
jhendrixMSFT, line 57: Might want to cut this down to something like 200ms (making it up) so that all the tests run fast.

test/autorest/generated/lrogroup/lroretrys.go
```diff
@@ -184,7 +244,25 @@ func (client *lroRetrysOperations) post202Retry200HandleResponse(resp *azcore.Re
...
+	req, err := client.postAsyncRelativeRetrySucceededCreateRequest(lroRetrysPostAsyncRelativeRetrySucceededOptions)
+	if err != nil {
+		return nil, err
+	}
+	// send the first request to initialize the poller
+	resp, err := client.p.Do(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	pt := pollingTrackerPost{
```
jhendrixMSFT, line 256: I would move this creation logic into an internal method in pollers_helpers.go that creates the tracker and calls `initializeState()` (you can get the request verb from the resp object).

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -0,0 +1,628 @@
...
+		b, err := ioutil.ReadAll(pt.resp.Body)
+		if err != nil {
+			return errors.New("failed to read response body")
+		}
+		// observed in 204 responses over HTTP/2.0; the content length is -1 but body is empty
+		if len(b) == 0 {
+			return nil
+		}
+		// put the body back so it's available to other callers
+		pt.resp.Body = ioutil.NopCloser(bytes.NewReader(b))
```
jhendrixMSFT, line 176: I believe this can be removed as we should already be doing this in the body download policy (but please verify).

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+	// if the operation has failed update the error state
+	if pt.hasFailed() {
+		pt.updateErrorFromResponse()
+	}
+	return nil
+}
+
+func (pt pollingTrackerBase) pollingError() error {
+	if pt.Err == nil {
```
jhendrixMSFT, line 238: I think this if block is redundant?

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -0,0 +1,628 @@
...
+}
+
+// verify that the URL is valid and absolute
+func isValidURL(s string) bool {
+	u, err := url.Parse(s)
+	return err == nil && u.IsAbs()
+}
+
+// PollingMethodType defines a type used for enumerating polling mechanisms.
+type PollingMethodType string
```
jhendrixMSFT, line 614: Unexport this type and its constants?

test/autorest/generated/lrogroup/lroretrys.go
jhendrixMSFT: Looks like this method already exists createPollingTracker, just need to invoke it?

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -23,25 +25,71 @@ type lrOSCustomHeaderPost202Retry200Poller struct {
...
+// ID ...
 func (p *lrOSCustomHeaderPost202Retry200Poller) ID() string {
-	return "NYI"
+	return ""
 }
 
 func (p *lrOSCustomHeaderPost202Retry200Poller) Poll(ctx context.Context) (*LrOSCustomHeaderPost202Retry200Response, error) {
-	return nil, nil
+	done, err := p.doneWithContext(ctx)
+	if done {
```
jhendrixMSFT, line 44: Remember your happy line.
```go
if done, err := p.doneWithContext(ctx); !done {
    return nil, err
}
// remaining code...
```

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -23,25 +25,71 @@ type lrOSCustomHeaderPost202Retry200Poller struct {
...
+func (p *lrOSCustomHeaderPost202Retry200Poller) response() *azcore.Response {
+	if p.pt == nil {
+		return nil
+	}
+	return p.pt.latestResponse()
+}
+
+// doneWithContext queries the service to see if the operation has completed.
+func (p *lrOSCustomHeaderPost202Retry200Poller) doneWithContext(ctx context.Context) (done bool, err error) {
+	if p.pt == nil {
```
jhendrixMSFT, line 69: I think this if block can go away now as with the new design we can't create a poller with an uninitialized tracker right?

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -23,25 +25,71 @@ type lrOSCustomHeaderPost202Retry200Poller struct {
...
+// Response returns the last HTTP response.
+func (p *lrOSCustomHeaderPost202Retry200Poller) response() *azcore.Response {
+	if p.pt == nil {
+		return nil
+	}
+	return p.pt.latestResponse()
+}
+
+// doneWithContext queries the service to see if the operation has completed.
+func (p *lrOSCustomHeaderPost202Retry200Poller) doneWithContext(ctx context.Context) (done bool, err error) {
```
jhendrixMSFT, line 68: Rename these methods to `done(ctx context.Context)` (the `WithContext` was due to pre-Go 1.7 implementation when `context.Context` didn't yet exist).

src/generator/pollers.ts
jhendrixMSFT: Do you mean splitting this up into multiple helper files?

test/autorest/lrogroup/lrogroup_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package lrogrouptest
```
jhendrixMSFT, line 4: Can more tests be added or is there not enough working yet?

jhendrixMSFT: Hmm maybe add them once `Wait()` is implemented as that calls the other methods.

src/generator/pollers.ts
catalinaperalta: No no, I simply want to split up this huge chunk of text into smaller named vars. Like constDeclarations, pollerBaseDefinition, pollertrackerdeleteMethods, things like that, that could make it more manageable to make changes to this code. What do you think? That's what I came up with to make this easier to modify in the future.

test/autorest/lrogroup/lrogroup_test.go
catalinaperalta: Done. I had left it as 1 Second since, at least for these two tests, we dont loop twice either way and I think most tests in the testserver will behave like that. But you're right it's good to have a shorter wait time between test requests.

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: So I made this change, but it involved declaring the pollingTracker interface instead of the concrete poller type (pollingTrackerDelete for instance) in each of the internal poller structs that we are creating for each operation. I think it's fine to use the pollingTracker interface which is why I made the change. If you want to revert or discuss let me know.

test/autorest/generated/lrogroup/pollers_helper.go
catalinaperalta: Yes we do handle this in the download policy: https://github.com/Azure/azure-sdk-for-go/blob/7da5bb2d5ffd1a370c899d15fac9b59f4332c5bd/sdk/azcore/policy_body_download.go#L26-L32

I'll remove this line

test/autorest/generated/lrogroup/pollers_helper.go
catalinaperalta: Done

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: So I've got a problem here and I have a note on my local version about this. What should be done when done() returns an error? I want to return immediately. Which is what I am going to push. 
So it would end up something like this: 

```
if done, err := p.doneWithContext(ctx); !done || err != nil {
    return nil, err
}
// success case here without errors
```
What do you think? I believe this does vary from what the guidelines are saying

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Yes, since we initialize in the Begin method we can trust that when Poll is called a poller has been initialized. I'll remove it.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Done

test/autorest/lrogroup/lrogroup_test.go
catalinaperalta: Yes I was waiting to add all tests until we finalize most (if not all) aspects of the LROs.

test/autorest/generated/lrogroup/pollers_helper.go
catalinaperalta: Removed. Done.

catalinaperalta: Removed. Done.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: I'll merge this for now. But let me know what you think and I'll update my next PR accordingly.

jhendrixMSFT: The condition works however won't `done()` always return `false` in the event that its returned error is not nil?

src/generator/pollers.ts
jhendrixMSFT: I see what you mean now.  What would be great is if we had a way to have templated Go code (i.e. actual `.go` source files) that we could consume as part of code generation.  Kind of like what we had in track 1 with the razor files, however we only use them for things like this where the majority of the code is constant (was thinking about this when I did pagers).

test/autorest/generated/lrogroup/lroretrys.go
jhendrixMSFT: Yep I think that's great.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Probably, but not necessarily. This is the code we have in done()

```
func (p *lrOSCustomHeaderPost202Retry200Poller) done(ctx context.Context) (done bool, err error) {
	if p.pt.hasTerminated() {
		return true, p.pt.pollingError()
	}
	if err := p.pt.pollForStatus(ctx, p.client.p); err != nil {
		return false, err
	}
	if err := p.pt.checkForErrors(); err != nil {
		return p.pt.hasTerminated(), err
	}
	if err := p.pt.updatePollingState(p.pt.provisioningStateApplicable()); err != nil {
		return false, err
	}
	if err := p.pt.initPollingMethod(); err != nil {
		return false, err
	}
	if err := p.pt.updatePollingMethod(); err != nil {
		return false, err
	}
	return p.pt.hasTerminated(), p.pt.pollingError()
}
```

catalinaperalta: Because if polling fails with an error from the service, terminated will be true but there will also be an error.

PR #365, opened 2020-04-28
Adding wait method to LRO pollers

catalinaperalta:
Adding the Wait method on LRO pollers, add Wait to LRO tests and remove extra nil check in response() method.

### DIFF
--- a/src/generator/pollers.ts
+++ b/src/generator/pollers.ts
@@ -24,7 +24,6 @@ export async function generatePollers(session: Session<CodeModel>): Promise<stri
   imports.add('github.com/Azure/azure-sdk-for-go/sdk/azcore');
   imports.add('net/http');
   imports.add('time');
-  imports.add('strconv');
   text += imports.text();
 
   const pollers = <Array<PollerInfo>>session.model.language.go!.pollerTypes;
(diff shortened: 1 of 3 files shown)


src/generator/pollers.ts
```diff
@@ -24,17 +24,20 @@ export async function generatePollers(session: Session<CodeModel>): Promise<stri
...
   imports.add('time');
+  imports.add('strconv');
   text += imports.text();
 
   const pollers = <Array<PollerInfo>>session.model.language.go!.pollerTypes;
   pollers.sort((a: PollerInfo, b: PollerInfo) => { return sortAscending(a.name, b.name) });
   for (const poller of values(pollers)) {
     const pollerInterface = pascalCase(poller.name);
-    let responseType = '';
+	let responseType = '';
```
jhendrixMSFT, line 34: Nit, spacing (next line too).

src/generator/pollers.ts
```diff
@@ -74,15 +77,34 @@ func (p *${poller.name}) Poll(ctx context.Context) (*${responseType}, error) {
...
-	return nil, nil
+	for {
+		resp, err := p.Poll(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		if p.Done() {
+			return resp, err
+		}
+		if ra := resp${rawResponse}.Header.Get(azcore.HeaderRetryAfter); len(ra) > 0 {
```
jhendrixMSFT, line 92: You can use `resp.RetryAfter()` which has all this code in it already.

catalinaperalta: SO I had the same thought and tried to do that until I realized that the response that we receive here is the response envelope for the operation.

src/generator/pollers.ts
catalinaperalta: Done

src/generator/pollers.ts
jhendrixMSFT: Per our conversation can use `p.response()`.

catalinaperalta: Done

PR #367, opened 2020-04-29
Infer package name from output directory

jhendrixMSFT:
Removes the need for --namespace.

### DIFF
--- a/src/generator/time.ts
+++ b/src/generator/time.ts
@@ -6,6 +6,7 @@
 import { Session } from '@azure-tools/autorest-extension-base';
 import { comment } from '@azure-tools/codegen'
 import { CodeModel } from '@azure-tools/codemodel';
+import { contentPreamble } from './helpers'
 
 // represents the generated content for an operation group
 export class Content {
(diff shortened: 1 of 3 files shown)


src/generator/time.ts
```diff
@@ -21,22 +22,19 @@ export class Content {
...
   }
   return content;
 }
 
-function generateRFC1123Helper(header: string, packageName: string): string {
-  return `${header}
-
-package ${packageName}
+function generateRFC1123Helper(preamble: string): string {
+  return `${preamble}
```
catalinaperalta, line 37: Semi colon

src/generator/time.ts
```diff
@@ -73,10 +71,8 @@ func (t *timeRFC1123) UnmarshalText(data []byte) error {
 `;
 }
 
-function generateRFC3339Helper(header: string, packageName: string): string {
-  return `${header}
-
-package ${packageName}
+function generateRFC3339Helper(preamble: string): string {
+  return `${preamble}
```
catalinaperalta, line 75: semicolon

src/generator/time.ts
jhendrixMSFT: It's at the very end of the text (same with other type).

catalinaperalta: Oops you're right I see now😅

Issue #297, opened 2020-03-10
Implement x-ms-parameterized-host functionality


catalinaperalta:
https://github.com/Azure/autorest/blob/master/docs/extensions/readme.md#x-ms-parameterized-host

We implicitly implemented some of this in order to support blobs.  Needs to be completed.

catalinaperalta:
Added in #368

PR #368, opened 2020-04-30
Adding host work for x-ms-parameterized-host

catalinaperalta:
This PR adds the missing part for parsing host parameters for a parameterized host scheme. 

WIP: reference parameters that are on the client where the pipeline is when creating the request or add them to a url in the client

### DIFF
--- a/src/generator/client.ts
+++ b/src/generator/client.ts
@@ -143,6 +143,15 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
         }
       }
     }
+    for (const param of values(paramHostInfo.clientParams)) {
+      let pointer = '';
+      if (param.clientDefaultValue) {
+        pointer = '*';
+      }
+      text += `\t${param.language.go!.name} ${param.schema.language.go!.name}\n`;
+      clientOnlyParamsFuncSig += `${param.language.go!.name} ${pointer}${param.schema.language.go!.name}, `;
+      passClientOnlyParams += `${param.language.go!.name}, `;
+    }
   } else {
     text += `\t${urlVar} *url.URL\n`;
   }
@@ -158,27 +167,15 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
   if (!session.model.security.authenticationRequired) {
     credParam = '';
   }
-  if (defaultsExist(clientOnlyParams)) {
+  if (defaultsExist(clientOnlyParams) && !paramHostInfo.addParamHost) {
     text += `// ${newDefaultClient} creates an instance of the ${client} type using the ${defaultEndpoint}.\n`;
     text += `func ${newDefaultClient}(${credParam}options *${clientOptions}) (*${client}, error) {\n`;
     let cred = 'cred, ';
     if (!session.model.security.authenticationRequired) {
       cred = '';
     }
-    if (paramHostInfo.addParamHost) {
-      text += `\treturn ${newClient}(`;
-      for (const param of values(clientOnlyParams)) {
-        if (param.clientDefaultValue !== undefined) {
-          text += `"${param.clientDefaultValue}", `;
-        } else {
-          text += '"", ';
-        }
-      }
-      text += `${cred}options)\n`;
-    } else {
-      // when a parameterized host is not set, the default endpoint is set for the url on the client
-      text += `\treturn ${newClient}(${defaultEndpoint}, ${cred}options)\n`;
-    }
+    // when a parameterized host is not set, the default endpoint is set for the url on the client
+    text += `\treturn ${newClient}(${defaultEndpoint}, ${cred}options)\n`;
     text += '}\n\n';
   }
 
@@ -248,9 +245,9 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
       if ((pp.implementation === ImplementationLocation.Client && group.language.go!.clientParams === undefined) || (pp.implementation === ImplementationLocation.Client && !(<Array<Parameter>>group.language.go!.clientParams).includes(pp))) {
         text += `\thostURL = strings.ReplaceAll(hostURL, "{${pp.language.go!.serializedName}}", client.Client.${pp.language.go!.serializedName})\n`;
       } else {
-        let paramValue = `url.PathEscape(${formatParamValue(pp, imports)})`;
+        let paramValue = `url.PathEscape(${formatParamValue(pp, imports, false)})`;
         if (skipURLEncoding(pp)) {
-          paramValue = formatParamValue(pp, imports);
+          paramValue = formatParamValue(pp, imports, false);
         } else {
           imports.add('net/url');
         }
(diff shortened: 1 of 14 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/297

src/generator/operations.ts
```diff
@@ -423,19 +423,50 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
...
+  const hostURLStr = <string>op.requests![0].protocol.http!.uri;
+  // for parameterized hosts that only in the format {x} will be ignored and the client url will be used
+  if (!((hostURLStr.match(/{/g) || []).length === 1 && (hostURLStr.match(/}/g) || []).length === 1 && hostURLStr.indexOf('{') === 0 && hostURLStr.indexOf('}') === hostURLStr.length - 1)) {
+    text += `\thost := "${hostURLStr}"\n`;
+    // replace url parameters
+    for (const pp of values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http!.in === 'uri'; })) {
+      imports.add('strings');
+      // host references may be found among other variables and therefore call to use the host 
+      // specified by the endpoint the client was created with
+      if (pp.language.go!.name == 'host') {
```
jhendrixMSFT, line 438: Can't this be removed?  I think it's dangerous to rely on a specific name (also should be `===`).

src/generator/operations.ts
```diff
@@ -423,19 +423,50 @@ function createProtocolRequest(client: string, op: Operation, imports: ImportMan
   const returns = ['*azcore.Request', 'error'];
   let text = `${comment(name, '// ')} creates the ${info.name} request.\n`;
   text += `func (client *${client}) ${name}(${getCreateRequestParametersSig(op)}) (${returns.join(', ')}) {\n`;
+  let includeParse = false;
+  let parseVar = '';
+  // hostURLStr will be used to check for the x-ms-parameterized-host functionality
+  const hostURLStr = <string>op.requests![0].protocol.http!.uri;
+  // for parameterized hosts that only in the format {x} will be ignored and the client url will be used
+  if (!((hostURLStr.match(/{/g) || []).length === 1 && (hostURLStr.match(/}/g) || []).length === 1 && hostURLStr.indexOf('{') === 0 && hostURLStr.indexOf('}') === hostURLStr.length - 1)) {
```
jhendrixMSFT, line 431: Can this be simplified to something like `if (!(hostURLStr[0] === '{' && hostURLStr[hostURLStr.length - 1] === '}'))` or are there other cases.

catalinaperalta: There are other cases like "{accountName}.{dnsSuffix}" Just making it up but it's in the test swaggers

src/generator/operations.ts
catalinaperalta: I'll double check the swaggers, to see if this case still exists.

src/generator/operations.ts
catalinaperalta: Example: https://github.com/Azure/autorest.testserver/blob/master/swagger/custom-baseUrl.json

src/generator/operations.ts
catalinaperalta: So there is a test in the testserver that relies on this. Actually it's the same as the one below: https://github.com/Azure/autorest.testserver/blob/master/swagger/custom-baseUrl.json

Without other identifying information, I'm not sure how to go about determining what params are host values. I agree it's not good to check based off of the name. :(

src/generator/operations.ts
jhendrixMSFT: I made a slight change to this condition and updated the comment, PTAL.

src/generator/operations.ts
jhendrixMSFT: Thinking about this, given this is parameterizing the host I think it's safe to say that the first path param in `hostURLStr` will always be the host and should be replaced with `client.u.Host` so we don't need to rely on the param name.  Any other params will use named replacement.  What do you think?

jhendrixMSFT:
Also, when `useSchemePrefix: true` we should omit the scheme check in `NewClientWithPipeline()` as it's not needed and gets in the way (see the [python](https://github.com/Azure/autorest.python/blob/master/test/vanilla/AcceptanceTests/test_custom_base_uri.py) version of this test to see what I'm talking about).

src/generator/operations.ts
catalinaperalta: So I looked at the change, if you think that it's correct to assume that whenever there is only 1 set of `{}` that means we should replace it with the host url, then it looks good to me. My only concern is if there was ever a case where is isn't the host. I did look through the REST API specs, so far it looks like this wont be a problem. Just want to know, how we know that this is correct?

catalinaperalta:
> Also, when `useSchemePrefix: true` we should omit the scheme check in `NewClientWithPipeline()` as it's not needed and gets in the way (see the [python](https://github.com/Azure/autorest.python/blob/master/test/vanilla/AcceptanceTests/test_custom_base_uri.py) version of this test to see what I'm talking about).

But we don't know that useSchemePrefix is true...this is simply returned to us through m4 without direct access to that info.

src/generator/operations.ts
jhendrixMSFT: The are two buckets of values that can be in `hostURLStr` (which comes from `op.requests![0].protocol.http!.uri`)

- `{whatever}`
- `{whatever}/some/path` (or variants of this with more `{stuff}`)

For the first case (which is what the condition checks for), I don't think it can be anything else but the host URL.

jhendrixMSFT: Made another change to the condition, PTAL.

test/autorest/generated/morecustombaseurigroup/client.go
```diff
@@ -69,22 +67,19 @@ func NewClient(endpoint string, options *ClientOptions) (*Client, error) {
...
-	return &Client{u: u, p: p}, nil
+func NewClientWithPipeline(p azcore.Pipeline) (*Client, error) {
+	return &Client{p: p}, nil
 }
 
 // PathsOperations returns the PathsOperations associated with this client.
-func (client *Client) PathsOperations(dnsSuffix string, subscriptionID string) PathsOperations {
-	return &pathsOperations{Client: client, dnsSuffix: dnsSuffix, subscriptionID: subscriptionID}
+// For dnsSuffix pass nil to use the default value of "host"
+func (client *Client) PathsOperations(dnsSuffix *string, subscriptionID string) PathsOperations {
```
jhendrixMSFT, line 80: Why is `dnsSuffix` on the operation group here?  It's a client param that's common to all operation groups.

src/generator/client.ts
```diff
@@ -96,31 +106,84 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
...
+    clientOnlyParamsFuncSig = '';
+    passClientOnlyParams = '';
+  }
   text += `type ${client} struct {\n`;
-  text += `\t${urlVar} *url.URL\n`;
+  if (paramHostInfo.addParamHost && !paramHostInfo.urlOnClient) {
+    if (clientOnlyParams.length > 0) {
+      for (const param of values(clientOnlyParams)) {
+        if (param.protocol.http!.in === 'uri') {
+          text += `\t${param.language.go!.name} ${schemaTypeToGoType(session.model, param.schema, false)}\n`;
```
jhendrixMSFT, line 142: Why didn't this param get converted during transformation?

test/autorest/generated/morecustombaseurigroup/client.go
catalinaperalta: You're right I was missing that check it's added now. Shared client params are now on the client in the case of parameterized host

src/generator/operations.ts
catalinaperalta: Agreed this was removed.

src/generator/client.ts
```diff
@@ -96,31 +106,66 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
...
+    clientOnlyParamsFuncSig = '';
+    passClientOnlyParams = '';
+  }
   text += `type ${client} struct {\n`;
-  text += `\t${urlVar} *url.URL\n`;
+  if (paramHostInfo.addParamHost && !paramHostInfo.urlOnClient) {
+    // add fields that are global client only params
+    for (const param of values(clientOnlyParams)) {
+      if (param.protocol.http!.in === 'uri') {
+        text += `\t${param.language.go!.name} ${param.schema.language.go!.name}\n`;
```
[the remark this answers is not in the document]
catalinaperalta: So the reason for having the `!` is because by this point in our generator we know that the params will have `language.go` on them, since we've already performed transforms and naming changes in this field for all params. I'm open to changing to check this in the future however if you recommend it. It's sort of implicitly known for us at this point.

test/autorest/generated/custombaseurlgroup/client.go
```diff
@@ -69,19 +68,18 @@ func NewClient(endpoint string, options *ClientOptions) (*Client, error) {
...
 // NewClientWithPipeline creates an instance of the Client type with the specified endpoint and pipeline.
-func NewClientWithPipeline(endpoint string, p azcore.Pipeline) (*Client, error) {
-	u, err := url.Parse(endpoint)
-	if err != nil {
-		return nil, err
+func NewClientWithPipeline(host *string, p azcore.Pipeline) (*Client, error) {
+	client := &Client{}
+	client.p = p
+	if host == nil {
+		*host = "host"
```
jhendrixMSFT, line 79: This is going to panic.

PR #369, opened 2020-04-30
Refactor long running operation's interface

catalinaperalta:
This PR updates the polling interface to include Response() and ResumeToken(). ResumeToken() will fail on a poller that's in a terminal state. Also changes Poll() to return a boolean. Updating tests to use the new model.

### DIFF
--- a/src/generator/pollers.ts
+++ b/src/generator/pollers.ts
@@ -65,7 +65,7 @@ func (p *${poller.name}) Poll(ctx context.Context) bool {
 	}
 }
 
-// Response ...
+// Response returns the final response from the polling operation
 func (p *${poller.name}) Response() (*${responseType}, error) {
 	resp := p.response()
 	if resp == nil {
@@ -78,7 +78,8 @@ func (p *${poller.name}) Response() (*${responseType}, error) {
 	return result, nil
 }
 
-// ResumeToken ...
+// ResumeToken generates the string token that can be used with the Resume${pascalCase(poller.name)} method
+// on the client to create a new poller from the data held in the current poller type
 func (p *${poller.name}) ResumeToken() (string, error) {
 	if p.pt.hasTerminated() {
 		return "", errors.New("cannot create a ResumeToken from a poller in a terminal state")
@@ -108,7 +109,7 @@ func (p *${poller.name}) Wait(ctx context.Context, pollingInterval time.Duration
 	return resp, nil
 }
 
-// Response returns the last HTTP response.
+// response returns the last HTTP response.
 func (p *${poller.name}) response() *azcore.Response {
 	return p.pt.latestResponse()
 }
--- a/src/generator/pollers.ts
+++ b/src/generator/pollers.ts
@@ -65,7 +65,7 @@ func (p *${poller.name}) Poll(ctx context.Context) bool {
 	}
 }
 
-// Response returns the final response from the polling operation
+// Response returns the latest response that is stored from the latest polling operation
 func (p *${poller.name}) Response() (*${responseType}, error) {
 	resp := p.response()
 	if resp == nil {
@@ -86,7 +86,7 @@ func (p *${poller.name}) ResumeToken() (string, error) {
 	}
 	js, err := json.Marshal(p.pt)
 	if err != nil {
-		return "", fmt.Errorf("json.Marshal: %s", err.Error())
+		return "", err
 	}
 	return string(js), nil
 }
--- a/src/generator/pollers.ts
+++ b/src/generator/pollers.ts
@@ -96,7 +96,7 @@ func (p *${poller.name}) ResumeToken() (string, error) {
 // is specified will be used to wait between polling requests. 
 func (p *${poller.name}) Wait(ctx context.Context, pollingInterval time.Duration) (*${responseType}, error) {
 	for p.Poll(context.Background()) {
-		if delay, found := p.response().RetryAfter(); found && delay > 0 {
+		if delay := p.response().RetryAfter(); delay > 0 {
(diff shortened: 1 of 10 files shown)


src/generator/pollers.ts
```diff
@@ -53,50 +56,60 @@ type ${poller.name} struct {
...
+	} else {
+		return !done
+	}
 }
 
-func (p *${poller.name}) Poll(ctx context.Context) (*${responseType}, error) {
-	if done, err := p.done(ctx); !done || err != nil {
-		return nil, err
-	}
+// Response returns the final response from the polling operation
```
jhendrixMSFT, line 68: We might want to change this to return the last response for consumers that need to write custom end-to-end polling (kubernetes); this would mean that we'd never return a nil response envelope as it would always have a raw response, not sure how I feel about that.  We can discuss later and update/change as required in a future PR.

src/generator/pollers.ts
```diff
@@ -53,50 +56,60 @@ type ${poller.name} struct {
...
 
+// ResumeToken generates the string token that can be used with the Resume${pascalCase(poller.name)} method
+// on the client to create a new poller from the data held in the current poller type
+func (p *${poller.name}) ResumeToken() (string, error) {
+	if p.pt.hasTerminated() {
+		return "", errors.New("cannot create a ResumeToken from a poller in a terminal state")
+	}
+	js, err := json.Marshal(p.pt)
+	if err != nil {
+		return "", fmt.Errorf("json.Marshal: %s", err.Error())
```
jhendrixMSFT, line 89: I would return just the `err` here.

src/generator/pollers.ts
```diff
@@ -53,50 +56,60 @@ type ${poller.name} struct {
...
-	for {
-		resp, err := p.Poll(context.Background())
-		if err != nil {
-			return nil, err
-		}
-		if p.Done() {
-			return resp, err
-		}
+	for p.Poll(context.Background()) {
 		if delay, found := p.response().RetryAfter(); found && delay > 0 {
```
jhendrixMSFT, line 99: When will this be updated to use `azcore` v0.6.0 that just returns `int`?

catalinaperalta: I'll update that in this PR, it slipped my mind.

src/generator/pollers.ts
catalinaperalta: No, I think it's fine to return just the err here. It was an automatic reaction to wrap it.

catalinaperalta: Also done.

src/generator/pollers.ts
catalinaperalta: Done

src/generator/pollers.ts
catalinaperalta: I think my comment was misleading, I've updated it now.

src/generator/pollers.ts
```diff
@@ -53,50 +56,60 @@ type ${poller.name} struct {
...
+	} else {
+		return !done
+	}
 }
 
-func (p *${poller.name}) Poll(ctx context.Context) (*${responseType}, error) {
-	if done, err := p.done(ctx); !done || err != nil {
-		return nil, err
-	}
+// Response returns the latest response that is stored from the latest polling operation
```
jhendrixMSFT, line 68: Does this actually work right now though?  If the poller is in a non-terminal state and I call `Response()` won't the response handler fail as there's no payload to unmarshal?  We don't have to fix this now but we do need a solution for it.

src/generator/pollers.ts
```diff
@@ -53,50 +56,60 @@ type ${poller.name} struct {
...
 			time.Sleep(delay)
 		} else {
 			time.Sleep(pollingInterval)
 		}
 	}
+	resp, err := p.Response()
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
```
jhendrixMSFT, line 109: Can this be condensed to `return p.Response()`?  I think `Response()` already returns `nil` for the response envelope if there's an error.

src/generator/pollers.ts
catalinaperalta: According to me it does work that way. Because I have not changed the underlying functionality of the poller, in that it saves the current response to the response field on it. But I'll double check this by stepping through one of the tests in the testserver. p.response() returns the resp field that's on the poller.

catalinaperalta: OH

catalinaperalta: Ignore my previous comment, I understand what you mean now. Yes we need to think of a better solution for handling the response.

jhendrixMSFT: One idea is to add a `LastResponse() *http.Response` method to the interface and rename `Response()` to `Result()`

src/generator/pollers.ts
catalinaperalta: Done

src/generator/pollers.ts
catalinaperalta: Yes that could work. Of course currently in the handleResponse method we always assign at a minimum the RawResponse to the RawResponse field in the envelope. Would that be enough? (I like your idea it's clean and clear, just wondering about what we already have)

jhendrixMSFT: We could, it just feels a little clunky to me.  Plus I think we'd have to change the codegen for the responder to properly handle this case.  The three of us should probably have a (hopefully quick) chat about this scenario.

catalinaperalta: Ok sounds good to me. I'll schedule a meeting for next week. For now I'll leave it as is.

Issue #362, opened 2020-04-24
LRO: adding LROs that return pagers


catalinaperalta:

catalinaperalta:
Added in #382

jhendrixMSFT:
Fixed in https://github.com/Azure/autorest.go/pull/427

PR #371, opened 2020-05-01
Adding resume widget poller method

catalinaperalta:


src/generator/operations.ts
```diff
@@ -726,6 +731,11 @@ function createInterfaceDefinition(group: OperationGroup, imports: ImportManager
     }
     const returns = generateReturnsInfo(op, false);
     interfaceText += `\t${opName}(${getAPIParametersSig(op, imports)}) (${returns.join(', ')})\n`;
+    // Add resume LRO poller method for each Begin poller method
+    if (isLROOperation(op) && !op.extensions!['x-ms-pageable']) {
```
jhendrixMSFT, line 735: Why not for `x-ms-pageable`, or will this be removed once we fix intersection of LROs and pageables?

catalinaperalta: Yes it will be fixed once I add pagers. Here's the pending issue #362. Also I'll add that after I finish adding all the tests. It shouldnt be too complicated now that everything else is in place.

src/generator/operations.ts
```diff
@@ -726,6 +731,11 @@ function createInterfaceDefinition(group: OperationGroup, imports: ImportManager
...
+      interfaceText += `\t// Resume${pascalCase(op.language.go!.pollerType.name)} - Used to create a new poller from a resume token of a previously implemented poller\n`;
```
jhendrixMSFT, line 736: "Used to create a new instance of this poller from the resume token of a previous instance of this poller type."  Something like that maybe.

src/generator/operations.ts
```diff
@@ -892,3 +902,42 @@ function generateReturnsInfo(op: Operation, forHandler: boolean): string[] {
...
+  text += `\t}\n`;
+  text += `\t// now unmarshal into the tracker\n`;
+  text += `\terr = json.Unmarshal([]byte(id), &poller.pt)\n`;
+  text += `\tif err != nil {\n`;
+  text += `\t\treturn nil, err\n`;
+  text += `\t}\n`;
+  text += `\treturn poller, nil\n`;
+  text += `}\n`;
+  return text;
+}
```
jhendrixMSFT, line 943: Missing new-line.

test/autorest/generated/lrogroup/lroretrys.go
```diff
@@ -56,6 +73,40 @@ func (client *lroRetrysOperations) BeginDelete202Retry200(ctx context.Context) (
...
+	case http.MethodDelete:
+		poller.pt = &pollingTrackerDelete{}
+	case http.MethodPatch:
+		poller.pt = &pollingTrackerPatch{}
+	case http.MethodPost:
+		poller.pt = &pollingTrackerPost{}
+	case http.MethodPut:
+		poller.pt = &pollingTrackerPut{}
+	default:
+		return nil, fmt.Errorf("ResumeLroRetrysDelete202Retry200Poller: unsupoorted method '%s'", method)
```
jhendrixMSFT, line 100: "unsupported"

test/autorest/generated/lrogroup/lroretrys.go
```diff
@@ -56,6 +73,40 @@ func (client *lroRetrysOperations) BeginDelete202Retry200(ctx context.Context) (
 	}, nil
 }
 
+func (client *lroRetrysOperations) ResumeLroRetrysDelete202Retry200Poller(id string) (LroRetrysDelete202Retry200Poller, error) {
```
jhendrixMSFT, line 76: What would happen if we called this with the token from a different poller type?  We should probably guard against this by embedding the poller type name in the JSON and making sure they match.  It doesn't have to be fixed now if there are more pressing things, just open an issue and we can fix it later.

jhendrixMSFT: We might also need to include some kind of version number as well to guard against the scenario where there are different SDK versions on each end.

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: Done

src/generator/operations.ts

src/generator/operations.ts

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: Hmmm ok, so I think I'll fix it as soon as I finish the tests. If I manage to finish the tests I'll add that today, if not then Monday morning. Here's the issue #372

PR #380, opened 2020-05-04
Split response error unmarshalling into its own method

jhendrixMSFT:
Generate a '<operation>HandleError' method that unmarshalls the error
payload, and wired up the handler into the LRO polling tracker.
Removed constructors for error types.
Moved resuming pollers common code into helpers.
Refactored body of Poll() method to avoid 'go vet' complaining.

### DIFF
--- a/test/autorest/generated/lrogroup/lrosads.go
+++ b/test/autorest/generated/lrogroup/lrosads.go
@@ -7,12 +7,9 @@ package lrogroup
 
 import (
 	"context"
-	"encoding/json"
-	"fmt"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"net/http"
 	"strconv"
-	"strings"
 )
 
 // LrosaDsOperations contains the methods for the LrosaDs group.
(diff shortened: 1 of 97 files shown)


test/autorest/generated/lrogroup/lrosads.go
```diff
@@ -211,6 +185,15 @@ func (client *lrosaDsOperations) delete202NonRetry400HandleResponse(resp *azcore
...
 }
 
+// delete202NonRetry400HandleError handles the Delete202NonRetry400 error response.
+func (client *lrosaDsOperations) delete202NonRetry400HandleError(resp *azcore.Response) error {
+	err := CloudError{}
+	if err := resp.UnmarshalAsJSON(&err); err != nil {
+		return err
+	}
+	return err
+}
```
catalinaperalta, line 195: Does this unmarshal correctly now?

jhendrixMSFT: Oh was it broken someplace?

catalinaperalta: If I recall correctly. Some testserver tests have errors that are not in the CloudError format

jhendrixMSFT: I see, sadly I'm not surprised.  This shouldn't fix any of those cases, we'd either need to update the swagger or the test server so the definitions match.

catalinaperalta: Yes probably. I'll still take a look when I go over the missing tests.

PR #382, opened 2020-05-05
Adding work for LRO pagers

catalinaperalta:
This PR adds support for LROs that return pageable types. 
Moves some functions in operations.ts to helpers.ts to be shared among different files.


src/common/helpers.ts
```diff
@@ -47,6 +47,7 @@ export interface PollerInfo {
   operationName: string;
```
jhendrixMSFT, line 47: With the addition of `op` I think you can remove `operationName` and `schema`.

src/generator/pollers.ts
```diff
@@ -26,26 +26,52 @@ export async function generatePollers(session: Session<CodeModel>): Promise<stri
...
   for (const poller of values(pollers)) {
     const pollerInterface = pascalCase(poller.name);
     let responseType = '';
     let rawResponse = ''; // used to access the raw response field on response envelopes
     if (poller.schema === undefined) {
-      responseType = 'http.Response';
+      responseType = '*http.Response';
     } else {
-      responseType = poller.schema.language.go!.responseType.name;
+      if (poller.op.language.go!.pageableType) {
```
jhendrixMSFT, line 40: This `if/else` block can be condensed into the block above.

catalinaperalta: Done

src/common/helpers.ts
catalinaperalta: Done

test/autorest/generated/paginggroup/pollers.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Response returns the latest response that is stored from the latest polling operation
+func (p *pagingGetMultiplePagesLroPoller) Response() (ProductResultPager, error) {
+	client := p.client
+	u, err := url.Parse(p.pt.pollingURL())
+	if err != nil {
+		return nil, err
+	}
+	req := azcore.NewRequest(http.MethodGet, *u)
```
jhendrixMSFT, line 50: I'm surprised we have to make another request here, i.e. when the poller reaches a successful terminal state I would expect we'd already have the first page of data.  I realize though that this is incongruent with how pagers work in the non-LRO case.  In addition, the polling URL isn't guaranteed to return the result, it's service-dependent (the "final GET" URL).

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/362

catalinaperalta:
Closing, a lot of the work here has changed with recent changes.

PR #383, opened 2020-05-05
Fix variable reference

jhendrixMSFT:


eng/pipelines/ci.yml
```diff
@@ -100,5 +100,5 @@ jobs:
             $currentVersion = node -p -e "require('./src/package.json').version";
             $currentVersion="$currentVersion-$(Build.BuildNumber)";
             npm pack;
-            npx publish-release --token $(github-token) --repo autorest.go --owner azure --name "Autorest for Go v$(currentVersion)" --tag v$(currentVersion) --notes='Preview version of Autorest for Go track 2' --prerelease --editRelease false --assets autorest-go-$(currentVersion).tgz --target_commitish $(Build.SourceBranchName);
```
catalinaperalta, line 103: Why didn't it work with the ()?

jhendrixMSFT: For vars defined in the YAML we need to use `$()`, but `$currentVersion` is defined in the script so we reference it like a regular PS var.

PR #385, opened 2020-05-05
Change path during publishing

jhendrixMSFT:


eng/pipelines/ci.yml
```diff
@@ -99,6 +99,7 @@ jobs:
         - pwsh: |
             $currentVersion = node -p -e "require('./src/package.json').version";
             $currentVersion="$currentVersion-$(Build.BuildNumber)";
+            cd src
```
catalinaperalta, line 102: Why are we doing this?

catalinaperalta: To only publish relevant generator files?

jhendrixMSFT: The package.json for this project is under the `src` directory.

catalinaperalta: Ah got it. Thank you!

PR #388, opened 2020-05-06
Move publishing dev releases to its own pipeline

jhendrixMSFT:


eng/pipelines/publish-dev-release.yml
```diff
@@ -0,0 +1,90 @@
+trigger: none
```
catalinaperalta, line 1: How do we trigger this manually?

jhendrixMSFT: It will be through the Azure DevOps page, I'll send a link once I have the pipeline created.

PR #390, opened 2020-05-06
Fix Response() and done() to update with the latest error

catalinaperalta:
This PR moves done() to poller_helpers.go and updates it to only return a boolean value, as well as ensuring that the error on the poller is updated to the last error returned. Response() now inlcudes a check for pollingErrors() before checking for a response.

### DIFF
--- a/src/generator/pollers.ts
+++ b/src/generator/pollers.ts
@@ -42,22 +42,22 @@ export async function generatePollers(session: Session<CodeModel>): Promise<stri
     }
     text += `// ${pollerInterface} provides polling facilities until the operation completes
 type ${pollerInterface} interface {
-	Poll(context.Context) bool
-	Response() (*${responseType}, error)
-	ResumeToken() (string, error)
-	Wait(ctx context.Context, pollingInterval time.Duration) (*${responseType}, error)
+  Poll(context.Context) bool
+  Response() (*${responseType}, error)
+  ResumeToken() (string, error)
+  Wait(ctx context.Context, pollingInterval time.Duration) (*${responseType}, error)
 }
 
 type ${poller.name} struct {
-	// the client for making the request
-	client *${poller.client}
-	// polling tracker
-	pt pollingTracker
+  // the client for making the request
+  client *${poller.client}
+  // polling tracker
+  pt pollingTracker
 }
 
 // Poll returns false if there was an error or polling has reached a terminal state
 func (p *${poller.name}) Poll(ctx context.Context) bool {
-	return !done(ctx, p.client.p, p.pt)
+  return !lroPollDone(ctx, p.client.p, p.pt)
 }
 
 // Response returns the latest response that is stored from the latest polling operation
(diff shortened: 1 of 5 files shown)


src/generator/pollers.ts
```diff
@@ -455,7 +434,35 @@ export async function generatePollersHelper(session: Session<CodeModel>): Promis
 	type pollingTrackerDelete struct {
 		pollingTrackerBase
 	}
-	
+  
+  // done queries the service to see if the operation has completed.
+  func (pt *pollingTrackerDelete) done(ctx context.Context, p azcore.Pipeline) bool {
```
jhendrixMSFT, line 439: Can this be moved to `pollingTrackerBase` to avoid the duplication?  Probably need to add it to the interface.

catalinaperalta: I tried that initially, but the methods called inside like updatePollingMethod() are implemented on the trackers and it marks an error when attempting to place it on the pollingTrackerBase.

catalinaperalta: I could add similar methods on pollingTrackerBase

jhendrixMSFT: Oh yes good point.  What about making `done()` a func instead that takes a `pollingTracker` interface as its third argument?

catalinaperalta: Yes that could work, I'll do that then

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -173,14 +194,16 @@ func (pt *pollingTrackerBase) updateRawBody() error {
 		defer pt.resp.Body.Close()
 		b, err := ioutil.ReadAll(pt.resp.Body)
 		if err != nil {
-			return errors.New("failed to read response body")
+			pt.Err = errors.New("failed to read response body")
```
jhendrixMSFT, line 197: This should wrap the error returned from `ioutil.ReadAll()`.

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -173,14 +194,16 @@ func (pt *pollingTrackerBase) updateRawBody() error {
...
+			return pt.Err
 		}
 		// observed in 204 responses over HTTP/2.0; the content length is -1 but body is empty
 		if len(b) == 0 {
 			return nil
 		}
 		if err = json.Unmarshal(b, &pt.rawBody); err != nil {
-			return errors.New("failed to unmarshal response body")
+			pt.Err = errors.New("failed to unmarshal response body")
```
jhendrixMSFT, line 205: Wrap error.

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -189,13 +212,15 @@ func (pt *pollingTrackerBase) updateRawBody() error {
...
 	if err != nil {
+		pt.Err = err
 		return err
 	}
 	req := azcore.NewRequest(http.MethodGet, *u)
 	resp, err := client.Do(ctx, req)
 	pt.resp = resp
 	if err != nil {
-		return errors.New("failed to send HTTP request")
+		pt.Err = errors.New("failed to send HTTP request")
```
jhendrixMSFT, line 222: Wrap error, or maybe just omit wrapping and use the returned error?

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -229,7 +254,8 @@ func (pt *pollingTrackerBase) updatePollingState(provStateApl bool) error {
 				pt.State = operationSucceeded
 			}
 		} else {
-			return errors.New("the response from the async operation has an invalid status code")
+			pt.Err = errors.New("the response from the async operation has an invalid status code")
```
jhendrixMSFT, line 257: Include the invalid status code (missed this in track 1).

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -31,7 +31,7 @@ type lrOSCustomHeaderPost202Retry200Poller struct {
 
 // Poll returns false if there was an error or polling has reached a terminal state
 func (p *lrOSCustomHeaderPost202Retry200Poller) Poll(ctx context.Context) bool {
-	return !p.pt.done(ctx, p.client.p)
+	return !done(ctx, p.client.p, p.pt)
```
jhendrixMSFT, line 34: Since this is no longer a method on the tracker we might want to give this a less generic name, maybe `lroDone` or `lroTerminated` or something.  It's internal so the name doesn't really matter, but might be nice.

PR #393, opened 2020-05-07
Fix operation groups with no name

jhendrixMSFT:
For operations without a group name, use the pascal-cased swagger title.
Consolidated params in PagerInfo and PollerInfo.
Added clientName field to operations and consolidated its usage.
Fixed tests where operation group names changed.
Fixed a bug in poller's Wait() method to use the caller-supplied
context.
Updated to the latest M4 which includes the deduplication switch.

### DIFF
--- a/src/generator/pagers.ts
+++ b/src/generator/pagers.ts
@@ -5,11 +5,12 @@
 
 import { Session } from '@azure-tools/autorest-extension-base';
 import { camelCase } from '@azure-tools/codegen';
-import { CodeModel } from '@azure-tools/codemodel';
+import { CodeModel, SchemaResponse } from '@azure-tools/codemodel';
 import { values } from '@azure-tools/linq';
 import { PagerInfo } from '../common/helpers';
 import { contentPreamble, sortAscending } from './helpers';
 import { ImportManager } from './imports';
+import { open } from 'fs';
 
 // Creates the content in pagers.go
 export async function generatePagers(session: Session<CodeModel>): Promise<string> {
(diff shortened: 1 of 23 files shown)


src/generator/pagers.ts
```diff
@@ -5,11 +5,12 @@
 
 import { Session } from '@azure-tools/autorest-extension-base';
 import { camelCase } from '@azure-tools/codegen';
-import { CodeModel } from '@azure-tools/codemodel';
+import { CodeModel, SchemaResponse } from '@azure-tools/codemodel';
 import { values } from '@azure-tools/linq';
 import { PagerInfo } from '../common/helpers';
 import { contentPreamble, sortAscending } from './helpers';
 import { ImportManager } from './imports';
+import { open } from 'fs';
```
catalinaperalta, line 13: I dont think this is actuall used, the same thing happened to me when I typed op.

jhendrixMSFT: Ugh I hate that, will fix.

src/transform/namer.ts
```diff
@@ -67,17 +67,22 @@ export async function namer(session: Session<CodeModel>) {
...
+    const groupDetails = <Language>group.language.go;
+    // use the swagger title as the default name for operation groups that don't specify a group name
+    if (groupDetails.name.length === 0) {
+      groupDetails.name = pascalCase(session.model.info.title);
+    }
+    groupDetails.name = capitalizeAcronyms(pascalCase(groupDetails.name));
+    groupDetails.clientName = `${groupDetails.name}Operations`;
+    if (groupDetails.name === 'Operations') {
+      // if the group name is 'Operations' don't name it 'OperationsOperations'
+      groupDetails.clientName = groupDetails.clientName;
```
jhendrixMSFT, line 79: This is wrong.

src/transform/namer.ts
```diff
@@ -67,17 +67,22 @@ export async function namer(session: Session<CodeModel>) {
...
-    details.name = capitalizeAcronyms(pascalCase(details.name));
-    // we don't call GetEscapedReservedName here since any operation group that uses a reserved word will have 'Operations' attached to it
-    details.clientName = `${details.name}Operations`;
-    // this sets Operations as the default name for operation groups that don't specify a group name
-    if (length(details.name) === 0) {
-      details.name = `Operations`;
+    const groupDetails = <Language>group.language.go;
+    // use the swagger title as the default name for operation groups that don't specify a group name
+    if (groupDetails.name.length === 0) {
+      groupDetails.name = pascalCase(session.model.info.title);
```
catalinaperalta, line 73: Isn't the call to pascalCase repetitive here since we call it directly below as well?

jhendrixMSFT: Yep, will fix.

Issue #378, opened 2020-05-01
Design/implement method to include access to the latest LRO response


catalinaperalta:
Suggestion: 
Have LatestResponse() and Result() methods on the poller

PR #396, opened 2020-05-14
Updating LRO design

catalinaperalta:
This PR includes:
- update to the LRO interface
- adding a poller specific response handler
- commenting out tests related to LROs since tests will all need to be updated to reflect the new interface design 
- fix the handleresponse method for each operation


src/generator/models.ts
```diff
@@ -404,7 +404,11 @@ function generateStruct(lang: Language, props?: Property[]): StructDef {
     imports.add('fmt');
   }
   if (lang.responseType) {
-    imports.add("net/http");
+    imports.add('net/http');
+  }
+  if (lang.needsTimeAndContext) {
```
jhendrixMSFT, line 409: I wonder if this would be better named `lang.isLRO` (or something similar)?  A bit more specific to the type.

src/generator/operations.ts
```diff
@@ -284,17 +285,47 @@ function generateOperation(clientName: string, op: Operation, imports: ImportMan
     text += `\tif err != nil {\n`;
     text += `\t\treturn nil, err\n`;
     text += `\t}\n`;
+    if (!op.responses) {
+      text += '\tresult := &HttpResponse{\n';
```
jhendrixMSFT, line 289: This should be `HTTPResponse` else the linter will complain.

src/generator/models.ts
catalinaperalta: Yes I like that more, I was worried about having to add in other imports in the future, will update the name.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -11,4536 +11,73 @@ import (
...
-	resp := p.response()
-	if resp == nil {
-		return nil, errors.New("did not find a response on the poller")
+func (p *httpPoller) FinalResponse(ctx context.Context) (*HttpResponse, error) {
+	if p.pt.finalGetURL() == "" {
+		// we can end up in this situation if the async operation returns a 200
+		// with no polling URLs.  in that case return the response which should
+		// contain the JSON payload (only do this for successful terminal cases).
+		if lr := p.pt.latestResponse(); lr != nil && p.pt.hasSucceeded() {
+			result, err := httpPollerHandleResponse(lr, p.pipeline)
```
jhendrixMSFT, line 51: Won't we be calling the handler on the client?

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -4551,200 +88,84 @@ func (p *lrosaDsPut200InvalidJSONPoller) ResumeToken() (string, error) {
...
+func httpPollerHandleResponse(resp *azcore.Response, p azcore.Pipeline) (*HttpResponse, error) {
+	result := &HttpResponse{RawResponse: resp.Response}
+	return result, nil
 }
 
-// response returns the last HTTP response.
-func (p *lrosaDsPut200InvalidJSONPoller) response() *azcore.Response {
-	return p.pt.latestResponse()
+// httpPoller handles the error response.
+func httpPollerHandleError(resp *azcore.Response) error {
```
jhendrixMSFT, line 97: This doesn't appear to be referenced.

src/generator/models.ts
catalinaperalta: Done

src/generator/operations.ts
catalinaperalta: Fixed

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Well we could but the handlers are specific to the poller response types. I think it will reduce code for it to be part of the pollers generation. Since the response handling should be the same for responses that use the same models. I could add it on the poller. Or why do you think it should be on the client?

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: You're right, I updated it. It's not being referenced when creating the polling trackers.

test/autorest/generated/lrogroup/pollers.go
jhendrixMSFT: It's already on the client though right?  I was suggesting we use the one that's already there.

catalinaperalta: Well the ones that are on the client are like xOperationHandleResponse (like delete202HandleResponse), but since these pollers need to be common among all operations, then we cant really use each operation's handle response method.

test/autorest/generated/lrogroup/pollers.go
jhendrixMSFT: Can't we use the one on the client and omit this one?

catalinaperalta: Yes we can, I cant remember why this was separated.

test/autorest/generated/lrogroup/lroretrys.go
```diff
@@ -93,20 +90,11 @@ func (client *lroRetrysOperations) delete202Retry200CreateRequest() (*azcore.Req
...
-	}
-	if val := resp.Header.Get("Retry-After"); val != "" {
-		retryAfter32, err := strconv.ParseInt(val, 10, 32)
-		retryAfter := int32(retryAfter32)
-		if err != nil {
-			return nil, err
-		}
-		result.RetryAfter = &retryAfter
+func (client *lroRetrysOperations) delete202Retry200HandleResponse(resp *azcore.Response) (*HTTPResponse, error) {
+	if !resp.HasStatusCode(pollingCodes[:]...) {
```
jhendrixMSFT, line 94: This should be the list of status codes defined by the swagger.

test/autorest/generated/lrogroup/pollers.go
jhendrixMSFT: Presumably because of the different set of status codes for success.  We can place a func on the poller that's set to the operation's response handler as they should all have the same signature, just different bodies.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -11,4536 +11,73 @@ import (
...
-// duration specified in the retry-after header, if the header is not specified then the pollingInterval that
-// is specified will be used to wait between polling requests.
-func (p *lrosaDsPostNonRetry400Poller) Wait(ctx context.Context, pollingInterval time.Duration) (*LrosaDsPostNonRetry400Response, error) {
-	for p.Poll(ctx) {
-		if delay := p.response().RetryAfter(); delay > 0 {
-			time.Sleep(delay)
-		} else {
-			time.Sleep(pollingInterval)
+	resp, err := p.pipeline.Do(ctx, req)
+	if err == nil && resp.Body != nil {
```
jhendrixMSFT, line 68:
```go
	resp, err := p.pipeline.Do(ctx, req)
	if err != nil {
		return nil, err
	}
	return subProductPollerHandleResponse(resp, p.pipeline)
```

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: Done

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Ok, will do.

src/transform/transform.ts
```diff
@@ -483,19 +526,51 @@ function createResponseType(codeModel: CodeModel, group: OperationGroup, op: Ope
...
+          declareResume: false,
+          op: op,
+        };
+        op.language.go!.pollerType = tempPoller;
+        return;
+      }
+    }
     // Adding the operation group name to the poller name for polling operations that need to be unique to that operation group
-    const name = `${camelCase(group.language.go!.name)}${op.language.go!.name}Poller`;
+    // const name = `${camelCase(group.language.go!.name)}${op.language.go!.name}Poller`;
```
jhendrixMSFT, line 552: Remove?

src/transform/transform.ts
```diff
@@ -483,19 +526,51 @@ function createResponseType(codeModel: CodeModel, group: OperationGroup, op: Ope
     if (codeModel.language.go!.pollerTypes === undefined) {
       codeModel.language.go!.pollerTypes = new Array<PollerInfo>();
     }
+    let type = 'HTTP';
```
jhendrixMSFT, line 529: Took me a bit to figure out what this chunk of code is doing, might be worth a comment.  Also, I think you can simplify it a bit.
```ts
let type = 'HTTP';
if (isSchemaResponse(firstResp) && firstResp.schema.language.go!.responseType.value) {
  type = firstResp.schema.language.go!.responseType.value;
}
```

test/autorest/generated/lrogroup/lroretrys.go
```diff
@@ -60,25 +46,36 @@ func (client *lroRetrysOperations) BeginDelete202Retry200(ctx context.Context) (
...
-	pt, err := resumePollingTracker("lroRetrysDelete202Retry200Poller", token, client.delete202Retry200HandleError)
+	pt, err := createPollingTracker("HTTPPoller", resp, client.delete202Retry200HandleError)
 	if err != nil {
 		return nil, err
 	}
-	return &lroRetrysDelete202Retry200Poller{
-		client: client,
-		pt:     pt,
-	}, nil
+	result.GetPoller = func() HTTPPoller {
```
jhendrixMSFT, line 57: I just had a thought.  Does this really need to be a func, couldn't it just be a field of type `HTTPPoller`?  Then we can do this.
```go
	result.GetPoller = &httpPoller{
		pt:       pt,
		pipeline: client.p,
	}
```
Something to discuss in scrum tomorrow.

src/transform/transform.ts
catalinaperalta: Will add a comment about this

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: Yes I dont see why not.

src/generator/operations.ts
```diff
@@ -746,9 +782,9 @@ function createInterfaceDefinition(group: OperationGroup, imports: ImportManager
     const returns = generateReturnsInfo(op, false);
     interfaceText += `\t${opName}(${getAPIParametersSig(op, imports)}) (${returns.join(', ')})\n`;
     // Add resume LRO poller method for each Begin poller method
-    if (isLROOperation(op) && !op.extensions!['x-ms-pageable']) {
-      interfaceText += `\t// Resume${pascalCase(op.language.go!.pollerType.name)} - Used to create a new instance of this poller from the resume token of a previous instance of this poller type.\n`;
-      interfaceText += `\tResume${pascalCase(op.language.go!.pollerType.name)}(id string) (${pascalCase(op.language.go!.pollerType.name)}, error)\n`;
+    if (isLROOperation(op) && !op.extensions!['x-ms-pageable'] && op.language.go!.pollerType.declareResume) {
+      interfaceText += `\t// Resume${op.language.go!.pollerType.name} - Used to create a new instance of this poller from the resume token of a previous instance of this poller type.\n`;
+      interfaceText += `\tResume${op.language.go!.pollerType.name}(id string) (${op.language.go!.pollerType.name}, error)\n`;
```
jhendrixMSFT, line 787: `(token string)` for param.

test/autorest/generated/lrogroup/pollers.go

src/transform/transform.ts
catalinaperalta: Done

test/autorest/generated/lrogroup/lroretrys.go
catalinaperalta: Done

src/generator/operations.ts

src/transform/transform.ts
catalinaperalta: Done

test/autorest/generated/lrogroup/pollers.go

catalinaperalta:
Adding an issue to track the status code check in the response handlers: #399

src/generator/operations.ts
```diff
@@ -605,15 +622,22 @@ function createProtocolResponse(client: string, op: Operation, imports: ImportMa
...
+    } else if (isLROOperation(op)) {
+      statusCodes = statusCodes.concat(op.responses[i].protocol.http?.statusCodes);
     }
   }
-  // LROs will skip this check since the status code is checked by the poller
-  if (!isLROOperation(op)) {
-    text += `\tif !resp.HasStatusCode(${formatStatusCodes(statusCodes)}) {\n`;
-    text += `\t\treturn nil, client.${info.protocolNaming.errorMethod}(resp)\n`;
-    text += '\t}\n';
+  if (isLROOperation(op) && statusCodes.find(element => element == '204') == undefined) {
```
jhendrixMSFT, line 629: Should be `===` for comparison check.

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/378

PR #397, opened 2020-05-18
Refactor discriminated type generation

jhendrixMSFT:
Use composition for discriminated types hierarchy.
Change marker method to an exported GetType method used to retrieve the
base-type content.
Removed internal enums for discriminator values.
Split marshallers and unmarshallers for discriminated types separate
from counterparts for time and XML wrappers.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -304,14 +304,17 @@ function generateUnmarshallerForResponseEnvelope(structDef: StructDef) {
   let unmarshaller = `func (${receiver} *${structDef.Language.name}) UnmarshalJSON(data []byte) error {\n`;
   // add a custom unmarshaller to the response envelope
   // find the discriminated type field
-  let field = 'FIND';
-  let type = 'FIND';
+  let field = '';
+  let type = '';
   for (const prop of values(structDef.Properties)) {
     if (prop.isDiscriminator) {
       field = prop.language.go!.name;
       type = prop.schema.language.go!.discriminatorInterface;
       break;
     }
+  }
+  if (field === '' || type === '') {
+    throw console.error(`failed to the discriminated type field for response envelope ${structDef.Language.name}`);
   }
   unmarshaller += `\tt, err := unmarshal${type}(data)\n`;
   unmarshaller += '\tif err != nil {\n';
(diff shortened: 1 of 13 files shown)


src/generator/models.ts
```diff
@@ -423,3 +287,229 @@ function generateParamGroupStruct(lang: Language, params: Parameter[]): StructDe
...
+    }
+  }
+  if (!found) {
+    return;
+  }
+  const receiver = structDef.Language.name[0].toLowerCase();
+  let unmarshaller = `func (${receiver} *${structDef.Language.name}) UnmarshalJSON(data []byte) error {\n`;
+  // add a custom unmarshaller to the response envelope
+  // find the discriminated type field
+  let field = 'FIND';
```
catalinaperalta, line 307: Is FIND the default name for the discriminated type field?

catalinaperalta: Or will this never happen and is just a filler value? If so, how do we know that this never happens?

jhendrixMSFT: It's a filler value and should never happen.  In the event that it does, the resultant code won't compile.

jhendrixMSFT: Although we could rework this part to throw an error in that case instead (might be better).

catalinaperalta: Hmmm maybe throwing would be better, then we know immediately that there's a problem with a swagger?

jhendrixMSFT: I will change it to throw (it's likely less a swagger issue, more likely a bug in the generator).

jhendrixMSFT: Fixed

src/generator/models.ts
```diff
@@ -423,3 +287,232 @@ function generateParamGroupStruct(lang: Language, params: Parameter[]): StructDe
...
+  unmarshaller += '\t}\n';
+  if (!obj.language.go!.rootDiscriminator && parentType) {
+    unmarshaller += `\treturn json.Unmarshal(data, &${receiver}.${parentType.language.go!.name})\n`;
+  } else {
+    unmarshaller += '\treturn nil\n';
+  }
+  unmarshaller += '}\n\n';
+  // there's a corner-case where a derived type might not add any new fields (Cookiecuttershark).
+  // in this case skip adding the unmarshaller as it's not necessary and doesn't compile.
+  if (structDef.Properties && structDef.Properties.length > 0) {
```
catalinaperalta, line 426: Wouldnt this be better at the beginning of the function? To avoid processing all of that work for this case where the unmarshaller isnt needed?

jhendrixMSFT: Fixed

PR #400, opened 2020-05-20
Final response handler for Pollers and updating tests

catalinaperalta:
This PR removes the operation specific response handler for handling final responses on pollers and adds some updated lro tests based on the new behavior.

### DIFF
--- a/test/autorest/generated/lrogroup/pollers.go
+++ b/test/autorest/generated/lrogroup/pollers.go
@@ -19,7 +19,7 @@ import (
 type HTTPPoller interface {
 	Done() bool
 	Poll(ctx context.Context) (*http.Response, error)
-	FinalResponse(ctx context.Context) (*HTTPResponse, error)
+	FinalResponse() *http.Response
 	ResumeToken() (string, error)
 }
 
@@ -43,33 +43,8 @@ func (p *httpPoller) Poll(ctx context.Context) (*http.Response, error) {
 	return nil, p.pt.pollingError()
 }
 
-func (p *httpPoller) FinalResponse(ctx context.Context) (*HTTPResponse, error) {
-	if p.pt.finalGetURL() == "" {
-		// we can end up in this situation if the async operation returns a 200
-		// with no polling URLs.  in that case return the response which should
-		// contain the JSON payload (only do this for successful terminal cases).
-		if lr := p.pt.latestResponse(); lr != nil && p.pt.hasSucceeded() {
-			result, err := p.handleResponse(lr)
-			if err != nil {
-				return nil, err
-			}
-			return result, nil
-		}
-		return nil, errors.New("missing URL for retrieving result")
-	}
-	u, err := url.Parse(p.pt.finalGetURL())
-	if err != nil {
-		return nil, err
-	}
-	req := azcore.NewRequest(http.MethodGet, *u)
-	if err != nil {
-		return nil, err
-	}
-	resp, err := p.pipeline.Do(ctx, req)
-	if err != nil {
-		return nil, err
-	}
-	return p.handleResponse(resp)
+func (p *httpPoller) FinalResponse() *http.Response {
+	return p.pt.latestResponse().Response
 }
 
 // ResumeToken generates the string token that can be used with the ResumeHTTPPoller method
--- a/test/autorest/generated/lrogroup/pollers.go
+++ b/test/autorest/generated/lrogroup/pollers.go
@@ -61,10 +61,13 @@ func (p *httpPoller) ResumeToken() (string, error) {
 }
 
 func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.Duration) (*http.Response, error) {
-	for !p.Done() {
+	for {
 		resp, err := p.Poll(ctx)
 		if err != nil {
 			return nil, err
+		}
+		if p.Done() {
+			break
 		}
 		if delay := azcore.RetryAfter(resp); delay > 0 {
 			time.Sleep(delay)
@@ -73,14 +76,6 @@ func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.D
 		}
 	}
 	return p.FinalResponse(), nil
-}
-
-func (p *httpPoller) handleResponse(resp *azcore.Response) (*HTTPResponse, error) {
-	result := HTTPResponse{RawResponse: resp.Response}
-	if !resp.HasStatusCode(pollingCodes[:]...) {
-		return nil, p.pt.handleError(resp)
-	}
-	return &result, nil
 }
 
 // ProductPoller provides polling facilities until the operation completes
@@ -154,10 +149,13 @@ func (p *productPoller) ResumeToken() (string, error) {
 }
 
 func productPollerPollUntilDone(ctx context.Context, p ProductPoller, frequency time.Duration) (*ProductResponse, error) {
-	for !p.Done() {
+	for {
 		resp, err := p.Poll(ctx)
 		if err != nil {
 			return nil, err
+		}
+		if p.Done() {
+			break
 		}
 		if delay := azcore.RetryAfter(resp); delay > 0 {
 			time.Sleep(delay)
@@ -250,10 +248,13 @@ func (p *skuPoller) ResumeToken() (string, error) {
 }
 
 func skuPollerPollUntilDone(ctx context.Context, p SkuPoller, frequency time.Duration) (*SkuResponse, error) {
-	for !p.Done() {
+	for {
 		resp, err := p.Poll(ctx)
 		if err != nil {
 			return nil, err
+		}
+		if p.Done() {
+			break
 		}
 		if delay := azcore.RetryAfter(resp); delay > 0 {
 			time.Sleep(delay)
@@ -346,10 +347,13 @@ func (p *subProductPoller) ResumeToken() (string, error) {
 }
 
 func subProductPollerPollUntilDone(ctx context.Context, p SubProductPoller, frequency time.Duration) (*SubProductResponse, error) {
-	for !p.Done() {
+	for {
 		resp, err := p.Poll(ctx)
 		if err != nil {
 			return nil, err
+		}
+		if p.Done() {
+			break
 		}
 		if delay := azcore.RetryAfter(resp); delay > 0 {
 			time.Sleep(delay)
--- a/test/autorest/generated/lrogroup/pollers.go
+++ b/test/autorest/generated/lrogroup/pollers.go
@@ -30,8 +30,6 @@ type httpPoller struct {
 	pipeline azcore.Pipeline
 	// polling tracker
 	pt pollingTracker
-	// use the response handler to check for accepted status codes
-	response httpResponseHandleResponse
 }
 
 // Done returns true if there was an error or polling has reached a terminal state
@@ -53,7 +51,7 @@ func (p *httpPoller) FinalResponse(ctx context.Context) (*HTTPResponse, error) {
 		// with no polling URLs.  in that case return the response which should
 		// contain the JSON payload (only do this for successful terminal cases).
 		if lr := p.pt.latestResponse(); lr != nil && p.pt.hasSucceeded() {
-			result, err := p.response(lr)
+			result, err := httpPollerHandleResponse(lr)
 			if err != nil {
 				return nil, err
 			}
@@ -73,7 +71,7 @@ func (p *httpPoller) FinalResponse(ctx context.Context) (*HTTPResponse, error) {
 	if err != nil {
 		return nil, err
 	}
-	return p.response(resp)
+	return httpPollerHandleResponse(resp)
 }
 
 // ResumeToken generates the string token that can be used with the ResumeHTTPPoller method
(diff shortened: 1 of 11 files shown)


test/autorest/generated/lrogroup/pollers.go
```diff
@@ -104,6 +102,12 @@ func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.D
 	return p.FinalResponse(ctx)
 }
 
+func httpPollerHandleResponse(resp *azcore.Response) (*HTTPResponse, error) {
+	result := HTTPResponse{RawResponse: resp.Response}
+
```
jhendrixMSFT, line 107: Nit, extra new-line.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -104,6 +102,12 @@ func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.D
 	return p.FinalResponse(ctx)
 }
 
+func httpPollerHandleResponse(resp *azcore.Response) (*HTTPResponse, error) {
```
jhendrixMSFT, line 105: Since these are specific to the poller type I was thinking these should be methods on the matching pollers.  It's all internal code though so it doesn't make much difference.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -193,6 +195,14 @@ func productPollerPollUntilDone(ctx context.Context, p ProductPoller, frequency
 	return p.FinalResponse(ctx)
 }
 
+func productPollerHandleResponse(resp *azcore.Response) (*ProductResponse, error) {
+	result := ProductResponse{RawResponse: resp.Response}
+	if resp.HasStatusCode(http.StatusNoContent) {
+		return &result, nil
+	}
+	return &result, resp.UnmarshalAsJSON(&result.Product)
```
jhendrixMSFT, line 203: This will end up unmarshalling on a non-successful status code which will likely fail as there's no response body.  At the least I'm thinking we should ensure that the status code here is `pollingCodes`, what do you think?

catalinaperalta: Mmm, I think there is a possibility that that could happen. I'll add a check with `pollingCodes` I think that's a good idea.

test/autorest/lrogroup/lros_test.go
```diff
@@ -96,421 +96,532 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
+	poller := resp.Poller
+	rt, err := poller.ResumeToken()
+	if err != nil {
+		t.Fatal(err)
+	}
+	poller, err = op.ResumeDelete202Retry200(rt)
+	if err != nil {
+		t.Fatal(err)
+	}
+	for !poller.Done() {
```
jhendrixMSFT, line 114: Since PollUntilDone() uses this machinery internally I don't think it's necessary to duplicate the test.  Just call PollUntilDone().

test/autorest/lrogroup/lros_test.go
```diff
@@ -96,421 +96,532 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
+	helpers.VerifyStatusCode(t, resp.RawResponse, 202)
+	resp1, err := op.BeginDeleteAsyncNoRetrySucceeded(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp, err = resp1.PollUntilDone(context.Background(), time.Duration(1)*time.Second)
+	if err != nil {
+		t.Fatal(err)
+	}
+	helpers.VerifyStatusCode(t, resp.RawResponse, 202)
```
jhendrixMSFT, line 245: This really ends with a 202?  Seems odd.

test/autorest/lrogroup/lros_test.go
```diff
@@ -96,421 +96,532 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
+	helpers.VerifyStatusCode(t, resp.RawResponse, 200)
+	resp1, err := op.BeginPost200WithPayload(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	resp, err = resp1.PollUntilDone(context.Background(), time.Duration(1)*time.Second)
+	if err != nil {
+		t.Fatal(err)
+	}
+	helpers.VerifyStatusCode(t, resp.RawResponse, 200)
```
jhendrixMSFT, line 547: Should verify the returned payload.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: So I had to add a handleError method on pollingTrackerBase. If you can go look at that change.

test/autorest/generated/lrogroup/pollers.go

test/autorest/generated/lrogroup/pollers.go

test/autorest/lrogroup/lros_test.go
catalinaperalta: Done

test/autorest/lrogroup/lros_test.go
catalinaperalta: Done

test/autorest/lrogroup/lros_test.go
catalinaperalta: The poller is not functioning correctly in this case :/ Looking into it

catalinaperalta: Fixed

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -101,7 +72,15 @@ func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.D
 			time.Sleep(frequency)
 		}
 	}
-	return p.FinalResponse(ctx)
+	return p.FinalResponse(), nil
+}
+
+func (p *httpPoller) handleResponse(resp *azcore.Response) (*HTTPResponse, error) {
```
jhendrixMSFT, line 78: It looks like this isn't actually used for things that return an `*http.Response`, can be omitted.

catalinaperalta: You're right will remove

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -89,19 +60,22 @@ func (p *httpPoller) ResumeToken() (string, error) {
 	return string(js), nil
 }
 
-func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.Duration) (*HTTPResponse, error) {
-	for !p.Done() {
+func httpPollerPollUntilDone(ctx context.Context, p HTTPPoller, frequency time.Duration) (*http.Response, error) {
```
jhendrixMSFT, line 63: I just realized this should be a method on `httpPoller` as it takes a specific poller arg (my bad for not noticing this when I made the original suggestion).

```go
func (p *httpPoller) pollUntilDone(ctx context.Context, frequency time.Duration) (*http.Response, error)
```

It will require a slight change when creating the `PollUntilDone` func however; either a type assertion or capture the `httpPoller` in a local first (the latter might be better as we know the type and can avoid using reflection).

catalinaperalta: So instead of having something like: 
`poller := result.Poller.(*httpPoller}` 
You're thinking of: 
`poller := &httpPoller{}`

catalinaperalta: ?

catalinaperalta: Ignore my comment, I see what you mean now

test/autorest/lrogroup/lros_test.go
```diff
@@ -37,22 +39,23 @@ func httpClientWithCookieJar() azcore.Transport {
...
-// 	if err != nil {
-// 		t.Fatal(err)
-// 	}
-// 	_, err = op.ResumeProductPoller(rt)
-// 	if err == nil {
-// 		t.Fatal("expected an error but did not find receive one")
-// 	}
-// }
+func TestLROResumeWrongPoller(t *testing.T) {
+	// TODO discuss what sort of check we want with the new design to avoid resuming from the wrong poller type
```
jhendrixMSFT, line 43: Is this comment still applicable?

test/autorest/lrogroup/lros_test.go
```diff
@@ -37,22 +39,23 @@ func httpClientWithCookieJar() azcore.Transport {
...
+	resp, err := op.BeginDelete202NoRetry204(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	poller := resp.Poller
+	rt, err := poller.ResumeToken()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = op.ResumePost200WithPayload(rt)
```
jhendrixMSFT, line 54: Might be good to capture the poller and ensure it's nil.

test/autorest/lrogroup/lros_test.go
catalinaperalta: Yes, we need to improve this functionality with the new changes you can resume pollers of the same type even if they were used for different operations. I'll add an issue as well

test/autorest/lrogroup/lros_test.go
```diff
@@ -69,1142 +72,974 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
 	}
-	for !poller.Done() {
-		_, err := poller.Poll(context.Background())
-		if err != nil {
-			t.Fatal(err)
-		}
-		time.Sleep(200 * time.Millisecond)
-	}
-	resp, err = poller.FinalResponse(context.Background())
+	resp, err = resp.PollUntilDone(context.Background(), time.Duration(1)*time.Millisecond)
```
jhendrixMSFT, line 75: Nit, casting `1` to `time.Duration` is unnecessary.

test/autorest/lrogroup/lros_test.go
catalinaperalta: Here's the issue #402

jhendrixMSFT: I see.  I think you can fix this by having the `method` info in the token be a combination of the receiver type and method name.  What do you think?

test/autorest/lrogroup/lros_test.go
catalinaperalta: Done

test/autorest/lrogroup/lros_test.go
jhendrixMSFT: Oops `method` is the wrong thing.  I think it would be `pollerType`.  You could change this to capture the `receiver.methodName` info.

test/autorest/lrogroup/lros_test.go
catalinaperalta: Done

test/autorest/lrogroup/lros_test.go
jhendrixMSFT: We can fix it in a separate PR if you prefer.

catalinaperalta: Something like ProductPollerDelete?

catalinaperalta: Yes I'll add it in my next PR, but is my previous comment something we could do? Would it be a sufficient check?

jhendrixMSFT: More like `lroRetrysOperations.Delete202Retry200`.  That way we know for sure that the poller is associated with this API.

catalinaperalta: Ok will do in the next PR 👍

test/autorest/lrogroup/lros_test.go
```diff
@@ -69,1142 +75,974 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
+	rt, err := poller.ResumeToken()
+	if err != nil {
+		t.Fatal(err)
+	}
+	poller, err = op.ResumePutAsyncRetryFailed(rt)
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, err = resp.PollUntilDone(context.Background(), 1*time.Millisecond)
+	if err == nil {
```
jhendrixMSFT, line 895: Should we be checking that this is a specific error type as defined in the swagger?

test/autorest/lrogroup/lros_test.go
```diff
@@ -69,1142 +75,974 @@ func TestLROBeginDelete202NoRetry204(t *testing.T) {
...
+		t.Fatal("expected an error but did not receive one")
+	}
+}
+
+func TestLROBeginPut201CreatingSucceeded200(t *testing.T) {
+	op := getLROSOperations(t)
+	resp, err := op.BeginPut201CreatingSucceeded200(context.Background(), nil)
```
jhendrixMSFT, line 707: It's interesting that none of these REST endpoints fail without a payload.  Might be good to add one for some of the APIs (in track 1 we just specified an empty `lrogroup.Product{}`).

catalinaperalta: Done I added it for a few of the tests.

catalinaperalta: And yes it's strange that they dont fail without a payload

test/autorest/lrogroup/lros_test.go
catalinaperalta: We should, I'll be adding this when we update the testserver version, since we currently have CloudError code that breaks because of an incorrectly named field.

PR #401, opened 2020-05-21
Fix an edge case that the list request may return empty value but non-null nextlink 

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/9510

### DIFF
--- a/test/src/tests/generated/paginggroup/models.go
+++ b/test/src/tests/generated/paginggroup/models.go
@@ -8,12 +8,11 @@ package paginggroup
 
 import (
 	"context"
-	"net/http"
-
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/Azure/go-autorest/tracing"
+	"net/http"
 )
 
 // The package's fully qualified name.
--- a/test/src/tests/generated/paginggroup/models.go
+++ b/test/src/tests/generated/paginggroup/models.go
@@ -174,7 +174,7 @@ func (page *OdataProductResultPage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page OdataProductResultPage) NotDone() bool {
-	return page.opr.OdataNextLink != nil && len(*page.opr.OdataNextLink) != 0
+	return (page.opr.OdataNextLink != nil && len(*page.opr.OdataNextLink) != 0)  || !page.opr.IsEmpty()
 }
 
 // Response returns the raw server response from the last page request.
@@ -393,7 +393,7 @@ func (page *ProductResultPage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page ProductResultPage) NotDone() bool {
-	return page.pr.NextLink != nil && len(*page.pr.NextLink) == 0
+	return (page.pr.NextLink != nil && len(*page.pr.NextLink) == 0) || !page.pr.IsEmpty()
 }
 
 // Response returns the raw server response from the last page request.
@@ -537,7 +537,7 @@ func (page *ProductResultValuePage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page ProductResultValuePage) NotDone() bool {
-	return page.prv.NextLink != nil && len(*page.prv.NextLink) != 0
+	return (page.prv.NextLink != nil && len(*page.prv.NextLink) != 0) || !page.prv.IsEmpty()
 }
 
 // Response returns the raw server response from the last page request.
--- a/test/src/tests/generated/paginggroup/models.go
+++ b/test/src/tests/generated/paginggroup/models.go
@@ -8,11 +8,12 @@ package paginggroup
 
 import (
 	"context"
+	"net/http"
+
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/Azure/go-autorest/tracing"
-	"net/http"
 )
 
 // The package's fully qualified name.
@@ -174,7 +175,7 @@ func (page *OdataProductResultPage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page OdataProductResultPage) NotDone() bool {
-	return !page.opr.IsEmpty()
+	return page.opr.OdataNextLink != nil || !page.opr.IsEmpty()
 }
 
 // Response returns the raw server response from the last page request.
@@ -393,7 +394,7 @@ func (page *ProductResultPage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page ProductResultPage) NotDone() bool {
-	return !page.pr.IsEmpty()
+	return page.pr.NextLink != nil || !page.pr.IsEmpty()
 }
 
 // Response returns the raw server response from the last page request.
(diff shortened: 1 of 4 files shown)


ArcturusZhang:
Hi @jhendrixMSFT could you please have a review?

test/src/tests/generated/paginggroup/models.go
```diff
@@ -174,7 +174,7 @@ func (page *OdataProductResultPage) Next() error {
 
 // NotDone returns true if the page enumeration should be started or is not yet complete.
 func (page OdataProductResultPage) NotDone() bool {
-	return !page.opr.IsEmpty()
+	return page.opr.OdataNextLink != nil || !page.opr.IsEmpty()
```
jhendrixMSFT, line 177: What about the case where the next link is the empty string?  This happens for some services and is how we check if we need to advance in track 2, see [here](https://github.com/Azure/autorest.go/blob/track2/test/autorest/generated/paginggroup/pagers.go#L51).  I also think you can remove the `IsEmpty()` check unless you can find a case where this is actually necessary.

ArcturusZhang: To double confirm, does this function `NotDone` suppose to return `true` when there is not more pages?

BTW, I have changed this according to the track 2 reference.

ArcturusZhang: And I noticed that the track 2 generator is purely written in typescript, and frankly I think it is harder to read without the handy templates...

ArcturusZhang:
Hi @jhendrixMSFT is it something wrong in my changes or something wrong in the pipeline? I did not quite get the point of the error in the pipeline.

test/src/tests/generated/paginggroup/models.go
jhendrixMSFT: Hmm it looks like this has caused some of the tests in the `paginggroup` to fail (search for "FAIL: paging_test.go" in the raw log).  Can you please investigate?

ArcturusZhang: I added the `IsEmpty` check back... hope this helps.

ArcturusZhang:
Curious that after I ran `npm run build` and `npm run regenerate`, the content in `generated` directory still show there is diff. I had to make another commit to manually change them back. Something from `1.034e-20` to `1.034E-20`

jhendrixMSFT:
Have you verified this fixes the original issue?

ArcturusZhang:
> Have you verified this fixes the original issue?

A bit awkward that we only identify this could happen on the customer's log, and we actually cannot reproduce this. The customer says this only happens on occasion. 
Do you think we should add some more test cases in the testserver to cover this case?

wsupernor:
@ArcturusZhang - 

My company (ParkMyCloud) has been experiencing this issue as well, and have had to patch the SDK ourselves to deal with empty page handling.  We hit the problem again with a customer a couple days ago, much to our dismay. Short version: we figured out how to reproduce the problem on demand, and we have created a script that you can use to try it out. Script is at https://github.com/parkmycloud/useful_tools/blob/master/Azure/azure-cli/other_scripts/azure_create_resources.sh 

TL;DR version:
We were able to observe the list of instances just before and just after the problem occurred, and noted that all of the  missing instances were _all_ alphabetically after a certain point in the returned list of instances, where alphabetically means the lexical sequence of the full resource id, like this:
`/subscriptions/11111--big-subscription-id-111111/resourceGroups/DefaultResourceGroup-CUS/providers/Microsoft.Compute/virtualMachines/ubuntu18`

We are calling the SDK method `ListComplete` with the filter:
`(resourceType eq 'Microsoft.Compute/virtualMachines' or resourceType eq 'Microsoft.Compute/virtualMachineScaleSets')
`

We believe this is what is happening:
- The public Azure API for List (https://docs.microsoft.com/en-us/rest/api/resources/resources/list) is probably calling an underlying/backend Azure API that returns an _unfiltered and paginated_ response containing every single resource in the subscription, and the response is alphabetical by resource ID.  
- The code for the public List API is then running a streaming filter against a page of the resource ID's using the filter shown above.  
- If the filter results in _all_ of the resource IDs being stripped from the page, the public List API forwards an empty page to the caller, _except_ for the `nextLink` field.
- The GoLang SDK is seeing the empty page and treating it as end-of-data, _ignoring_ the nextLink. (Which hopefully this PR will fix..)

The Python SDK behind the azure CLI appears to handle this correctly, as does the Java SDK used by a partner company of ours.

We reproduced the problem by creating a LOT of non-virtual machine resources (snapshots) that were in the same resource group. We do not know the precise number that triggers the issue. 2600 did not do it, 3100 did.  The resource identifier for snapshots is _less than_ that of virtual machines (looking at the next column in the identifier after Microsoft.Compute):
`/subscriptions/11111--big-subscription-id-111111/resourceGroups/TestRG /providers/Microsoft.Compute/`**snapshots**`/ASnapshot-10999`
vs.
`/subscriptions/11111--big-subscription-id-111111/resourceGroups/TestRG/providers/Microsoft.Compute/`**virtualMachines**`/ubuntu18`

Basically:
- Download the script from https://github.com/parkmycloud/useful_tools/blob/master/Azure/azure-cli/other_scripts/azure_create_resources.sh
- Edit the variables at the top of the script for your subscription ID and anything else you want to tweak. Use a resource group with a name like "TestRG" (this does not matter so much for the test, except insofar as all of the other resource groups in the subscription with names greater than TestRG are about to disappear from GoLang SDK visibility...)
- Run the script.
- Select create a resource group 
- Select create VMs
- Check the GoLangSDK ListComplete API - it should return all of the VMs correctly.
- In the script, select create snapshots. 
- Check the GoLangSDK ListComplete API - it should return none of the VMs VM from a resource group with a name greater than RGTest
- Apply the SDK fix from this PR
- Check the GoLangSDK ListComplete API - hopefully the VMs magically reappear

Let us know of any thoughts/feedback.

wsupernor:
@ArcturusZhang - An added note...

It looks like the code in this PR is returning empty pages to the customer code.  This is the same thing we did in our code, and we ended up having to _also_ create a new SDK `IsEmpty()` function to help our code determine if it needed to skip to the next page.  

Our new function looks like this (we applied the change to the downloadable SDK, not to the _autorest_, so I am sure your solution would have to look a bit different...):
```
func (iter ListResultIterator) IsEmpty() bool {
	return iter.page.lr.IsEmpty()
}
```

Otherwise we battled with trying to use reflection when trying to see if the returned page was empty before trying to dereference the contents.  If you are not doing to add a public `ListResultIterator` `IsEmpty()` method, could you please give a hint in the function comments as to how customer/client code can safely determine if the page is empty?

Thanks much!

jhendrixMSFT:
I think that for iterators, it should handle the empty page case internally and just move to the next non-empty page.  That way consumers don't have to worry about checking for a `nil` item (plus it just seems weird that an iterator would return a `nil` item).

wsupernor:
Yes, we thought of that approach as well...our concern was that it could result in a very rapid sequence of API calls that could invoke rate limiting/throttling on the Azure side, invisible to the client app.  I have not looked at the SDK in detail recently to see if there is increased visibility to rate limiting responses, so it may be that you have added code to handle that better...  With the proliferation of 3rd-party products monitoring large enterprise cloud performance, we have definitely seen an increase in the number of times we run into rate limiting issues.  Ref: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling

(A perfect-world-wish-list item for the SDK would be to propagate the `x-ms-ratelimit*` header items from the HTTP response...maybe in the context or something...?)

ArcturusZhang:
> I think that for iterators, it should handle the empty page case internally and just move to the next non-empty page. 

Agreed...

jhendrixMSFT:
Good point about throttling.  The retry logic in `go-autorest` (see `DoRetryForStatusCodes()`) will detect throttling and sleep for the value specified in the `Retry-After` header before retrying the request.  Note that in earlier versions the code would *infinitely* retry on a 429; this was fixed in `autorest/v0.10.0` (v14.0.0 if you're not using modules).  If you want to opt out of retries due to throttling you can remove the `http.StatusTooManyRequests` from the slice of status codes for retry [here](https://github.com/Azure/go-autorest/blob/master/autorest/client.go#L47).

ArcturusZhang:
Hi @wsupernor I could reproduce this now, thanks! And I will continue working on this to provide a more elegant fix

ArcturusZhang:
Hi @jhendrixMSFT and @wsupernor I am closing this PR in favor of PR #433

Issue #402, opened 2020-05-21
Improve resume poller check 


catalinaperalta:
Currently can resume from a poller of the same type but which was used with a different operation.

catalinaperalta:
Fixed in #403

PR #403, opened 2020-05-21
Update Resume method and LRO retry tests

catalinaperalta:
This PR updates the resume method to make sure pollers can only be resumed from tokens of pollers for the same operation on the same client. Also updating LRO retry tests to use the new design.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/402

PR #404, opened 2020-05-22
Adding x-ms-long-running-operations-options compatibility

catalinaperalta:
This PR adds functionality for x-ms-long-running-operations-options that allows setting the final get URL for LRO operation to those specified in the final-state-via field.

### DIFF
--- a/test/autorest/generated/lrogroup/pollers.go
+++ b/test/autorest/generated/lrogroup/pollers.go
@@ -107,6 +107,12 @@ func (p *productPoller) Poll(ctx context.Context) (*http.Response, error) {
 }
 
 func (p *productPoller) FinalResponse(ctx context.Context) (*ProductResponse, error) {
+	// checking if there was a FinalStateVia configuration to re-route the final GET
+	// request to the value specified in the FinalStateVia property on the poller
+	err := p.pt.setFinalState()
+	if err != nil {
+		return nil, err
+	}
 	if p.pt.finalGetURL() == "" {
 		// we can end up in this situation if the async operation returns a 200
 		// with no polling URLs.  in that case return the response which should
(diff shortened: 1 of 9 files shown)


test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -87,6 +87,12 @@ type pollingTracker interface {
 
 	// converts an *azcore.Response to an error
 	handleError(resp *azcore.Response) error
+
+	// returns the desired configuration for the final GET URI that overrides the value set in the finalGetURL
+	finalStateVia() string
```
jhendrixMSFT, line 92: Looks like this method is referenced anywhere, can it be removed?

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -107,6 +107,12 @@ func (p *productPoller) Poll(ctx context.Context) (*http.Response, error) {
 }
 
 func (p *productPoller) FinalResponse(ctx context.Context) (*ProductResponse, error) {
+	// checking if there was a FinalStateVia configuration to re-route the final GET
+	// request to the value specified in the FinalStateVia property on the poller
+	err := p.pt.setFinalState()
```
jhendrixMSFT, line 112: I think you can move the call to `setFinalState()` to `createPollingTracker()` and `resumePollingTracker()` so you don't have to call it from the generated code.  It would also let you remove the method from the interface.

test/autorest/generated/lrogroup/pollers_helper.go
catalinaperalta: Yes it can, will remove it.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: So I thought about that when developing this. However, I think that could lead to some problems, since the final polling url could technically be modified from a subsequent polling request. Therefore, I thought it was better to have it in the FinalResponse to override those changes and consolidate this functionality in one place. What do you think?

jhendrixMSFT: Good point, let's leave it as-is.

test/autorest/generated/lrogroup/pollers_helper.go
```diff
@@ -119,6 +125,13 @@ type pollingTrackerBase struct {
 	// the URL to GET for the final result
 	FinalGetURI string `json:"resultURI"`
 
+	// stores the name of the header that the final get should be performed on,
+	//can be empty which will go to default behavior
```
jhendrixMSFT, line 129: Nit, missing space after `//`

catalinaperalta: Done

test/autorest/generated/lrogroup/pollers_helper.go
catalinaperalta: Done

PR #405, opened 2020-05-22
Enabling Lro tests again

catalinaperalta:
Enabling Lrosads tests and fixing by ref passing in pollers_helper.go which was missing.


catalinaperalta:
> What prompted changing the receiver (not that I'm against it), did it fix some of the tests?

Yes, since we were not passing if by reference, certain field were not updated when they were returned.

PR #406, opened 2020-05-26
Generate go.mod file

jhendrixMSFT:
Create a go.mod file if the --module switch is specified.  The module
name will be the value provided to the switch.

### DIFF
--- a/src/generator/gomod.ts
+++ b/src/generator/gomod.ts
@@ -1,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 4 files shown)


src/generator/gomod.ts
```diff
@@ -0,0 +1,21 @@
...
+import { CodeModel } from '@azure-tools/codemodel';
+
+// Creates the content in go.mod if the --module switch was specified
+export async function generateGoModFile(session: Session<CodeModel>): Promise<string> {
+  const modName = await session.getValue('module', 'none');
+  if (modName === 'none') {
+    return '';
+  }
+  let text = `module ${modName}\n\n`;
+  text += 'go 1.13\n\n';
```
catalinaperalta, line 16: At some future point we should also think about what to do when we want to update go versions.

PR #407, opened 2020-05-26
Add support for x-ms-error-response

jhendrixMSFT:
Update error handler to return different error types depending on the
HTTP status code.
Added support for errors that are discriminated types.
Added errorsgroup from the test server.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/377

PR #409, opened 2020-05-27
Track2 updatetestserver

catalinaperalta:


test/autorest/generated/paginggroup/paging.go
```diff
@@ -46,6 +46,10 @@ type PagingOperations interface {
 	GetSinglePages() (ProductResultPager, error)
 	// GetSinglePagesFailure - A paging operation that receives a 400 on the first call
 	GetSinglePagesFailure() (ProductResultPager, error)
+	// GetWithQueryParams - A paging operation that includes a next operation. It has a different query parameter from it's next operation nextOperationWithQueryParams. Returns a ProductResult
+	GetWithQueryParams(requiredQueryParameter int32) (ProductResultPager, error)
+	// NextOperationWithQueryParams - Next operation for getWithQueryParams. Pass in next=True to pass test. Returns a ProductResult
+	NextOperationWithQueryParams(ctx context.Context) (*ProductResultResponse, error)
```
jhendrixMSFT, line 52: This should be omitted from the interface as we call this internally.  There should already be code to do this, curious why this case was missed.

catalinaperalta: NextOperationWithQueryParams is declared as its own operation in the swagger. Is there some field or condition that should be checked to omit this?

jhendrixMSFT: In `generateOperation()` the first few lines should omit this (it does for other Next* methods).

catalinaperalta: Ok, will see why this is being skipped then

src/generator/operations.ts
```diff
@@ -245,7 +245,7 @@ function formatHeaderResponseValue(propName: string, header: string, schema: Sch
 }
 
 function generateOperation(clientName: string, op: Operation, imports: ImportManager): string {
-  if (isPageableOperation(op) && op.language.go!.paging.member === op.language.go!.name) {
+  if (isPageableOperation(op) && op.language.go!.paging.member === op.language.go!.name || op.language.go!.name.startsWith('Next')) {
```
jhendrixMSFT, line 248: I'm wondering how safe/reliable this name check is.  Did you figure out why the existing check doesn't work?  It works for the other methods.

catalinaperalta: Yes I agree, I think the true fix might end up being in autorest. The problem is that the way nextOperationWithQueryParams is declared varies from the other member operations. However, I believe next link being null is purposeful for this operation: 

```
"x-ms-pageable": { "nextLinkName": "odata.nextLink", "itemName": "values", "operationName": "Paging_nextFragment" },
        "operationId": "Paging_nextFragment",
```

vs

```
 "x-ms-pageable": { "nextLinkName": null, "itemName": "values" },
        "operationId": "Paging_nextOperationWithQueryParams",
```
Simply adding operationName isn't enough for it to be marked as a member

jhendrixMSFT: I see, I was looking at the wrong thing.

The existing heuristic I introduced is half-baked :p.  I think a better solution is during transformation find all the operation IDs that also appear in `x-ms-pageable.operationName` and flag them for exclusion.  What do you think?

jhendrixMSFT: What change is needed in the swagger?

catalinaperalta: Adding `x-ms-pageable.operationName`

jhendrixMSFT: It's already there, see operation `Paging_getWithQueryParams`

catalinaperalta: Ah ok, I think I hadn't fully grasped where you wanted to get the operationName from. Got it now

PR #410, opened 2020-05-29
Adding LRO corner case for PUT requests

catalinaperalta:
This PR includes:
- Adding a check for pollers that were created from a PUT request so that they final response can be returned immediately if the response included the final payload. 
- minor code fixes
- adding internal methods to get the poller method verb
- enabling/fixing missing tests

### DIFF
--- a/test/autorest/generated/lrogroup/pollers.go
+++ b/test/autorest/generated/lrogroup/pollers.go
@@ -24,63 +24,63 @@ type HTTPPoller interface {
 }
 
 type httpPoller struct {
-  resp *HTTPResponse
+	resp *HTTPResponse
 	// the client for making the request
 	pipeline azcore.Pipeline
-  pt pollingTracker
+	pt       pollingTracker
 }
 
 // returns the method verb used in the initial request for the operation
 func (p *httpPoller) pollerMethodVerb() string {
-  return p.pt.pollerMethodVerb()
+	return p.pt.pollerMethodVerb()
 }
 
 // Done returns true if there was an error or polling has reached a terminal state
 func (p *httpPoller) Done() bool {
-  return p.pt.hasTerminated()
+	return p.pt.hasTerminated()
 }
 
 // Poll will send poll the service endpoint and return an http.Response or error received from the service
 func (p *httpPoller) Poll(ctx context.Context) (*http.Response, error) {
-  if lroPollDone(ctx, p.pipeline, p.pt) {
-    return p.pt.latestResponse().Response, p.pt.pollingError()
-  }
-  return nil, p.pt.pollingError()
+	if lroPollDone(ctx, p.pipeline, p.pt) {
+		return p.pt.latestResponse().Response, p.pt.pollingError()
+	}
+	return nil, p.pt.pollingError()
 }
 
 func (p *httpPoller) FinalResponse() *http.Response {
-      return p.pt.latestResponse().Response;
-    }
+	return p.pt.latestResponse().Response
+}
 
 // ResumeToken generates the string token that can be used with the ResumeHTTPPoller method
 // on the client to create a new poller from the data held in the current poller type
 func (p *httpPoller) ResumeToken() (string, error) {
-  if p.pt.hasTerminated() {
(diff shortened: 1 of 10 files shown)


test/autorest/generated/lrogroup/pollers.go
```diff
@@ -24,10 +24,15 @@ type HTTPPoller interface {
...
+	resp *HTTPResponse
 	// the client for making the request
 	pipeline azcore.Pipeline
-	// polling tracker
-	pt pollingTracker
+	pt       pollingTracker
+}
+
+// returns the method verb used in the initial request for the operation
+func (p *httpPoller) pollerMethodVerb() string {
```
jhendrixMSFT, line 34: Could we omit this method and just access `p.pt.pollerMethodVerb()` directly?

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -107,6 +117,9 @@ func (p *productArrayPoller) Poll(ctx context.Context) (*http.Response, error) {
 }
 
 func (p *productArrayPoller) FinalResponse(ctx context.Context) (*ProductArrayResponse, error) {
+	if p.Done() && p.pollerMethodVerb() == "PUT" && p.resp != nil && p.resp.ProductArray != nil {
```
jhendrixMSFT, line 120: Can use `http.MethodPut` instead of `"PUT"`.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -212,6 +231,10 @@ func (p *productPoller) Poll(ctx context.Context) (*http.Response, error) {
 }
 
 func (p *productPoller) FinalResponse(ctx context.Context) (*ProductResponse, error) {
+	product := Product{}
```
jhendrixMSFT, line 234: If `product` isn't used elsewhere you can omit this definition and just use `Product{}` on the next line.

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: Yes, will remove

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: The compiler doesnt like that, there's a problem with the braces
if ... Product{} {
 ...
}

jhendrixMSFT: Ah, looks like parens are needed around the expression `(*p.resp.Product != Product{})`

catalinaperalta: Ah ok, will do then

test/autorest/generated/lrogroup/pollers.go

test/autorest/generated/lrogroup/pollers.go

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -161,6 +167,7 @@ func (p *productArrayPoller) pollUntilDone(ctx context.Context, frequency time.D
 			return nil, err
 		}
 		if p.Done() {
+			p.resp, _ = p.handleResponse(p.pt.latestResponse())
```
jhendrixMSFT, line 170: Presumably it's ok to ignore the error here because it will get returned from the call to `FinalResponse()` yes?

test/autorest/generated/lrogroup/pollers.go

test/autorest/generated/lrogroup/pollers.go
catalinaperalta: That was the original intent, but taking a closer look at the unmarshalling section of handleResponse, I think this error could be hidden away. I'll add the immediate error check and return here as well.

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -24,10 +24,10 @@ type HTTPPoller interface {
 }
 
 type httpPoller struct {
+	resp *HTTPResponse
```
jhendrixMSFT, line 27: looks like this is no longer being used

Issue #250, opened 2020-02-11
Check how to unmarshal/marshal base64URL encoded values


catalinaperalta:

catalinaperalta:
Necessary to get some stringgroup operations working properly.

jhendrixMSFT:
Swagger doesn't specify if base-64 encoded values omit padding, we will try both cases.

jhendrixMSFT:
Fixed

Issue #243, opened 2020-02-10
String group acceptance test problems


catalinaperalta:
- GetBase64URLEncoded: returns an unmarshalling error. We need to add an unmarshal/marshal method on azcore.Response and azcore.Request. Also need to invoke these methods in codegen when working with base64 data. 
- PutBase64URLEncoded: same as GetBase64URLEncoded
- TestStringPutNull: pending query about whether PutNull should send null in the request body or send an undefined request body.

jhendrixMSFT:
Fixed

PR #411, opened 2020-06-01
Fix base-64 URL encoded strings

jhendrixMSFT:
Add support for base-64 standard and URL-encoded strings.
Updated azcore and test server dependencies.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/243

jhendrixMSFT:
Also fixes https://github.com/Azure/autorest.go/issues/250

PR #414, opened 2020-06-03
Fix dictionary type ignore discriminator

ArcturusZhang:
`testserver` does not have a test case for this scenario, therefore I leave the testserver version unchanged.
Additionally, upgrading testserver version will introduce some cases that the generated code cannot properly compile. Will investigate those cases in the future PRs.

Since our code will never call something like this: `make(map[string]*SomeStruct)`, I think this fix is valid. I have tested the generation of the problematic RP, it turns out fine.


fearthecowboy:
## success (version: 2.1.147)

PR #415, opened 2020-06-04
Adding LRO fix for missing response model fields and string value check

catalinaperalta:
This PR includes:
- fixing missing LRO fields in response models by adding a check for whether the operation is an LRO and adding the response to the global response schemas list
- Adding a string value check for PUT checks
- Adding partial poller functionality for paging LROs WIP


test/autorest/generated/paginggroup/models.go
```diff
@@ -87,6 +91,11 @@ type ProductResult struct {
 
 // ProductResultResponse is the response envelope for operations that return a ProductResult type.
 type ProductResultResponse struct {
+	// PollUntilDone will poll the service endpoint until a terminal state is reached or an error is received
+	PollUntilDone func(ctx context.Context, frequency time.Duration) (*ProductResultResponse, error)
```
jhendrixMSFT, line 95: So now this gets shared with synchronous APIs.  I have mixed feelings about this.  Probably worth discussing in scrum once Jeff is back.

src/transform/transform.ts
```diff
@@ -465,6 +465,18 @@ function createResponseType(codeModel: CodeModel, group: OperationGroup, op: Ope
         prop.language.go!.fromHeader = item.value.header;
         (<Array<Property>>object.language.go!.properties).push(prop);
       }
+      const firstResp = <SchemaResponse>op.responses![0];
+      if (isLROOperation(op)) {
```
jhendrixMSFT, line 469: Would this actually be hit?  Seems like line 428 would prevent this case from ever getting hit.

catalinaperalta: I think you're right, I thought there would be a case where we could reach this without having added the LRO related fields. But on second look I dont think so. I'll remove this change.

src/transform/transform.ts
```diff
@@ -523,13 +523,44 @@ function createResponseType(codeModel: CodeModel, group: OperationGroup, op: Ope
...
+      }
     }
   }
   // create pageable type info
   if (isPageableOperation(op)) {
     if (codeModel.language.go!.pageableTypes === undefined) {
       codeModel.language.go!.pageableTypes = new Array<PagerInfo>();
     }
+    // Add a poller
+    if (isLROOperation(op)) {
```
jhendrixMSFT, line 543: Can we move this to the LRO+paging PR so all the changes are together?

catalinaperalta: I added this since the paging LRO is the only one that checks this functionality and in order to build we need to have the poller definition. I could add a skip for adding poller fields if it's also a paging LRO, then I could remove this section. Should I skip them then?

jhendrixMSFT: Ah got it, no let's leave it.

jhendrixMSFT: Actually, looking more at this, we already have a `isLROOperation()` check on line 583 which should create the poller after creating the pager, so this seems redundant.  It was working before.

catalinaperalta: So the problem is that the pagers implementation has a return inside of it, so the subsequent LRO work can sometimes we reached and other times not depending on whether or not the return is reached in the pager code.

catalinaperalta: I had also thought of moving LRO creation above pager creation, but LRO creation now has a return in it as well. So we would have the same problem for creating a pager in the LRO.

catalinaperalta: Actually the returns should be breaks.

jhendrixMSFT: Looks like the returns are just to avoid adding duplicate pagers/pollers.  I think they could be replaced with some boolean flags?

jhendrixMSFT: Yeah breaks + a flag I think.

catalinaperalta: Yes will do 👍

catalinaperalta: Done

Issue #412, opened 2020-06-01
Check why some media tags are undefined


catalinaperalta:
Check lrogroup, paginggroup and azurespecialsgroup for these undefined tags.

jhendrixMSFT:
The affected types aren't used as parameter or return values, so when we recursively add the marhsalling format they're skipped.  Should be able to apply the doc-level produces/consumes to determine the marshalling format.

jhendrixMSFT:
Dependent on https://github.com/Azure/perks/issues/108

jhendrixMSFT:
We've implemented a work-around, the underlying issue is in perks.

PR #416, opened 2020-06-05
Workaround for undefined tags

jhendrixMSFT:
Set the marshalling format to JSON if it's missing; this is to work
around a bug in AutoRest.
Removed some redundant code.


jhendrixMSFT:
Workaround for https://github.com/Azure/autorest.go/issues/412

PR #417, opened 2020-06-08
Add switch to disable exporting of client types

jhendrixMSFT:
Added --export-client, which when set to false, will not export client
types.  Updated azblob to use this switch.
Updated minimum version of azcore in go.mod template.

### DIFF
--- a/rushScripts/regeneration.js
+++ b/rushScripts/regeneration.js
@@ -48,7 +48,7 @@ for (namespace in goMappings) {
 } 
 
 const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
-generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage"');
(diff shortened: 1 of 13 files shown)


rushScripts/regeneration.js
```diff
@@ -48,7 +48,7 @@ for (namespace in goMappings) {
 } 
 
 const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
-generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage"');
+generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage" --export-client="false"');
```
catalinaperalta, line 51: Why dont we want to export clients for storage?

jhendrixMSFT: This is just for testing purposes.  For the actual shipping bits, the storage team wants to export hand-written clients created in the convenience layer that consume the internally generated clients (this is what they did in track 1.8).

catalinaperalta: Got it. Thanks for the info!

PR #419, opened 2020-06-08
Add support for Unix time

jhendrixMSFT:
Generate internal type timeUnix for marshalling and unmarshalling time
in Unix format.

Fixes https://github.com/Azure/autorest.go/issues/339


jhendrixMSFT:
See also https://github.com/Azure/autorest.go/issues/339

PR #421, opened 2020-06-10
Add support for multiple response types

jhendrixMSFT:
Operations that return different types depending on HTTP status code now
return interface{}.
Generate a method comment that includes the possible return types.
LROs are excluded from this conversion as it's not applicable.
Fixed response envelope generation for multi-response methods.
Remove HTTP redirect status codes for cases where the Go HTTP client
implicitly follows them and added missing tests.
Fixed codegen for checking HTTP status codes of single-response APIs.
Updated test server version which fixed a few urlgroup tests.

### DIFF
--- a/test/autorest/generated/httpinfrastructuregroup/httpredirects.go
+++ b/test/autorest/generated/httpinfrastructuregroup/httpredirects.go
@@ -14,37 +14,38 @@ import (
 // HTTPRedirectsOperations contains the methods for the HTTPRedirects group.
 type HTTPRedirectsOperations interface {
 	// Delete307 - Delete redirected with 307, resulting in a 200 after redirect
-	Delete307(ctx context.Context) (*HTTPRedirectsDelete307Response, error)
+	Delete307(ctx context.Context) (*http.Response, error)
 	// Get300 - Return 300 status code and redirect to /http/success/200
-	Get300(ctx context.Context) (*HTTPRedirectsGet300Response, error)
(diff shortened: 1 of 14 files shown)


test/autorest/generated/httpinfrastructuregroup/httpredirects.go
```diff
@@ -130,15 +128,23 @@ func (client *httpRedirectsOperations) get300CreateRequest() (*azcore.Request, e
 }
 
 // get300HandleResponse handles the Get300 response.
-func (client *httpRedirectsOperations) get300HandleResponse(resp *azcore.Response) (*HTTPRedirectsGet300Response, error) {
-	if !resp.HasStatusCode(http.StatusOK) {
+func (client *httpRedirectsOperations) get300HandleResponse(resp *azcore.Response) (interface{}, error) {
+	switch resp.StatusCode {
+	case http.StatusOK:
+		result := HTTPRedirectsGet300Response{RawResponse: resp.Response}
```
catalinaperalta, line 134: Just for clarity, we want this to be a HTTPRedirectsGet300Response? Or should it be a http.Response?

jhendrixMSFT: This operation returns a model for HTTP 300, and since 300 aren't automatically redirected we return a model here.

catalinaperalta: I see. OK I was checking because I saw that we're avoiding redirects. But looks good to me if the swagger is returning the model.

Issue #242, opened 2020-02-10
AdditionalProperties field missing


catalinaperalta:
AdditionalProperties needs to include two additionalProperties fields with different dictionary values.

jhendrixMSFT:
Fixed

PR #423, opened 2020-06-11
Add support for additional properties

jhendrixMSFT:
Add an 'AdditionalProperties' map on applicable types along with a
custom marshaller and unmarshaller.
If a type supports additional properties and also has a field named
'AdditionalProperties' rename the field by appending a '1'.
Note that the current implementation doesn't work for discriminated
types that contain additional properties, or types that use XML.  Will
be fixed in later PRs.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/242

src/generator/models.ts
```diff
@@ -532,3 +548,94 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
...
+    }
+    marshaller += emitMarshaller(prop);
+  }
+  for (const prop of values(structDef.Properties)) {
+    if (prop.language.go!.isAdditionalProperties) {
+      continue;
+    }
+    marshaller += emitMarshaller(prop);
+  }
+  marshaller += `\tif ${receiver}.AdditionalProperties != nil {\n`;
```
catalinaperalta, line 582: Just checking, will additional properties always be in a field called AdditionalProperties?

jhendrixMSFT: Yep!  This is what we did for track 1 and is what the other languages do as well.

PR #425, opened 2020-06-16
Adding response payload to error messages when exceptions are not defined

catalinaperalta:
This PR adds more information for errors when the swagger doesnt define exceptions.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -753,39 +753,35 @@ function createProtocolErrHandler(client: string, op: Operation, imports: Import
   const name = info.protocolNaming.errorMethod;
   let text = `${comment(name, '// ')} handles the ${info.name} error response.\n`;
   text += `func (client *${client}) ${name}(resp *azcore.Response) error {\n`;
-  // if the response doesn't define any error types return a generic error
-  if (!op.exceptions) {
+
+  // define a generic error for when there are no exceptions or no error schema
+  const generateGenericError = function () {
     imports.add('errors');
     imports.add('io/ioutil');
     imports.add('fmt');
-    text += `body, err := ioutil.ReadAll(resp.Body)
-      if err != nil {
-        return fmt.Errorf("%s; failed to read response body: %w", resp.Status, err)
-      }
-      if len(body) == 0 {
-        return errors.New(resp.Status)
-      }
-      return errors.New(string(body))
+    return `body, err := ioutil.ReadAll(resp.Body)
+    if err != nil {
+      return fmt.Errorf("%s; failed to read response body: %w", resp.Status, err)
     }
-    
+    if len(body) == 0 {
+      return errors.New(resp.Status)
+    }
+    return errors.New(string(body))
     `;
+    
+  }
+
+  // if the response doesn't define any error types return a generic error
+  if (!op.exceptions) {
+    text += generateGenericError();
+    text += '}\n\n';
     return text;
   }
+
   const generateUnmarshaller = function (exception: Response, prefix: string) {
     let unmarshaller = '';
     if (exception.language.go!.genericError) {
-      imports.add('errors');
-      imports.add('io/ioutil');
-      imports.add('fmt');
-      unmarshaller += `${prefix}body, err := ioutil.ReadAll(resp.Body)
-      if err != nil {
-        return fmt.Errorf("%s; failed to read response body: %w", resp.Status, err)
-      }
-      if len(body) == 0 {
-        return errors.New(resp.Status)
-      }
-      return errors.New(string(body))
-      `;
+      unmarshaller += `${prefix}${generateGenericError()}`;
       return unmarshaller;
     }
     const schemaError = (<SchemaResponse>exception).schema;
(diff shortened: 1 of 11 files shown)


test/autorest/generated/azurespecialsgroup/xmsclientrequestid.go
```diff
@@ -63,7 +64,15 @@ func (client *xmsClientRequestIdoperations) getHandleResponse(resp *azcore.Respo
 
 // getHandleError handles the Get error response.
 func (client *xmsClientRequestIdoperations) getHandleError(resp *azcore.Response) error {
-	return errors.New(resp.Status)
+	msg := resp.Status
```
jhendrixMSFT, line 67: I think we should omit this if there's a response body.

catalinaperalta: Yes, I agree. Will update

test/autorest/generated/azurespecialsgroup/xmsclientrequestid.go
```diff
@@ -63,7 +64,15 @@ func (client *xmsClientRequestIdoperations) getHandleResponse(resp *azcore.Respo
...
+	if resp.Body != nil {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
```
jhendrixMSFT, line 71: Probably fall back to returning `resp.Status` in this case as it's more descriptive.  We could include this error in the message though.  Maybe something like
`return fmt.Errorf("%s; failed to read response body: %w", resp.Status, err)`

test/autorest/generated/azurespecialsgroup/xmsclientrequestid.go
```diff
@@ -63,6 +65,13 @@ func (client *xmsClientRequestIdoperations) getHandleResponse(resp *azcore.Respo
 
 // getHandleError handles the Get error response.
 func (client *xmsClientRequestIdoperations) getHandleError(resp *azcore.Response) error {
+	if resp.Body != nil {
```
jhendrixMSFT, line 68: I would invert this to keep your happy-line happy.
```go
if resp.Body == nil {
  return errors.New(resp.Status)
}
...
```

jhendrixMSFT: And actually, checking for `nil` is incorrect as the `Body` field will always be non-nil according to the standard lib.  Now we could change this in the body downloader policy to set `Body` to nil if it was empty however it changes the contract given by the stdlib so I'm not in favor of that.
Maybe we should have a method on `azcore.Response` to create a generic error from the response?

catalinaperalta: Hmmm...well I agree that we shouldn't do something that changes the contract for the stdlib. Maybe it could be worth it to have an ErrorFromResponse or PayloadError method on `azcore.Response`. Having a generic error for `azcore.Response` does sound nice. It would simplify things. However, I will say that it would add more surface area to azcore, while keeping it here would be an implementation detail of handling errors for each service.

catalinaperalta: So are you thinking that we could add something in azcore.Response that uses the payload() method? It would be convenient.

jhendrixMSFT: I was thinking about the public surface area thing too, I'm a bit torn on it, but we should probably avoid it.  I guess this means that you have to read the body and if the length is zero then return the generic error.

jhendrixMSFT: Yes uses payload, just like the unmarshallers.

catalinaperalta: Ok cool, will use the length of the body.

catalinaperalta: Done

test/autorest/generated/azurespecialsgroup/xmsclientrequestid.go
catalinaperalta: Done

test/autorest/generated/azurespecialsgroup/xmsclientrequestid.go
catalinaperalta: Done

src/generator/operations.ts
```diff
@@ -756,15 +756,36 @@ function createProtocolErrHandler(client: string, op: Operation, imports: Import
...
+    
+    `;
     return text;
   }
   const generateUnmarshaller = function (exception: Response, prefix: string) {
     let unmarshaller = '';
     if (exception.language.go!.genericError) {
       imports.add('errors');
-      unmarshaller += `${prefix}return errors.New(resp.Status)\n`;
+      imports.add('io/ioutil');
```
jhendrixMSFT, line 778: Since this has grown probably should refactor it into a single definition to avoid the duplication (can probably do the same thing as `generateUnmarshaller` above).

PR #426, opened 2020-06-16
Add SDK and module info to telemetry

jhendrixMSFT:


src/transform/namer.ts
```diff
@@ -51,6 +51,10 @@ export async function namer(session: Session<CodeModel>) {
   // copy all the .language.default data into .language.go
   cloneLanguageInfo(model);
 
+  // default namespce to the output folder
+  const outputFolder = await session.getValue<string>('output-folder');
+  model.language.go!.packageName = outputFolder.substr(outputFolder.lastIndexOf('/') + 1);
```
catalinaperalta, line 56: So the namespace will always be defined by the last directory in the output-folder path?

jhendrixMSFT: Correct, the two have to match (this is how it was, I just moved it to here).

PR #428, opened 2020-06-17
Add more test server tests

jhendrixMSFT:
Added missing urlgroup tests.
Added extensible enum codegen and tests.


test/autorest/urlgroup/queries_test.go
```diff
@@ -103,6 +273,27 @@ func TestGetBooleanTrue(t *testing.T) {
...
 }
 
+// GetIntNegativeOneMillion - Get '-1000000' integer value
+func TestGetIntNegativeOneMillion(t *testing.T) {
+	client := getQueriesClient(t)
+	result, err := client.GetIntNegativeOneMillion(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	helpers.VerifyStatusCode(t, result, http.StatusOK)
```
catalinaperalta, line 283: Wondering if this should have a deepequal check for the value?

catalinaperalta: Same for several of these tests that do GETs.

jhendrixMSFT: This one doesn't return a model, just an `*http.Resposne` (looks like they all do)..

PR #429, opened 2020-06-22
Separate poller response envelopes

catalinaperalta:
This PR creates the PollerResponse envelopes that contain poller related functionality (PollUntilDone and the initialized Poller). Also separating pollers from widget response envelopes. 

Headers are excluded for LRO operations. 
Pager related improvements and specific response envelopes will be added in the LRO+pager PR.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -296,7 +296,7 @@ function generateOperation(clientName: string, op: Operation, imports: ImportMan
   if (isLROOperation(op)) {
     // TODO remove LRO for pageable responses NYI
     if (op.extensions!['x-ms-pageable']) {
-      text += `\treturn nil, nil`;
+      text += `\treturn nil, nil\n`;
       text += '}\n\n';
       return text;
     }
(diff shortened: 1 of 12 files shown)


src/transform/transform.ts
```diff
@@ -564,43 +563,51 @@ function createResponseType(codeModel: CodeModel, group: OperationGroup, op: Ope
...
-        }
       }
     }
     // create pageable type info
     if (isPageableOperation(op)) {
       if (codeModel.language.go!.pageableTypes === undefined) {
         codeModel.language.go!.pageableTypes = new Array<PagerInfo>();
       }
-      const name = `${(<SchemaResponse>response).schema.language.go!.name}Pager`;
+      let suffix = 'Pager';
```
jhendrixMSFT, line 609: Should be `const`, or you could just do `const name = <name>+'Pager;'`.

src/generator/operations.ts
```diff
@@ -1039,14 +1043,19 @@ function generateReturnsInfo(op: Operation, forHandler: boolean): string[] {
...
-    if (!forHandler && isPageableOperation(op)) {
+    const firstResp = <SchemaResponse>op.responses![0];
+    // must check pageable first as all pageable operations are also schema responses, 
+    // but LRO operations that return a pager are an exception and need to return LRO specific
+    // responses
+    if (!forHandler && isPageableOperation(op) && !isLROOperation(op)) {
       returnType = op.language.go!.pageableType.name;
     } else if (isSchemaResponse(firstResp)) {
       returnType = '*' + firstResp.schema.language.go!.responseType.name;
+      if (isLROOperation(op) || (isLROOperation(op) && forHandler)) {
```
jhendrixMSFT, line 1054: The second half of the OR clause will never be true.

catalinaperalta: Oops, I thought I had removed this. Will do

src/transform/transform.ts
catalinaperalta: Done

src/generator/operations.ts

PR #433, opened 2020-07-06
Fix paging when list request returns empty value but non-null nextlink

ArcturusZhang:
This PR supersedes PR #401 and provides a more elegant fix for the empty-page issue.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/9510

This PR makes the following changes:
1. the `NextWithContext` function to make it keeping moving to next page until the values array is not empty or nextLink is empty (no more pages) 
1. the List call will call the `NextWithContext` function to ensure the first page is non-empty


ArcturusZhang:
By taking the approach introduced by @wsupernor in PR #401 , I could reproduce this issue and this PR fixes this issue.

ArcturusZhang:
@wsupernor here is a sample repo that demonstrate how this could resolve this issue. You can find the changes introduced by this PR here:
1. https://github.com/ArcturusZhang/azure-sdk-for-go-list-empty-page/blob/04f5770cf5ad7c730db8f968954159e7e0ff6cd7/vendor/github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources/resources.go#L899
1. https://github.com/ArcturusZhang/azure-sdk-for-go-list-empty-page/blob/04f5770cf5ad7c730db8f968954159e7e0ff6cd7/vendor/github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2019-05-01/resources/models.go#L1374

test/src/tests/generated/paginggroup/models.go
```diff
@@ -157,11 +157,16 @@ func (page *OdataProductResultPage) NextWithContext(ctx context.Context) (err er
...
-	next, err := page.fn(ctx, page.opr)
-	if err != nil {
-		return err
+	for {
+		next, err := page.fn(ctx, page.opr)
+		if err != nil {
+			return err
+		}
+		page.opr = next
+		if next.OdataNextLink == nil || (next.Values != nil && len(*next.Values) != 0) {
```
jhendrixMSFT, line 166: I think this also needs to check for an empty next link.  Might be good to generate a helper method "HasNextLink() bool" which would also be used in e.g. `odataProductResultPreparer()`.

Also can't the second condition be rewritten to use `IsEmpty()`?

jhendrixMSFT:
I'm trying to convince myself that there are no cases where this could cause an infinite loop.  Have you thought about this potential?

@catalinaperalta we should discuss something similar for track 2 codegen.

ArcturusZhang:
@jhendrixMSFT the condition of break the loop is either it has an empty next link or it has no values. The only scenario that an infinite loop can happen should be the service constantly return results with no values but non-empty next link, which I suppose should be a service side problem.

test/src/tests/generated/paginggroup/models.go
ArcturusZhang: Very reasonable. I will take a look to add the new function `hasNextLink` and I suppose this function could be un-exported.

And the second part of the condition is exactly `!IsEmpty()`

ArcturusZhang:
Hi @jhendrixMSFT I have made some changes, please have a look, thanks

fearthecowboy:
## success (version: 2.1.150)

PR #434, opened 2020-07-07
Various fixes

jhendrixMSFT:
Fixed infinite loop when walking discriminators hierarchy.
Renamed "k, v" locals to "key, val" to avoid clashing with receivers.
Added 304 to list of HTTP status codes.

### DIFF
--- a/src/transform/transform.ts
+++ b/src/transform/transform.ts
@@ -842,6 +842,11 @@ function getRootDiscriminator(obj: ObjectSchema): ObjectSchema {
       if (isObjectSchema(parent) && parent.discriminator) {
         root.language.go!.discriminatorParent = parent.language.go!.discriminatorInterface;
         root = parent;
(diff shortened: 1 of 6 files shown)


src/transform/transform.ts
```diff
@@ -842,6 +842,7 @@ function getRootDiscriminator(obj: ObjectSchema): ObjectSchema {
       if (isObjectSchema(parent) && parent.discriminator) {
         root.language.go!.discriminatorParent = parent.language.go!.discriminatorInterface;
         root = parent;
+        obj = root;
```
catalinaperalta, line 845: Is this what's fixing the infinite loop? How does it do it?

jhendrixMSFT: Yep this is the fix for that.  I've added a comment explaining how this works.

catalinaperalta: Great, thanks! :)

PR #435, opened 2020-07-07
Adding LRO retry after fix for initial response delay 

catalinaperalta:
This PR includes a fix to consider retry-after headers in the initial response from the endpoing for LROs. Also adds consideration for ctx time out.


test/autorest/generated/lrogroup/pollers.go
```diff
@@ -60,6 +60,16 @@ func (p *httpPoller) ResumeToken() (string, error) {
 }
 
 func (p *httpPoller) pollUntilDone(ctx context.Context, frequency time.Duration) (*http.Response, error) {
+	// initial check for a retry-after header existing on the initial response
+	if retryAfter := azcore.RetryAfter(p.pt.latestResponse().Response); retryAfter > 0 {
+		select {
```
jhendrixMSFT, line 65: I think this could be refactored to avoid the duplicate code.  Maybe something like this in `poller_helpers.go`
```go
func delay(ctx context.Context, delay time.Duration) error {
	select {
	case <-time.After(delay):
		return nil
	case <-ctx.Done():
		return ctx.Err()
	}
}
```

Issue #440, opened 2020-07-16
Fix url path parsing


catalinaperalta:
In create request methods we parse endpoint paths on the client url, but this parsing replaces any pre-existing path on the url. Instead combine the previously declared path to the new one like this: `client.u.Parse(path.Join(u.Path, "/foo"))`

jhendrixMSFT:
Fixed.

PR #442, opened 2020-07-16
Fixing path parsing to include previously set path on client url

catalinaperalta:
This PR fixes path parsing, so that new paths are appended to previously declared paths on the endpoint url. 

This helps ensure proper functioning for cases where the host is declared as: `{endpoint}/foo` so that the `/foo` is not lost upon request creation.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -397,7 +397,8 @@ function createProtocolRequest(client: string, op: Operation, group: OperationGr
     text += '\t\treturn nil, err\n';
     text += '\t}\n';
   } else if (includeParse) {
-    text += `\tu, err := client.u.Parse(${parseVar})\n`;
+    imports.add('path');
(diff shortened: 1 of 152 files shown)


src/generator/operations.ts
```diff
@@ -397,7 +397,8 @@ function createProtocolRequest(client: string, op: Operation, group: OperationGr
     text += '\t\treturn nil, err\n';
     text += '\t}\n';
   } else if (includeParse) {
-    text += `\tu, err := client.u.Parse(${parseVar})\n`;
+    imports.add('path');
```
jhendrixMSFT, line 400: Why only for the `includeParse` case?

catalinaperalta: The `includeParse` case is the only one where we add `client.u.Parse()`, so joining the paths will exist on all requests that want to add a path to the client url.

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/440

PR #443, opened 2020-07-17
Param host url and host check updates

catalinaperalta:
This PR updates the url on the client in parameterized host cases to also be based by value. 

The host parameter check now accepts x-ms-priority 0 or 1.


catalinaperalta:
Closing this since these changes are no longer relevant as per offline discussion.

PR #445, opened 2020-07-20
Update handling for client vars with default values

catalinaperalta:
This PR improves code to check client vars that have default values to avoid panics and assign them on the client.

### DIFF
--- a/src/generator/client.ts
+++ b/src/generator/client.ts
@@ -256,7 +256,8 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
       let pointer = '';
       if (pp.clientDefaultValue) { // TODO support more than just string params
         text += `if ${pp.language.go!.name} == nil {
-          *${pp.language.go!.name} = "${pp.clientDefaultValue}"
+          temp := "${pp.clientDefaultValue}"
+          ${pp.language.go!.name} = &temp
         }
         `;
         pointer = '*';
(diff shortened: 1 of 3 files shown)


src/generator/client.ts
```diff
@@ -273,29 +274,38 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
...
+        text += `\t\t${param.language.go!.name}: "${param.clientDefaultValue}",\n`;
+      }
+    }
+    text += `\t}\n`;
     for (const param of values(clientOnlyParams)) {
-      let pointer = '';
       if (param.clientDefaultValue) { // TODO support more than just string params
-        text += `if ${param.language.go!.name} == nil {
-          *${param.language.go!.name} = "${param.clientDefaultValue}"
+        text += `if ${param.language.go!.name} != nil {
```
jhendrixMSFT, line 292: !==

src/generator/client.ts
```diff
@@ -273,29 +274,38 @@ export async function generateClient(session: Session<CodeModel>): Promise<strin
...
       }
-      text += `\tclient.${param.language.go!.name} = ${pointer}${param.language.go!.name}\n`;
     }
     for (const param of values(paramHostInfo.clientParams)) {
-      let pointer = '';
       if (param.clientDefaultValue) { // TODO support more than just string params
-        text += `if ${param.language.go!.name} == nil {
-          *${param.language.go!.name} = "${param.clientDefaultValue}"
+        text += `if ${param.language.go!.name} != nil {
+          client.${param.language.go!.name} = *${param.clientDefaultValue}
```
jhendrixMSFT, line 303: Shouldn't this line be `client.${param.language.go!.name} = *${param.language.go!.name}`

catalinaperalta: Yes, thanks for catching that! I've updated it. (There's no test that covers that)

PR #446, opened 2020-07-21
Add marshaler for the structs with readonly properties

### LINKED ISSUE #438: Ensure that we do not marshal read-only fields into request body in any level
In the current go SDK, we do not have custom marshaler for every struct, and some of them may have read-only fields. Without a custom marshaler, the read-only fields will be marshaled into the request body.

For instance, the [`VirtualMachineProperties`](https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2020-06-01/compute/models.go#L10581) does not have a custom marshaler, despite the `VMID` field is read-only, it will still be in the request body if assigned.

Sample code of reproduce:
```go
func createVM(authorizer autorest.Authorizer) {
	client := compute.NewVirtualMachinesClient(subsID)
	client.Authorizer = authorizer
	vm := compute.VirtualMachine{
		Location: to.StringPtr("eastus"),
		VirtualMachineProperties: &compute.VirtualMachineProperties{
			VMID: to.StringPtr("TEST"),
		},
	}
	future, err := client.CreateOrUpdate(ctx, "test-rg", "test-vm", vm)
	if err != nil {
		panic(err)
	}
	if err := future.WaitForCompletionRef(ctx, client.Client); err != nil {
		panic(err)
	}
}
```
Log:
```
(2020-07-14T16:13:04.0542686+08:00) INFO: REQUEST: PUT https://management.azure.com/subscriptions/**REDACTED**/resourceGroups/test-rg/providers/
Microsoft.Compute/virtualMachines/test-vm?api-version=2019-12-01
Content-Type: application/json; charset=utf-8
User-Agent: Go/go1.13.6 (amd64-windows) go-autorest/v14.0.0 Azure-SDK-For-Go/1.0.0 compute/2019-12-01
Authorization: **REDACTED**
{"location":"eastus","properties":{"vmId":"TEST"}}
```

ArcturusZhang:
Fixes #438


ArcturusZhang:
Hi @jhendrixMSFT the CI has passed, please have a look, thanks!

ArcturusZhang:
And @jhendrixMSFT regarding the big files issue, I am wondering would this patch create more gigantic `models.go` files? Is it possible to split something into different files?

jhendrixMSFT:
It will increase the size yes.

If you want, you could split all the stuff for enums into its own file.  We did this in track 2 for just this reason.

fearthecowboy:
## success (version: 2.1.149)

PR #447, opened 2020-07-23
Simplify parameterized host functionality

catalinaperalta:
This PR simplifies the parameterized host functionality and consolidates client param grouping in the transform.

### DIFF
--- a/test/autorest/generated/custombaseurlgroup/paths.go
+++ b/test/autorest/generated/custombaseurlgroup/paths.go
@@ -44,8 +44,8 @@ func (client *pathsOperations) GetEmpty(ctx context.Context, accountName string)
 // getEmptyCreateRequest creates the GetEmpty request.
 func (client *pathsOperations) getEmptyCreateRequest(accountName string) (*azcore.Request, error) {
 	host := "http://{accountName}{host}"
+	host = strings.ReplaceAll(host, "{host}", client.host)
 	host = strings.ReplaceAll(host, "{accountName}", accountName)
-	host = strings.ReplaceAll(host, "{host}", client.Client.host)
 	urlPath := "/customuri"
 	u, err := url.Parse(host + urlPath)
 	if err != nil {
--- a/test/autorest/generated/morecustombaseurigroup/client.go
+++ b/test/autorest/generated/morecustombaseurigroup/client.go
@@ -53,13 +53,13 @@ func (c *ClientOptions) telemetryOptions() azcore.TelemetryOptions {
 
 // Client - Test Infrastructure for AutoRest
 type Client struct {
-	dnsSuffix      string
 	subscriptionID string
+	dnsSuffix      string
 	p              azcore.Pipeline
 }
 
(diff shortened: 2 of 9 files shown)


test/autorest/generated/morecustombaseurigroup/client.go
```diff
@@ -53,13 +53,13 @@ func (c *ClientOptions) telemetryOptions() azcore.TelemetryOptions {
...
 type Client struct {
-	dnsSuffix      string
 	subscriptionID string
+	dnsSuffix      string
 	p              azcore.Pipeline
 }
 
 // NewClient creates an instance of the Client type with the specified endpoint.
-func NewClient(dnsSuffix *string, subscriptionID string, options *ClientOptions) (*Client, error) {
+func NewClient(subscriptionID string, dnsSuffix *string, options *ClientOptions) (*Client, error) {
```
jhendrixMSFT, line 62: SubscriptionID is in the wrong place, it should be an operation group method param.

catalinaperalta: Fixed

test/autorest/generated/custombaseurlgroup/paths.go
```diff
@@ -44,8 +44,8 @@ func (client *pathsOperations) GetEmpty(ctx context.Context, accountName string)
 // getEmptyCreateRequest creates the GetEmpty request.
 func (client *pathsOperations) getEmptyCreateRequest(accountName string) (*azcore.Request, error) {
 	host := "http://{accountName}{host}"
+	host = strings.ReplaceAll(host, "{host}", client.host)
```
catalinaperalta, line 47: I think this is being generated incorrectly...shouldn't it be `client.Client.host`?
Also surprised that the tests pass with this...or am I missing something?

test/autorest/generated/morecustombaseurigroup/paths.go
```diff
@@ -45,12 +45,12 @@ func (client *pathsOperations) GetEmpty(ctx context.Context, vault string, secre
 // getEmptyCreateRequest creates the GetEmpty request.
 func (client *pathsOperations) getEmptyCreateRequest(vault string, secret string, keyName string, pathsGetEmptyOptions *PathsGetEmptyOptions) (*azcore.Request, error) {
 	host := "{vault}{secret}{dnsSuffix}"
+	host = strings.ReplaceAll(host, "{dnsSuffix}", client.dnsSuffix)
```
catalinaperalta, line 48: Same as above...shouldn't this be `client.Client.dnsSuffix`?

test/autorest/generated/custombaseurlgroup/paths.go
jhendrixMSFT: The `*Client` is anonymously embedded in `pathsOperations`, so we can access it directly.

catalinaperalta: Ah true, a mistake I made in the previous version. Thanks for fixing it!

PR #448, opened 2020-07-23
Separate enums from models

ArcturusZhang:
Follow up of #446


ArcturusZhang:
I wonder if we have some criteria to separate more of them from `models.go`? Reduction of the size of `models.go` may solve the issue of big files in which the intellisense stops working and will increase user experience a lot

ArcturusZhang:
Hi @jhendrixMSFT could you please have a look?

test/src/tests/generated/additionalproperties/enums.go
```diff
@@ -0,0 +1,7 @@
+package additionalproperties
```
jhendrixMSFT, line 1: You should avoid generating empty enum files.

jhendrixMSFT:
This is using `gopls` for intellisense?  Do you know if large files contributes to this problem?  I'm not against carving this up even more, however it would be nice to know that the work has some benefit.

Also, can you please sync with the Terraform codegen folk to make sure this split isn't going to cause any problems?

ArcturusZhang:
Hi @jhendrixMSFT I made some changes to make it does not generate `enums.go` when there were not any

fearthecowboy:
## success (version: 2.1.148)

PR #449, opened 2020-07-23
Emit header collections as a map

jhendrixMSFT:
For types annotated with x-ms-header-collection-prefix, emit them as a
map[string]<value>, where <value> is the type specified in the swagger.


catalinaperalta:
I think we should have a test in the testserver to cover this scenario.

Issue #259, opened 2020-02-13
Add custom marshalling for RO properties


catalinaperalta:
Related to complexgroup ReadOnlyProperty PutValid

jhendrixMSFT:
We should implement this in such a way that it only takes effect when sending requests to the REST endpoint.  We've had several customers complain that their custom logging/caching/round-tripping of JSON data is broken due to R/O fields being omitted during marshalling.

jhendrixMSFT:
https://gist.github.com/jhendrixMSFT/2340616d2583e853f3096e826953babb

jhendrixMSFT:
Fixed.

PR #450, opened 2020-07-28
Omit read-only content from request payloads

jhendrixMSFT:
Add `azure:"ro"` to fields marked as read-only.
Update azcore to version that supports exclusion of RO values.
Enabled RO test.
Added missing go.sum files.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/259

Issue #298, opened 2020-03-10
Convert header parsing to timerfc3339nano


catalinaperalta:

jhendrixMSFT:
Fixed

Issue #268, opened 2020-02-19
Finish implementing custom headers


catalinaperalta:
Need to consider cases where headers aren't specified in the swagger.

jhendrixMSFT:
To clarify, this is for sending custom headers.  Receiving all headers is possible via the raw response.

jhendrixMSFT:
Fixed

PR #452, opened 2020-07-28
enable test for setting custom header

jhendrixMSFT:


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/268

jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/298

src/generator/helpers.ts
```diff
@@ -11,7 +11,7 @@ import { ArraySchema, CodeModel, DictionarySchema, Language, Parameter, Schema,
 import { ImportManager } from './imports';
 
 export const dateFormat = '2006-01-02';
-export const datetimeRFC3339Format = 'time.RFC3339';
+export const datetimeRFC3339Format = 'time.RFC3339Nano';
```
catalinaperalta, line 14: Just wondering, there no case where we need time.RFC3339?

catalinaperalta: Or do the trailing zeros not matter?

jhendrixMSFT: Omitting the trailing zeros just saves a few bytes, which in itself isn't a big deal, however we use Nano in other places so it's more about consistency.

catalinaperalta: Ok, got it 👍

PR #454, opened 2020-07-29
Follow ups on the custom marshaler PR

ArcturusZhang:
- Add a condition to avoid generating empty marshalJSON function when the struct only has READ-ONLY fields.
- A minor refactor to increase readability

### DIFF
--- a/src/Model/CodeModelGo.cs
+++ b/src/Model/CodeModelGo.cs
@@ -71,9 +71,12 @@ namespace AutoRest.Go.Model
             //
             // we want the "2017-05-01-preview" portion
 
-            var i = _outDir.LastIndexOf('/');
-            var j = _outDir.LastIndexOf('/', i - 1);
-            return _outDir.Substring(j + 1, i - j - 1);
+            var segments = _outDir.Split('/');
+            if (segments.Length < 2)
+            {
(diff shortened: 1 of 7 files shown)


src/Model/CodeModelGo.cs
```diff
@@ -192,6 +192,10 @@ public IEnumerable<string> ModelImports
                 {
                     imports.UnionWith(CodeNamerGo.Instance.PageableImports);
                 }
+                if (ModelTypes.Any(mt => mt is CompositeTypeGo && mt.Cast<CompositeTypeGo>().NeedsCustomMarshalling))
+                {
+                    imports.Add(PrimaryTypeGo.GetImportLine("encoding/json"));
```
jhendrixMSFT, line 197: I found it, in CompositeTypeGo.cs:216.  Method AddImports(), you can move this to there.

ArcturusZhang: No wonder I just search `"encoding/json"` and found nothing...

ArcturusZhang:
Hi @jhendrixMSFT I just made some changes, could you please have a look?

fearthecowboy:
## success (version: 2.1.152)

PR #455, opened 2020-07-29
Add some validation to the namespace and refine error messages

### LINKED ISSUE #420: [CosmosDB] Unable to generate go sdk from specifications in azure-rest-api-specs. Error: FATAL: System.ArgumentOutOfRangeException: Index was out of range.
CosmosDB specifications exist [here](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/cosmos-db/resource-manager).
We created a PR to allow generation of SDK for GO [here](https://github.com/Azure/azure-rest-api-specs/pull/9793):.

Unfortunately, the autorest.go fails with the error: 
```
`specs/specification/cosmos-db/resource-manager/readme.md FATAL: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
[Azure/azure-sdk-for-go]Parameter name: startIndex
```
Detailed logs are [here]( https://sdkautomation.blob.core.windows.net/sdkautomation-pipeline/Azure/azure-rest-api-specs/9793/azure-sdk-for-go/logs.txt)

I tried generating the sdk locally as immediate solution, but the same issue persists. I am using AutoRest version 2.0.4413, Attached the screenshot for errors when attempted to run locally: 
image

ArcturusZhang:
Fixes #420

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 
 namespace AutoRest.Go
 {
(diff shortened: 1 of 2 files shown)


fearthecowboy:
## success (version: 2.1.151)

PR #457, opened 2020-07-29
Separate ARM vs data-plane poller files and generation

catalinaperalta:
This PR separates arm poller code generation from data-plane code generation, a slightly modified version of arm generation is used as a placeholder for dataplane generated code. 
Adding azartifacts to help track LRO progress for data-plane pollers.


src/generator/generator.ts
```diff
@@ -59,12 +61,22 @@ export async function protocolGen(host: Host) {
     if (pagers.length > 0) {
       host.WriteFile(`${filePrefix}pagers.go`, pagers, undefined, 'source-file-go');
     }
-    const pollers = await generatePollers(session);
-    if (pollers.length > 0) {
-      const pollingHelper = await generatePollersHelper(session);
-      host.WriteFile(`${filePrefix}pollers_helper.go`, pollingHelper, undefined, 'source-file-go');
-      host.WriteFile(`${filePrefix}pollers.go`, pollers, undefined, 'source-file-go');
+    if (openapiType === 'data-plane') {
```
jhendrixMSFT, line 64: Can this logic be pushed into `generatePollers()`, and have it return the correct content?

test/synapse/2019-06-01/azartifacts/pollers_helper.go
```diff
@@ -0,0 +1,768 @@
...
+const (
+	operationInProgress string = "InProgress"
+	operationCanceled   string = "Canceled"
+	operationFailed     string = "Failed"
+	operationSucceeded  string = "Succeeded"
+)
+
+var pollingCodes = [...]int{http.StatusNoContent, http.StatusAccepted, http.StatusCreated, http.StatusOK}
+
+type pollingTracker interface {
```
jhendrixMSFT, line 35: This is all the ARM stuff.  Can we have this file generate a little as possible?  Once we start generating real content the diff will be extremely noisy.

src/generator/generator.ts
catalinaperalta: So I personally think it's better to keep arm poller code separate from data-plane poller code and would like to keep them in two separate places. I can add this logic in generatePollers and then call the corresponding func from there.

test/synapse/2019-06-01/azartifacts/pollers_helper.go
catalinaperalta: done

src/generator/generator.ts
catalinaperalta: done

jhendrixMSFT: Agreed we should keep them separate.  But generator.ts shouldn't need to know anything about ARM/data-plane, it just calls some generic `createPollers()` func and the correct content is returned.

test/synapse/2019-06-01/azartifacts/pollers.go
```diff
@@ -0,0 +1,187 @@
...
+	Done() bool
+	Poll(ctx context.Context) (*http.Response, error)
+	FinalResponse(ctx context.Context) (*CreateDataFlowDebugSessionResponseResponse, error)
+	ResumeToken() (string, error)
+}
+
+type createDataFlowDebugSessionResponsePoller struct {
+	// the client for making the request
+	pipeline azcore.Pipeline
+	pt       pollingTracker
```
jhendrixMSFT, line 30: Where is pollingTracker defined?  Should it be deleted?

catalinaperalta: Deleted. Done.

src/generator/arm_pollers.ts
```diff
@@ -302,7 +302,7 @@ export async function generatePollers(session: Session<CodeModel>): Promise<stri
 
 
 // Creates the content in pollers_helper.go
-export async function generatePollersHelper(session: Session<CodeModel>): Promise<string> {
```
jhendrixMSFT, line 305: Doesn't this go away with your recent changes to armcore?

catalinaperalta: Yes, but that change will be implemented in a future PR. The code I have for the armcore implementation is ready to go once we get this in.

test/synapse/2019-06-01/azartifacts/notebook.go
```diff
@@ -0,0 +1,287 @@
...
+}
+
+// notebookOperations implements the NotebookOperations interface.
+type notebookOperations struct {
+	*Client
+}
+
+// CreateOrUpdateNotebook - Creates or updates a Note Book.
+func (client *notebookOperations) BeginCreateOrUpdateNotebook(ctx context.Context, notebookName string, notebook NotebookResource, notebookCreateOrUpdateNotebookOptions *NotebookCreateOrUpdateNotebookOptions) (*NotebookResourcePollerResponse, error) {
+	return nil, nil
```
catalinaperalta, line 43: @jhendrixMSFT FYI, this is how I'm outputting the operations at the moment. Or should I add an errors.New("Not yet implemented"). Processing and initializing a request for a data-plane poller is still TBD and the helper code is not being outputted.

jhendrixMSFT: Probably best to emit it as `errors.New("NYI")` just to be safe.

src/generator/models.ts
```diff
@@ -553,7 +553,7 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
     if (prop.schema.type === SchemaType.DateTime) {
       text += `\t\t${prop.language.go!.name} *${prop.schema.language.go!.internalTimeType} \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
     } else if (prop.language.go!.isAdditionalProperties || prop.language.go!.needsXMLDictionaryUnmarshalling) {
-      text += `\t\t${prop.language.go!.name} *additionalProperties \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
+      text += `\t\t${prop.language.go!.name} *map[string]string \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
```
jhendrixMSFT: This is incorrect.  The `additionalProperties` type is defined in a helper that's only emitted when `needsXMLDictionaryUnmarshalling` is true.  The fact that this is being emitted when the helper isn't suggests that the above condition has a bug.

jhendrixMSFT: It's also interesting that this failed only for Go 1.14.  Do you know why that is?

PR #458, opened 2020-07-29
Add support for unmarshalling XML into maps

jhendrixMSFT:
Use the element name for the map key and the element value for the map
value.


src/generator/xmlAdditionalProps.ts
```diff
@@ -0,0 +1,50 @@
...
+  text += imports.text();
+  text += `
+type additionalProperties map[string]string
+
+// UnmarshalXML implements the xml.Unmarshaler interface for additionalProperties.
+func (ap *additionalProperties) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
+	tokName := ""
+	for t, err := d.Token(); err == nil; t, err = d.Token() {
+		switch tt := t.(type) {
+		case xml.StartElement:
```
catalinaperalta, line 31: Just wondering, shouldn't this use the "start" param? So that it's actually used.

jhendrixMSFT: It depends on how you want to unmarshal the content.  In this case, we need to read each individual element name since they're not fixed.
The [docs](https://golang.org/pkg/encoding/xml/#Unmarshaler) talk a little bit more about this, although they are a bit vague.

PR #459, opened 2020-07-30
check null before checking namespace

ArcturusZhang:
For those packages that does not have a go configuration, null reference exception is thrown and gives a lot of noises

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -47,7 +47,8 @@ namespace AutoRest.Go
             // check if the namespace contains illegal characters
             var ns = Settings.Instance.Host.GetValue<string>("namespace").Result;
             var r = new Regex(@"^[a-z][a-z0-9_]*[a-z0-9]?$");
-            if (!r.IsMatch(ns))
+            // skip this check if namespace is not specified
+            if (!string.IsNullOrEmpty(ns) && !r.IsMatch(ns))
             {
                 throw new InvalidOperationException($"namespace can only contains lower case letters, numbers and underscore");
             }


src/CodeGeneratorGo.cs
```diff
@@ -47,7 +47,8 @@ public override async Task Generate(CodeModel cm)
             // check if the namespace contains illegal characters
             var ns = Settings.Instance.Host.GetValue<string>("namespace").Result;
             var r = new Regex(@"^[a-z][a-z0-9_]*[a-z0-9]?$");
-            if (!r.IsMatch(ns))
+            // skip this check if namespace is not specified
+            if (!string.IsNullOrEmpty(ns) && !r.IsMatch(ns))
```
jhendrixMSFT, line 51: Where do we get the namespace if it's not specified?

ArcturusZhang: There are several cases here:
1. They do not have a go configuration - therefore namespace is empty
1. They have a go configuration, but do not specify a namespace.

For the first case, the output-folder has a default value or a value inherited from other languages (usually something/Generated) - in some case this path is not in the go SDK directory, we should not worry about this.

For the second case, if they do not specify a namespace, everything could still go well if they just specify a valid output-folder - therefore I suppose we should just omit the validation if we could not get one

jhendrixMSFT: The convention in Go is that the package name (namespace here) must match the leaf directory name.  For track 2, we had the `--namespace` param but I ended up removing it, defaulting to using the leaf directory name (it was simpler then ensuring that the leaf directory name matched the namespace).  I feel like we default to using the leaf directory name in track 1 if `--namespace` is omitted.  Don't we also verify someplace that the namespace and leaf directory name match?  I'm wondering if we can remove the need for `--namespace`.  Maybe that's worth a separate investigation.

fearthecowboy:
## success (version: 2.1.153)

src/CodeGeneratorGo.cs
ArcturusZhang: Well, since they usually use the `namespace` this way:
```
output-folder: $(go-sdk-folder)/services/$(namespace)/mgmt/2020-07-31/$(namespace)
```
I suppose instead of validating the leaf directory, how about we just validate the whole output-folder?

If output-folder contains upper case letters, it would cause issues on windows machines. Also if output-folder contains dot (`.`), go vet would report issues.

jhendrixMSFT: Validating the entire directory is a good idea.  We already do some of this, e.g. `--preview-chk`.  Can probably expand it to more validation.  It just gets a bit tricky for data-plane as the directory structure isn't well-formed like ARM.

PR #461, opened 2020-08-03
Added test server coverage

jhendrixMSFT:
Implemented various tests that were missing.
Removed some superfluous code in xmlgroup_test.go
Removed other_mods.go now that CI has been fixed.

### DIFF
--- a/test/autorest/errorsgroup/errorsgroup_test.go
+++ b/test/autorest/errorsgroup/errorsgroup_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"generatortests/autorest/generated/errorsgroup"
 	"generatortests/helpers"
+	"net/http"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


test/autorest/errorsgroup/errorsgroup_test.go
```diff
@@ -104,7 +105,12 @@ func TestGetPetByIDSuccess1(t *testing.T) {
 }
 
 func TestGetPetByIDSuccess2(t *testing.T) {
-	t.Skip("multiple responses with different schemas NYI")
+	client := getClient(t)
+	result, err := client.GetPetByID(context.Background(), "django")
```
catalinaperalta, line 109: If this is a Get should there also be a comparison of what we actually got? (If we are receiving something)

jhendrixMSFT: The response for this test only returns a 200 (no response body or headers).

jhendrixMSFT: Wait let me look into this a bit more....

jhendrixMSFT: Per the swagger, for an HTTP 200 response you get back a Pet schema.  For an HTTP 202 response there is no schema.

catalinaperalta: Ah ok, that sounds good then! Is the 202 the final state though for this test?

catalinaperalta: Well yeah it should be, nvm. Looks good!

PR #463, opened 2020-08-04
Limit regeneration to eight concurrent builds

jhendrixMSFT:
Consolidate scripts to .scripts directory.


.scripts/semaphore.js
```diff
@@ -0,0 +1,103 @@
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// original sources can be found at https://github.com/abrkn/semaphore.js
+;(function(global) {
```
catalinaperalta, line 3: I'm unfamiliar with what the semi-colon means at the beginning here. What is it?

catalinaperalta: Nevermind, I looked it up.

PR #464, opened 2020-08-05
Adding tags on scalar value response envelopes

catalinaperalta:
Add json tags only on response envelopes that wrap a scalar value. 

The armcore PR will include the second part of this which is changing the unmarshalling statements for pollers that use these response envelopes.


src/common/helpers.ts
```diff
@@ -66,3 +66,17 @@ export function hasAdditionalProperties(obj: ObjectSchema): DictionarySchema | u
...
+    case SchemaType.Boolean:
+    case SchemaType.ByteArray:
+    case SchemaType.Integer:
+    case SchemaType.Number:
+    case SchemaType.String:
+      return true;
+    default:
+      return false;
+  }
+}
```
jhendrixMSFT, line 82: Nit: missing new-line

test/network/2020-03-01/armnetwork/models.go
```diff
@@ -10169,8 +10169,8 @@ type StringPollerResponse struct {
 // StringResponse is the response envelope for operations that return a string type.
 type StringResponse struct {
```
jhendrixMSFT, line 10170: There is also a `StringArrayResponse` in this code.  Not sure if it should be tagged, can you PTAL to see if it needs addressing.

test/autorest/generated/bytegroup/models.go
```diff
@@ -13,10 +13,10 @@ import (
 // ByteArrayResponse is the response envelope for operations that return a []byte type.
 type ByteArrayResponse struct {
 	// RawResponse contains the underlying HTTP response.
-	RawResponse *http.Response
+	RawResponse *http.Response `json:"-"`
```
jhendrixMSFT, line 16: These tags aren't necessary as the result isn't sent in an envelope (remember, the envelope is ARM-specific).  Same applies to all models.go here.

test/network/2020-03-01/armnetwork/models.go
catalinaperalta: Yes the condition for adding tags was limiting to the original scalarValueName, this has been changed now. So all array types in the response envelopes will have tags. Although I'm wondering should it only be array types of scalar types? Or any array type?

src/common/helpers.ts
catalinaperalta: Done

test/autorest/generated/bytegroup/models.go
catalinaperalta: So I added a condition to only add tags in the case that the open-api type we receive is ARM, but since the default is ARM even when not specified, all of this codegen defaults to ARM and adding the tags. :/

test/network/2020-03-01/armnetwork/models.go
```diff
@@ -10148,10 +10148,10 @@ type SessionIDs struct {
 // StringArrayResponse is the response envelope for operations that return a []string type.
 type StringArrayResponse struct {
 	// RawResponse contains the underlying HTTP response.
-	RawResponse *http.Response
+	RawResponse *http.Response `json:"-"`
 
 	// Response for ApplicationGatewayAvailableServerVariables API service call.
-	StringArray *[]string
+	StringArray *[]string `json:"StringArray,omitempty"`
```
jhendrixMSFT, line 10154: I believe this should be named `Value` with a tag of the same (that's how it is in track 1).

catalinaperalta: Hmmm let me look at the naming rules/ transform that happens before this. I've seen this sort of naming in several other files.

catalinaperalta: So this is what we do in the transform: 
```
case SchemaType.Array:
      const arraySchema = <ArraySchema>schema;
      const arrayElem = <Schema>arraySchema.elementType;
      return `${recursiveTypeName(arrayElem)}Array`;
```

catalinaperalta: Also we name ByteArrays as `ByteArray`

catalinaperalta: If we want to change this, I would say we should do it in a separate PR since it will impact many generated files

jhendrixMSFT: For a little more context, in track 1 we referred to the enclosing type (`StringArrayResponse` in this example) as a "synthetic type".  You can find the rules for determining when to create a synthetic type [here](https://github.com/Azure/autorest.go/blob/4842cb44812dfd73786d236d32abc944bf821654/src/Extensions.cs#L380).

catalinaperalta: Ok sounds good, I'll add a similar rule for track2, but I think it would be better to do it separately to not mix it in with these changes.

jhendrixMSFT: Hmm I think this also gets into a data-plane/management-plane thing.  For data-plane I think the naming as-is makes sense as there's no envelope requirement.  For ARM, naming it `Value` is more inline with their RPC requirements.  At any rate I agree that we can solve that separately.

catalinaperalta: I see, I'll consider different behaviors depending on whether we're generating for ARM or data-plane when I implement it.

src/common/helpers.ts
```diff
@@ -66,3 +66,18 @@ export function hasAdditionalProperties(obj: ObjectSchema): DictionarySchema | u
   }
   return undefined;
 }
+
+export function isScalar(schema: Schema): boolean {
```
jhendrixMSFT, line 70: Will this include enums and arrays of enums?

catalinaperalta: Added Choice and SealedChoice schema type to the switch. I was also wondering about time types and time arrays....should those also have tags?

catalinaperalta: I also saw Odata types...what does that mean? And should they be added?

jhendrixMSFT: For track 1 it was any  primary type.  The track 1 modeler defines primary types [here](https://github.com/Azure/autorest.common/blob/04b5593e34604e993bdda8467def411849f5b90b/src/Model/KnownPrimaryType.cs).  So yeah the time types will need to be added too (possibly others, I didn't look closely at both lists).

catalinaperalta: Ok I added all the schema types I think are applicable to us in isScalar.

src/common/helpers.ts
```diff
@@ -66,3 +66,26 @@ export function hasAdditionalProperties(obj: ObjectSchema): DictionarySchema | u
...
+  switch (schema.type) {
+    case SchemaType.Array:
+      return isScalar((<ArraySchema>schema).elementType);
+    case SchemaType.Binary:
```
jhendrixMSFT, line 74: I think we can remove this one since we don't create a response envelope for binary responses.  The rest look good.

catalinaperalta: Removed.

src/generator/models.ts
```diff
@@ -158,10 +158,12 @@ class StructDef {
         readOnly = ` azure:"ro"`;
       }
       let tag = ` \`${this.Language.marshallingFormat}:"${serialization}"${readOnly}\``;
-      // if this is a response type then omit the tag IFF the marshalling format is
-      // JSON, it's a header or is the RawResponse field.  XML marshalling needs a tag.
+      // if this is a response type then omit the tag IFF the marshalling formatvalue is non-scalar
+      // for JSON marshalling types or it's a header.  XML marshalling needs a tag.
       // also omit the tag for additionalProperties
-      if ((this.Language.responseType === true && (this.Language.marshallingFormat !== 'xml' || prop.language.go!.name === 'RawResponse')) || prop.language.go!.isAdditionalProperties) {
+      if (prop.language.go!.name === 'RawResponse' && prop.language.go!.addTag === true && this.Language.marshallingFormat !== 'xml') {
```
jhendrixMSFT, line 164: I think this can be simplified to remove the need for `addTag`.
`if (prop.language.go!.name === 'RawResponse' && this.Language.marshallingFormat !== 'xml')`

src/generator/models.ts
```diff
@@ -158,10 +158,12 @@ class StructDef {
...
+        tag = '`json:"-"`';
+      } else if (prop.language.go!.name === 'RawResponse' || prop.language.go!.isAdditionalProperties || (this.Language.responseType === true && prop.language.go!.addTag !== true && this.Language.marshallingFormat !== 'xml')) {
```
jhendrixMSFT, line 166: This can also be simplified to remove the need for `addTag`, just call `isScalar()` here.  Note you might need to make this its own `if` clause (not an `else if`) given that the previous `if` clause is compound.

test/network/2020-03-01/armnetwork/models.go
jhendrixMSFT, line 10170: I think that the `handleResponse()` will need to change for this to unmarshal into `StringResponse`, not `StringResponse.Value`.

src/generator/models.ts
catalinaperalta: So I need to leave `addTag` as part of the check here, because "RawResponse" can exist both on scalar and non-scalar response envelopes and I think it's the cleanest way of making sure it's only added on RawResponse when the other field is a scalar. 
Add tag is set during the transform when creating the scalar type response envelope, there for no additional looping or processing is needed later on.

src/generator/models.ts
catalinaperalta: So the reason why addTag is also necessary here is almost the same as above, with the addition that addTag is only added on scalar types being returned in the body and not the headers. There are response envelopes that return scalar types in headers and those are added on the response envelopes. So calling isScalar wont be sufficient for ruling out the header case.

catalinaperalta:
Closing since this change is not needed.

PR #465, opened 2020-08-05
Add armcore generation for ARM pollers

catalinaperalta:
Update ARM poller implementation to use the poller interface from the armcore module.


test/network/2020-03-01/armnetwork/pollers.go
```diff
@@ -8867,120 +3548,47 @@ type StringPoller interface {
...
-	u, err := url.Parse(p.pt.finalGetURL())
-	if err != nil {
-		return nil, err
-	}
-	req := azcore.NewRequest(http.MethodGet, *u)
-	if err != nil {
-		return nil, err
-	}
-	resp, err := p.pipeline.Do(ctx, req)
+	temp := ""
```
catalinaperalta, line 3565: @jhendrixMSFT look at this

test/autorest/generated/lrogroup/pollers.go
```diff
@@ -7,85 +7,48 @@ package lrogroup
...
-	"net/url"
 	"time"
 )
 
 // HTTPPoller provides polling facilities until the operation completes
 type HTTPPoller interface {
 	Done() bool
 	Poll(ctx context.Context) (*http.Response, error)
-	FinalResponse() *http.Response
+	FinalResponse(ctx context.Context) (*http.Response, error)
```
jhendrixMSFT, line 20: Why did this change?  Do we really need to perform a final GET for *http.Response operations?

catalinaperalta: We do not :( But we also do not expose the latest response method for the pollers :/

jhendrixMSFT: I wonder, can we just omit FinalResponse for HTTPPoller types?

catalinaperalta: I suppose that's feasible. I mean we do not perform an additional get for the final http.Response, but it does mean it will be special cased for people who decide to hand write the code without PollUntilDone. We would have to be clear in the documentation that Http Pollers will be different.

src/generator/arm_pollers.ts
```diff
@@ -300,779 +214,14 @@ export async function generateARMPollers(session: Session<CodeModel>): Promise<s
...
-		pollingRequestURI pollingMethodType = "RequestURI"
-	
-		// pollingUnknown indicates an unknown polling method and is the default value.
-		pollingUnknown pollingMethodType = ""
-	)
-	`;
-  return text;
-}
-
+}
```
jhendrixMSFT, line 227: Missing new-line.

test/autorest/generated/lrogroup/pollers.go
jhendrixMSFT: I see your point, however I personally think this is acceptable.  Having to call FinalResponse() in this case (which does nothing), just for sake of consistency, seems a bit odd to me.

@JeffreyRichter what are your thoughts.

test/network/2020-03-01/armnetwork/pollers.go
jhendrixMSFT: Yes I remember now.  I misspoke earlier, we don't need to populate `ResponseEnvelope.Widget` before handing it off to the unmarshaller, so we should be able to remove this.

catalinaperalta: But we do for scalar types, because if not it's nil and then json.Unmarhsal fails since it's nil.

catalinaperalta: Done

src/generator/arm_pollers.ts
catalinaperalta: done

test/autorest/generated/lrogroup/pollers.go
jhendrixMSFT: Per internal conversation will leave as-is for consistency and versioning.

test/autorest/generated/paginggroup/pollers.go
```diff
@@ -28,103 +27,41 @@ type productResultPagerPoller struct {
...
-		if err != nil {
-			return nil, err
-		}
+	s := &productResultPager{}
+	resp, err := p.pt.PollUntilDone(ctx, frequency, p.pipeline, s)
+	if err != nil {
+		return nil, err
 	}
-	return p.FinalResponse(ctx)
+	return p.handleResponse(&azcore.Response{resp})
```
jhendrixMSFT, line 64: Why did this change?

src/generator/arm_pollers.ts
```diff
@@ -123,18 +90,26 @@ export async function generateARMPollers(session: Session<CodeModel>): Promise<s
...
     let pollUntilDoneReturn = 'p.FinalResponse(), nil';
+    let pollUntilDone = `return p.pt.PollUntilDone(ctx, frequency, p.pipeline, nil)`;
     let handleResponse = '';
     const schemaResponse = <SchemaResponse>poller.op.responses![0];
     let unmarshalResponse = 'nil';
     let pagerFields = '';
-    let finalResponseCheckNeeded = false;
     if (isPageableOperation(poller.op)) {
+      function finalPagerProcessing(name: string, params: string): string {
+        return `s := &${camelCase(responseType)}{}
```
jhendrixMSFT, line 106: You'll want to use more than one letter here or you run the risk of it colliding with the receiver var which is always one letter (the first letter of the type).

src/generator/arm_pollers.ts
```diff
@@ -147,69 +122,43 @@ export async function generateARMPollers(session: Session<CodeModel>): Promise<s
...
-        }
-        if !resp.HasStatusCode(pollingCodes[:]...) {
-          return nil, p.pt.handleError(resp)
-        }
-        return &result, ${unmarshalResponse}
-      }
-      `;
       finalResponse = `FinalResponse(ctx context.Context) (*${responseType}, error) {`;
-      finalResponseCheckNeeded = true;
+      let respType = `s := &${responseType}{${schemaResponse.schema.language.go!.responseType.value}: &${schemaResponse.schema.language.go!.name}{}}`;
```
jhendrixMSFT, line 138: More than one letter 's'.

test/autorest/generated/paginggroup/pollers.go
catalinaperalta: Everything here is handled in PollUntilDone (including the call to FinalResponse on the armcore side), the only thing left to do is parse the final response we receive into the pager instance we want to return.

src/generator/arm_pollers.ts
catalinaperalta: I'm thinking of something like `temp` or maybe `respType` like the method param. Any preference?

jhendrixMSFT: respType sounds good to me

catalinaperalta: done

src/generator/arm_pollers.ts
catalinaperalta: done

test/autorest/generated/paginggroup/pollers.go
jhendrixMSFT: I get it now, and we have to do this here because it's slightly different for pager-pollers.

PR #467, opened 2020-08-10
update to latest test server and regenerate code

jhendrixMSFT:


src/generator/operations.ts
```diff
@@ -438,6 +447,14 @@ function createProtocolRequest(codeModel: CodeModel, client: string, op: Operati
         text += '\tunencodedParams := []string{}\n';
       }
       for (const qp of values(unencodedParams)) {
+        let setter: string;
+        if (qp.protocol.http?.explode === true) {
+          setter = `\tfor _, qv := range ${getParamName(qp)} {\n`;
+          setter += `\t\tunencodedParams = append(unencodedParams, "${qp.language.go!.serializedName}="+qv)\n`;
```
catalinaperalta, line 453: Just want to know what the = for after `${qp.language.go!.serializedName}`? I dont think I saw it in any of the testserver examples.

src/generator/operations.ts
```diff
@@ -426,7 +426,16 @@ function createProtocolRequest(codeModel: CodeModel, client: string, op: Operati
     if (encodedParams.length > 0) {
       text += '\tquery := u.Query()\n';
       for (const qp of values(encodedParams)) {
-        text += emitQueryParam(qp, `query.Set("${qp.language.go!.serializedName}", ${formatParamValue(qp, imports)})`);
+        let setter: string;
+        if (qp.protocol.http?.explode === true) {
```
catalinaperalta, line 430: For my own knowledge, what does explode mean for us in swagger?

jhendrixMSFT: It comes from https://swagger.io/specification/.  Unfortunately I can't link to the exact section but if you search for "explode" it's the first hit.

src/generator/operations.ts
jhendrixMSFT: When exploding a query param `queryParam` with the following values `foo,bar,baz` we want to generate the following sequence.

`queryParam=foo&queryParam=bar&queryParam=baz`

You are right there's no codegen with this as we don't have a test server example that uses exploded, unencoded query params.

catalinaperalta: Ok got it 👍 yeah we should have something that tests this in particular

src/generator/operations.ts
catalinaperalta: got it, thanks for the info!

PR #469, opened 2020-08-19
Make client constructors no-fail

jhendrixMSFT:
Moved parsing out of the constructor, into the API call.
Fixed an issue with imports not getting added to client.go files.

### DIFF
--- a/test/network/2020-03-01/armnetwork/vpnsites.go
+++ b/test/network/2020-03-01/armnetwork/vpnsites.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"net/http"
 	"net/url"
+	"path"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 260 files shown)


test/network/2020-03-01/armnetwork/vpnsites.go
```diff
@@ -161,11 +164,15 @@ func (client *vpnSitesOperations) ResumeDelete(token string) (HTTPPoller, error)
...
+	u, err := url.Parse(client.u)
+	if err != nil {
+		return nil, err
+	}
 	urlPath := "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Network/vpnSites/{vpnSiteName}"
 	urlPath = strings.ReplaceAll(urlPath, "{subscriptionId}", url.PathEscape(client.subscriptionID))
 	urlPath = strings.ReplaceAll(urlPath, "{resourceGroupName}", url.PathEscape(resourceGroupName))
 	urlPath = strings.ReplaceAll(urlPath, "{vpnSiteName}", url.PathEscape(vpnSiteName))
-	u, err := client.u.Parse(path.Join(client.u.Path, urlPath))
+	u, err = u.Parse(urlPath)
```
catalinaperalta, line 175: Just double checking here, this will append urlPath to the already existing path on the url?

jhendrixMSFT: Good catch, no it doesn't work as expected.  Will fix.

jhendrixMSFT: Fixed

PR #475, opened 2020-08-31
Migrate to Azure DevOps

jhendrixMSFT:


jhendrixMSFT:
I don't think the GoTool works with this ancient version of Node.

@ArcturusZhang did you ever get the track 1 generator to work with a modern version of Node?

ArcturusZhang:
@jhendrixMSFT I tried node v12, it could run but cannot build properly unfortunately...

jhendrixMSFT:
@ArcturusZhang getting closer but it looks to me that the build hung.  How do you build this locally?

ArcturusZhang:
> @ArcturusZhang getting closer but it looks to me that the build hung. How do you build this locally?

Usually I just run `npm run build` which should be an alias of `gulp build`. Should we run `npm run prepare` before run any other commands?

jhendrixMSFT:
I don't believe so.  How long does your local build take?  From the log, the `npm install` process is killed after 10 minutes.  I can try increasing the step time.

fearthecowboy:
## success (version: 2.1.159)

jhendrixMSFT:
There were two problems.

1. Enable submodule initialization in the devops pipeline
2. The OS locale was incorrect, which caused sort order changes.

Incidentally, fixing #2 also fixed that annoying issue we were seeing with the casing of `-1.034e-20`

ArcturusZhang:
> There were two problems.
> 
> 1. Enable submodule initialization in the devops pipeline
> 2. The OS locale was incorrect, which caused sort order changes.
> 
> Incidentally, fixing the locale also fixed that annoying issue we were seeing with the casing of `-1.034e-20`

I was driving crazy by the casing of -1.034e-20

PR #477, opened 2020-09-02
Separate operation group clients

jhendrixMSFT:
This change separates the operation groups into distinct public client
types.  They are no longer accessed through helper methods on the root
client type.

Made internal client types exported, renaming them to <Group>Client.
Aded operation group client constructors.
Added Do() method on group clients for access to its pipeline.
Exported API helper methods, they are only available via the client.
Fixed all tests to use the new codegen.


test/autorest/generated/arraygroup/array.go
```diff
@@ -156,30 +156,41 @@ type ArrayOperations interface {
 	PutUUIDValid(ctx context.Context, arrayBody []string) (*http.Response, error)
 }
 
-// arrayOperations implements the ArrayOperations interface.
-type arrayOperations struct {
+// ArrayClient implements the ArrayOperations interface.
```
catalinaperalta, line 159: So we are going to output the interface definition for all management plane but not data-plane?

jhendrixMSFT: Correct, that will happen in a separate PR.

catalinaperalta (approved):
LGTM

PR #478, opened 2020-09-03
Refine error messages for preview check

ArcturusZhang:
I noticed that when swagger PR was opened and similar questions come all the time, therefore I refine the error messages to include more information and a doc link to help them understand what is wrong

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -52,9 +52,12 @@ namespace AutoRest.Go
             {
                 throw new InvalidOperationException($"namespace can only contains lower case letters, numbers and underscore");
             }
-
-            // we always do a preview check here. To disable preview-chk, you need to explicitly assign `--ignore-preview-chk` flag
-            PreviewCheck(folder);
+            // if preview-chk:true is specified verify that preview swagger is output under a preview subdirectory.
+            // this is a bit of a hack until we have proper support for this in the swagger->sdk bot so it's opt-in.
+            if (Settings.Instance.Host.GetValue<bool>("preview-chk").Result)
+            {
+                PreviewCheck(folder);
+            }
 
             var codeModel = cm as CodeModelGo;
             if (codeModel == null)
@@ -165,8 +168,8 @@ namespace AutoRest.Go
 
         private void PreviewCheck(string folder)
         {
-            // skip the preview check when explicitly assigned flag --ignore-preview-chk
-            if (Settings.Instance.Host.GetValue<bool>("ignore-preview-chk").Result)
+            var isProfile = Settings.Instance.Host.GetValue<bool>("is-profile").Result;
+            if (isProfile) // if a tag is designed to serve profile building, we could omit the preview check
             {
                 return;
             }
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
Hi @jhendrixMSFT could you please have a look?

src/CodeGeneratorGo.cs
```diff
@@ -169,22 +169,25 @@ private string StagingDir()
 
         private void PreviewCheck(string folder)
         {
-            var isProfile = Settings.Instance.Host.GetValue<bool>("is-profile").Result;
-            if (isProfile) // if a tag is designed to serve profile building, we could omit the preview check
+            var ignore = Settings.Instance.Host.GetValue<bool>("ignore-preview-chk").Result;
```
jhendrixMSFT, line 172: Minor issue, but it seems weird to me that we'll now invoke the generator with `--preview-chk ---ignore-preview-chk` flags.  I understand why we do this, but I think leaving it as `--is-profile` (or possibly shorten it to `--profile`) doesn't look as strange.

ArcturusZhang: Yeah, I agree that this is very weird.

We have added more strict checking for the output-folder path, which is making some tags to be weird - for instance some big composite tags (package-composite-v1 in security). The input file of those tags may change constantly and may not have to a profile tag. 

Maybe we should switch to stable when all the swagger is not preview for this kind of tags?

jhendrixMSFT: So originally this is what I wanted to do (i.e. only generate the package as stable if all swaggers are stable) however this turned out to be problematic (sorry I don't recall the details anymore).  I think there's a higher-level question here about combining preview and stable swaggers in composite packages that needs to be addressed.

Mostly my comment was about naming the switch `--ignore-preview-chk`, as using it along with `--preview-chk` which seems counter-intuitive.

jhendrixMSFT (approved):

src/CodeGeneratorGo.cs
jhendrixMSFT: Hmm, what if we enabled preview-chk without a switch when building from a tag?

ArcturusZhang: Since the `preview-chk` flag is enabled by default in the pipeline (defined in azure-sdk-for-go), I suppose we could enable this as default but introduce another flag to disable this to achieve the same behaviour as previous (ie. optional preview check).
By introducing the native flag of `ignore-preview-chk`, we could hard-code this into the tag descriptions in `readme.go.md` to serve both profile tags and composite tags.

jhendrixMSFT: Per our offline conversation I think we need more data before changing the semantics of `preview-chk`.

Also, you might want to split the better error messages into a separate PR so you can get those benefits now while we sort the other stuff out.

ArcturusZhang: Left me separate the changes of the option flags out of this PR.

This PR will only serve the error message refinement.

ArcturusZhang:
Hi @jhendrixMSFT I have reverted all other changes except the refinement of error messages. Please have a look again

jhendrixMSFT (approved):

fearthecowboy:
## success (version: 2.1.156)

PR #479, opened 2020-09-03
Merge test server codegen with tests

jhendrixMSFT:
Keep the tests and codegen in the same dir which is the Go idiom.
Codegen files have the prefix 'zz_generated_' to make the easy to find
and sort at the bottom.
Removed --clear-output-folder from autorest.  Clean up all codegen files
before invoking autorest.


.scripts/regeneration.js
```diff
@@ -10,7 +11,7 @@ const swaggerDir = 'src/node_modules/@microsoft.azure/autorest.testserver/swagge
 const goMappings = {
     'additionalpropsgroup': 'additionalProperties.json',
     'arraygroup': 'body-array.json',
-    'azurereportgroup': 'azure-report.json',
+    //'azurereportgroup': 'azure-report.json', used for reporting test coverage
```
catalinaperalta, line 14: Why arent we including azurereportgroup and reportgroup?

catalinaperalta (approved):
LGTM. Just left one question

.scripts/regeneration.js
jhendrixMSFT: They're used as part of the `covreport` tool, so I moved the packages under its directory.  Unfortunately we can't specify custom output directories at present.

This will likely change again with the changes for generating data-plane packages as I would like to generate these packages in-place and not under a subdirectory.  It also means we'll need the ability to specify the open-api type per swagger as I know that some of them we want to treat as ARM (LROs maybe others?).

PR #480, opened 2020-09-03
Adding multiple requests support for operations

catalinaperalta:
Adding multiple request support for operations and generating media_types.json from the testserver. 

PR will be in draft until I add tests for mediatypesgroup.


test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
```diff
@@ -0,0 +1,206 @@
...
+	"net/url"
+	"path"
+)
+
+// MediaTypesClientOperations contains the methods for the MediaTypesClient group.
+type MediaTypesClientOperations interface {
+	// AnalyzeBody - Analyze body, that could be different media types.
+	AnalyzeBody(ctx context.Context, contentType ContentType, input azcore.ReadSeekCloser) (*StringResponse, error)
+	// AnalyzeBodyWithSourcePath - Analyze body, that could be different media types.
+	AnalyzeBodyWithSourcePath(ctx context.Context, mediaTypesClientAnalyzeBodyWithSourcePathOptions *MediaTypesClientAnalyzeBodyWithSourcePathOptions) (*StringResponse, error)
```
jhendrixMSFT, line 24: Yikes that second param type name is gnarly.  Is this an M4 issue?

src/common/helpers.ts
```diff
@@ -8,15 +8,14 @@ import { values } from '@azure-tools/linq';
...
-  if (op.requests!.length > 1) {
-    throw console.error('multiple requests NYI');
-  }
   let params = new Array<Parameter>();
   if (op.parameters) {
     params = params.concat(op.parameters);
   }
-  if (op.requests![0].parameters) {
-    params = params.concat(op.requests![0].parameters);
+  for (const req of values(op.requests)) {
```
jhendrixMSFT, line 15: I'm surprised this doesn't cause unnecessary params.  If there's some magic that happens during transformation that prevents it you should add a comment here explaining it.

src/transform/transform.ts
```diff
@@ -223,7 +223,19 @@ function processOperationRequests(session: Session<CodeModel>) {
...
+          let name = op.language.go!.name;
+          if (req.protocol.http!.knownMediaType === 'json') {
+            name = name + 'With' + req.parameters![0].schema.language.go!.name;
+          }
+          newOp.language.go!.name = name;
+          newOp.language.go!.protocolNaming = new protocolMethods(newOp.language.go!.name);
+          group.addOperation(newOp);
+        }
+        group.operations.splice(group.operations.indexOf(op), 1);
+        continue;
```
jhendrixMSFT, line 238: Looks like this `continue` will skip the rest of the processing for this operation.  I wonder if we need to perform the operation split first, in a separate loop?

catalinaperalta: I think performing the continue here is fine because if you see right above I am removing the original operation since it it unnecessary from this point onwards and wont ever be used. Per request, the operation is copied and appended to the list of operations. Each of those new operations will be processed fully.

jhendrixMSFT: D'oh I missed that, makes sense.

src/common/helpers.ts
catalinaperalta: I dont think this should be producing unnecessary params. In the case of multiple requests, those operations will be separated out after processing the initial operation. For most cases there will only be one request, therefore it will only return the parameters of that request. I can add a comment explaining that in the case of multiple requests the only time this will aggregate parameters for all requests is for the transform and that the operation with multiple requests will later be removed.

test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
catalinaperalta: I think this is more on our end, since we append the "ClientAnalyzeBodyWithSourcePath" and "Options"

src/common/helpers.ts
jhendrixMSFT: I see.  Does it need to aggregate params during transformation or is it safe to leave it as `op.requests![0]` then?

test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
jhendrixMSFT: Yes you're right.  I wonder if we should simply name the param `options` (the type name will likely have to remain as-is).  What do you think?

catalinaperalta: If I recall correctly we had this as options in a previous iteration of the code generator, but we changed it to be this way to support different parameter groupings and to use the name defined by the swagger. We could discuss maybe changing the name to "options" when there is only one set of optional params for an operation.

src/common/helpers.ts
catalinaperalta: aggregateParameters is also used in the namer before we get to the transform, so it is necessary to parse all of the parameters for the operation.

src/generator/operations.ts
```diff
@@ -681,7 +701,7 @@ function createProtocolResponse(op: Operation, imports: ImportManager): string {
       }
     }
     const mediaType = getMediaType(response.protocol);
-    if (mediaType === 'none' || mediaType === 'binary') {
+    if (mediaType === 'none' || mediaType === 'binary' || mediaType === 'text') {
```
catalinaperalta, line 704: Fixing this

jhendrixMSFT, line 704: It might be better to flip the condition to marshal for `json` and `xml`.  That way we don't have to keep updating it if/when other media types are added (e.g. this will likely come up again when adding support for form data).

src/generator/operations.ts
```diff
@@ -575,9 +575,29 @@ function createProtocolRequest(codeModel: CodeModel, op: Operation, imports: Imp
       text += '\treturn req, nil\n';
     }
   } else if (mediaType === 'binary') {
-    const contentType = op.requests![0].protocol.http!.mediaTypes[0];
+    let contentType = `"${op.requests![0].protocol.http!.mediaTypes[0]}"`;
+    if (op.requests![0].protocol.http!.mediaTypes.length > 1) {
+      for (const param of values(op.requests![0].parameters)) {
+        // If a request defined more than one possible media type, then the param is expected to be synthesized from modelerfour
+        // and should be a SealedChoice schema type that account for the acceptable media types defined in the swagger. 
+        if (param.origin !== undefined && param.origin === 'modelerfour:synthesized/content-type' && param.schema.type === SchemaType.SealedChoice) {
```
jhendrixMSFT, line 583: You should be able to omit the `undefined` check.

src/generator/operations.ts
```diff
@@ -575,9 +575,29 @@ function createProtocolRequest(codeModel: CodeModel, op: Operation, imports: Imp
...
+    text += `\treturn req, req.SetBody(${bodyParam?.language.go!.name}, ${contentType})\n`;
+  } else if (mediaType === 'text') {
+    imports.add('strings');
+    let bodyParam = '';
+    for (const param of values(op.requests![0].parameters)) {
+      if (param.protocol.http!.in !== undefined && param.protocol.http!.in === 'body') {
+        bodyParam = param.language.go!.name;
+      }
+    }
+    text += `\treq.Header.Set("Content-Type", "text/plain; encoding=UTF-8")\n`;
```
jhendrixMSFT, line 598: You can remove this line, the `content-type` is set in the call to `req.SetBody()`.

test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
```diff
@@ -0,0 +1,195 @@
...
+
+// AnalyzeBodyCreateRequest creates the AnalyzeBody request.
+func (client *MediaTypesClient) AnalyzeBodyCreateRequest(ctx context.Context, contentType ContentType, input azcore.ReadSeekCloser) (*azcore.Request, error) {
+	urlPath := "/mediatypes/analyze"
+	req, err := azcore.NewRequest(ctx, http.MethodPost, azcore.JoinPaths(client.u, urlPath))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", string(contentType))
+	req.Header.Set("Accept", "application/json")
```
jhendrixMSFT, line 69: You might want to follow up with David about this to see if it's expected to add this header in the case where we are specifying a non-JSON payload.

src/generator/operations.ts
catalinaperalta: I had it without explicitly setting the header originally but it seems there could be a problem in SetBody because the test fails when reading the content-type header without it being explicitly set like this

catalinaperalta: I'll test it out once more without it to see, there's probably something that needs to be fixed in SetBody, I havent looked at it yet

jhendrixMSFT: We should fix `SetBody()` then (the whole point of that change was to set the content-type there).

test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
catalinaperalta: Will do, then add the update in another PR whenever he does get the fix in, if it will be fixed.

src/generator/operations.ts
jhendrixMSFT: It's because you're sending a string of zero length, so the header (and body) is omitted.  Does the test expect a zero-length string?

catalinaperalta: https://github.com/Azure/autorest.testserver/blob/faa78f54f3e42d51dd001cb2644c2139244114a1/legacy/routes/mediatypes.js#L32

I see. The test doesnt care, I'll send some random string then for the value.

jhendrixMSFT: You can send any length string.

src/generator/operations.ts
catalinaperalta: done

src/generator/operations.ts

src/generator/operations.ts
catalinaperalta: done

test/autorest/mediatypesgroup/zz_generated_mediatypesclient.go
jhendrixMSFT: Thinking more, this is likely an artifact of us making a copy of the original method.  If it doesn't hurt anything we can leave it as-is.

jhendrixMSFT (approved):

PR #481, opened 2020-09-11
Updated codegen based on azcore refactor

jhendrixMSFT:
Removed context parameter from pipeline Do() method.
Request creation method now takes a context.
Added content type param to SetBody() calls for binary media types.
Pagers now use a closure to create the initial request in the call to
NextPage().  As a result, the pager APIs no longer return an error.
Removed some unused codegen in arm_pollers.ts.


catalinaperalta (approved):
LGTM

PR #482, opened 2020-09-11
Support header collections in responses

jhendrixMSFT:
Add support for x-ms-header-collection-prefix on header response types.
This required updating M4 to properly support which also introduced the
header 'Accept' which specifies the content type.


catalinaperalta (approved):
LGTM

PR #483, opened 2020-09-14
Call error handler outside of response handler

jhendrixMSFT:
Moved call to error handler out of the response handler and into the API
body.  The method now returns nil on success.
Added errorer field to pagers to be called for checking response, along
with errHandler to pager-pollers.
Omitted response handler for methods that don't return a model.
Added missing error check in pagers NextPage() method.
Function generateResponseUnmarshaller() was moved out of
createProtocolResponse(), its implementation did not change.


jhendrixMSFT:
Sigh I forgot this was incomplete.  We need to call the error handler from `armpoller`.  Investigating.

catalinaperalta (approved):
LGTM

PR #484, opened 2020-09-15
Change the way that we recognize a package as preview

ArcturusZhang:
This PR makes those tags that contains at least one preview swagger as a preview package.

Previously, a preview package is forced when all of the swaggers within a tag are preview.

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -193,7 +193,16 @@ namespace AutoRest.Go
         private static bool IsPreviewPackage(string[] files)
         {
             // from the breaking change perspective, we should regard a composite package as a preview package when at least one of the swagger is preview, since preview swagger files may receive breaking changes frequently
-            return files.Any(file => file.IndexOf(previewSubDir) >= 0);
+            return files.Any(IsPreviewSwagger);
+        }
+
+        private static bool IsPreviewSwagger(string file)
+        {
+            // a preview swagger should have the '/preview/' segment in the middle of its path, or start with preview/ (this happens in some data plane swaggers)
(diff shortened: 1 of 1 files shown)


tombuildsstuff:
Out of interest will this also catch services which release preview versions in stable API's? Using Networking & AKS as examples

ArcturusZhang:
This will ensure that if a go package is generated from several swaggers and there is at least one swagger is preview, the go package will be marked as preview.

Previously we do not check this unless a set of swaggers are all preview.

jhendrixMSFT:
Have you tried building the SDK with this change?  Do any packages now fail to build?

jhendrixMSFT (approved):

ArcturusZhang:
Hi @jhendrixMSFT I am running a full scan of the swagger repo to see how many readme files need modifying. I will post the result here soon

jhendrixMSFT:
Given our reliance on this and how much we're iterating on it I think it would be wise to add tests to cover this scenario in CI.

ArcturusZhang:
Hi @jhendrixMSFT I open a PR to fix the bad output-folders: https://github.com/Azure/azure-rest-api-specs/pull/10842, you could get the list in the file changes.

ArcturusZhang:
> Given our reliance on this and how much we're iterating on it I think it would be wise to add tests to cover this scenario in CI.

Could you please be more specific on this?

jhendrixMSFT:
Unit tests for the detection of preview packages.  I assume it's something run manually, likely building the entire SDK.  Would be great to have something that runs as part of CI (can be a separate PR).

fearthecowboy:
## success (version: 2.1.157)

Issue #294, opened 2020-03-10
Check sealed-choices that only have 1 choice


catalinaperalta:
Sealed-choices that only have one choice could be treated as a constant. This is currently not taken into consideration, so tests need to be carried out to see if many swaggers include sealed-choices with only 1 choice.

jhendrixMSFT:
Fixed.

PR #485, opened 2020-09-15
Include non-required header and query param constants

jhendrixMSFT:
Storage, and other services, contain optional header and query params
that are constants so include them in codegen.
Replaced 'TODO' type descriptions with actual descriptions.  They aren't
actually used but cleans up noise when searching for TODO comments.

### DIFF
--- a/test/storage/2019-07-07/azblob/zz_generated_models.go
+++ b/test/storage/2019-07-07/azblob/zz_generated_models.go
@@ -3214,6 +3214,8 @@ type ServiceGetUserDelegationKeyOptions struct {
 
 // ServiceListContainersSegmentOptions contains the optional parameters for the Service.ListContainersSegment method.
 type ServiceListContainersSegmentOptions struct {
+	// Include this parameter to specify that the container's metadata be returned as part of the response body.
+	Include *string
 	// A string value that identifies the portion of the list of containers to be returned with the next listing operation. The
 	// operation returns the NextMarker value within the response body if the listing operation did not return all containers
 	// remaining to be listed with the current page. The NextMarker value can be used as the value for the marker parameter in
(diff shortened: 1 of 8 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/294

test/storage/2019-07-07/azblob/zz_generated_models.go
```diff
@@ -3214,6 +3214,8 @@ type ServiceGetUserDelegationKeyOptions struct {
 
 // ServiceListContainersSegmentOptions contains the optional parameters for the Service.ListContainersSegment method.
 type ServiceListContainersSegmentOptions struct {
+	// Include this parameter to specify that the container's metadata be returned as part of the response body.
+	Include *string
```
catalinaperalta, line 3218: This is weird, is this supposed to be a boolean? Just asking because of the comment.

catalinaperalta: Nevermind, I see where it is used. I still think it's strange that it's not a boolean.

catalinaperalta (approved):
LGTM

test/storage/2019-07-07/azblob/zz_generated_models.go
jhendrixMSFT: I agree, and originally my thought was to change these to `bool` as that's how they behave.  However I wanted to get some feedback on how it currently works from the storage team.

catalinaperalta: Well, if it's how it's defined in the swagger, I think it's the storage team who should decide to change this or not in the convenience layer. At the moment we arent adding any more magic for it, which I think is good.

Issue #292, opened 2020-03-09
Check nested headers in swaggers


catalinaperalta:

jhendrixMSFT:
This is for the case where a response envelope is shared across operations.  Ensure that all the headers are present in the response envelope.

jhendrixMSFT:
Fixed.

PR #486, opened 2020-09-15
Include response headers from all operations

jhendrixMSFT:
For shared response envelopes, ensure that header responses from all
operations are included in the envelope.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/292

jhendrixMSFT:
Note that we don't have any swaggers that model this at present so I had to fabricate a local test to verify the fix.

catalinaperalta (approved):
LGTM

PR #487, opened 2020-09-16
Fix codegen for optional constant param

jhendrixMSFT:
Optional param group that contained only a header/query constant were
being dropped due to incorrect ordering of check.  This was found when
generating compute.
Add compute, synapse artifacts and spark to CI.
Add missing go.sum files.


catalinaperalta (approved):
LGTM

PR #488, opened 2020-09-16
Output initial request helper method for LROs

catalinaperalta:
Add a wrapper similar to the other non-LRO operations for the LRO endpoints.


catalinaperalta:
NOTE: response handlers and final response updates will be made in a separate PR

src/transform/namer.ts
```diff
@@ -103,7 +107,7 @@ export async function namer(session: Session<CodeModel>) {
       details.name = getEscapedReservedName(capitalizeAcronyms(pascalCase(details.name)), 'Method');
       // add the client name to the operation as it's needed all over the place
       details.clientName = groupDetails.clientName;
-      if (isLROOperation(op)) {
+      if (isLROOperation(op) && !isDataPlane) {
```
jhendrixMSFT, line 110: I don't think we should be adding the prefix, it just adds complexity.  Instead, for non-data-plane we create the additional `Begin*` method.

jhendrixMSFT:
I think you'll want to rebase on https://github.com/Azure/autorest.go/pull/487 once it's merged so that we can see the effects on data-plane pollers.

src/transform/namer.ts
catalinaperalta: Changed it, although I'm not sure I like the lingering MethodPrefix. I could remove it completely and we can add it back in if necessary

jhendrixMSFT:
FYI I rebased and regenerated.

src/transform/namer.ts
jhendrixMSFT: I went ahead and removed `methodPrefix`

jhendrixMSFT (approved):

PR #489, opened 2020-09-23
Add support for azcore.HTTPResponse interface

jhendrixMSFT:
Wrap API errors in azcore.NewResponseError() so callers have access to
the raw HTTP response.
Return error with schemas as pointer-to-type.
Remove azcore.RequestLogOptions per azcore breaking changes.


catalinaperalta (approved):
LGTM

PR #490, opened 2020-09-23
Add optional params placeholder struct to operations

jhendrixMSFT:
Always add an optional params struct to an operation even if it doesn't
contain any optional params.  This is to prevent future addition of
optional params from being a breaking change.
Use the param name 'options' for singular, optional params arg.


catalinaperalta (approved):
LGTM

PR #494, opened 2020-09-29
Remove hard-coded status codes for NextPage()

jhendrixMSFT:
Status codes for a successful page retrieval might vary depending on the
operation, so use the per-operation values.
Renamed a field in PagerInfo to clarify its usage.


catalinaperalta (approved):
LGTM

PR #496, opened 2020-09-30
Decouple PollerInfo from operation

jhendrixMSFT:
Removed Operation from PollerInfo, adding required fields.
Omitted support for pager-poller that uses a Next() operation.  No
services use the pattern and it's really complicated to implement.
Removed some dead code in arm_pollers.ts

### DIFF
--- a/src/transform/transform.ts
+++ b/src/transform/transform.ts
@@ -697,6 +697,11 @@ function createResponseEnvelope(codeModel: CodeModel, group: OperationGroup, op:
     }
     // create poller type info
     if (isLROOperation(op)) {
+      if (op.language.go!.paging && op.language.go!.paging.member) {
+        // implementing support for this is very complicated, and since
+        // no services at present use this pattern avoid it for now
+        throw console.error(`${op.language.go!.name}: unsupported pager-poller that uses next page operation`);
+      }
       // create the poller response envelope
       generateLROResponseEnvelope(response, op, codeModel);
       if (codeModel.language.go!.pollerTypes === undefined) {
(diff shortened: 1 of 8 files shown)


src/transform/transform.ts
```diff
@@ -697,6 +697,11 @@ function createResponseEnvelope(codeModel: CodeModel, group: OperationGroup, op:
     }
     // create poller type info
     if (isLROOperation(op)) {
+      if (op.language.go!.paging && op.language.go!.paging.member) {
+        // implementing support for this is very complicated, and since
+        // no services at present use this pattern avoid it for now
+        throw console.error(`${op.language.go!.name}: unsupported pager-poller that uses next page operation`);
```
catalinaperalta, line 703: So just making sure I understand, we wont support pager pollers that have the next operation? There's no current use case for this?

jhendrixMSFT: Correct.  I searched through the specs repo, the only specs I found that even use a next operation are for graphrbac.  Fortunately none of them are LROs.

jhendrixMSFT: And to elaborate on the problem.

Suppose you have two operations that return a `FooPagerPollerResponse`.  If each operation has its own `NextFoo()` operation with a discrete set of parameters, how do we carry/populate such parameters inside the pager-poller?

catalinaperalta: Ok, yes that does make it more difficult. Sounds good then, let's deal with it when it comes up.

catalinaperalta (approved):
LGTM

PR #497, opened 2020-10-01
Add additional comments for polymorphic types

jhendrixMSFT:
Added comment indicating possible concrete types for a discriminator
interface type.
Added comment for discriminator interface method.

### DIFF
--- a/test/synapse/2019-06-01/azartifacts/zz_generated_models.go
+++ b/test/synapse/2019-06-01/azartifacts/zz_generated_models.go
@@ -14,7 +14,18 @@ import (
 )
 
 // ActivityClassification provides polymorphic access to related types.
+// Call the interface's GetActivity() method to access the common type.
+// Use a type switch to determine the concrete type.  The possible types are:
+// - *Activity, *AppendVariableActivity, *AzureDataExplorerCommandActivity, *AzureFunctionActivity, *AzureMLBatchExecutionActivity,
+// - *AzureMLExecutePipelineActivity, *AzureMLUpdateResourceActivity, *ControlActivity, *CopyActivity, *CustomActivity, *DataLakeAnalyticsUsqlActivity,
+// - *DatabricksNotebookActivity, *DatabricksSparkJarActivity, *DatabricksSparkPythonActivity, *DeleteActivity, *ExecuteDataFlowActivity,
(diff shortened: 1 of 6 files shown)


test/synapse/2019-06-01/azartifacts/zz_generated_models.go
```diff
@@ -17582,7 +17688,30 @@ func (l *LinkedIntegrationRuntimeType) unmarshalInternal(rawMsg map[string]*json
 }
 
 // LinkedServiceClassification provides polymorphic access to related types.
+// Call the interface's GetLinkedService() method to access the common type.
```
jhendrixMSFT, line 17691: @JeffreyRichter comment improvements.  Note the large number of possible types.  I think having one method per possible type would be more harmful than useful.

catalinaperalta (approved):
LGTM

test/synapse/2019-06-01/azartifacts/zz_generated_models.go
catalinaperalta: There really are a lot of types. I do still wish there was something more to help with discoverability. Although, I do agree it could be overwhelming to see all of these methods and search through them, so maybe this one consolidated comment is enough. Customer feedback here would be great.

jhendrixMSFT: Agreed on customer feedback.  I think that should be the next step before we contemplate changing the design.

jhendrixMSFT: You can see how gnarly the method-based approach gets in track 1.

https://raw.githubusercontent.com/Azure/azure-sdk-for-go/090dc0ee4d8d2d60e2a9525774d967a4111a2b0c/services/datafactory/mgmt/2018-06-01/datafactory/models.go

Search for "func (amod AmazonMWSObjectDataset) As", there are 89(!) methods, and these are just simple one-liners.  Ugh.

catalinaperalta: It seems overwhelming to search through all 89 until you find the one you want. I would definitely like to see someone in a user study use a similarly large type and see what problems they run into or how they reason about this.

PR #500, opened 2020-10-08
Update codegen based on latest azcore

jhendrixMSFT:
Fixed logging policy name.
Fixed default telemetry params.
Removed unique request ID policy.


catalinaperalta (approved):
LGTM

Issue #498, opened 2020-10-06
Add support for --head-as-boolean autorest switch


jhendrixMSFT:
For HEAD, we'd generate something like this.

```go
type BoolResponse struct {
  RawResponse *http.Response
  Succeeded bool
}
```

For 2xx status codes, `Succeeded` would be `true`.
For 4xx status codes, `Succeeded` would be `false`.
For 5xx status codes return an error.

jhendrixMSFT:
Per internal discussion, we're leaning towards supporting this via a swagger extension, e.g. `x-ms-boolean-response` or something similar.

jhendrixMSFT:
Done

PR #502, opened 2020-10-20
Add support for head-as-boolean

jhendrixMSFT:
When enabled, all HEAD operations will return a Success field in their
response envelope.  For HTTP status >= 200 and < 300 Success will be set
to true.
For operations that don't model header responses, return a
BooleanResponse response envelope.
Update codegen for latest azcore and armcore.
Consolidated some checks into a single byValue flag.
Removed some duplicated code.
Updated autorest.testserver to latest version.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/498

src/generator/operations.ts
```diff
@@ -264,20 +264,32 @@ function generateOperation(op: Operation, imports: ImportManager): string {
...
-    text += '\t return resp, nil\n';
-  } else if (needsResponseHandler(op)) {
-    // also cheating here as at present the only param to the responder is an azcore.Response
-    text += `\tresult, err := client.${info.protocolNaming.responseMethod}(resp)\n`;
-    text += `\tif err != nil {\n`;
-    text += `\t\treturn nil, err\n`;
-    text += `\t}\n`;
-    text += `\treturn result, nil\n`;
+  // HAB with schema response is handled in protocol responder
+  if (op.language.go!.headAsBoolean && !(op.responses && isSchemaResponse(op.responses[0]))) {
```
catalinaperalta, line 268: This case will always be for operations that dont have responses? Or for operations that dont have schema responses? Just checking, we dont need to include it for those ever?

jhendrixMSFT: It will be for the case where HAB is enabled and there's no schema response.

This is a corner case.  By definition, an HTTP HEAD request will never return a response body (i.e. it's never a schema response); just an HTTP status code and possibly header values (modeling a response body for HEAD is an error).  For operation responses that don't model a response body but *do* model response headers, we create a schema response modeling the header values.  So we have this case where a HEAD operation becomes a schema response to model the response headers.

If an operation models only response headers, it will return a `<Group><Method>Response` response envelope.  In the case of a HEAD request with HAB enabled, we must use the operation-specific response envelope, not the generic `BooleanResponse` response envelope.

catalinaperalta: Got it, thank you for the explanation! :)

catalinaperalta (approved):
LGTM

PR #510, opened 2020-10-28
Add npm publishing pipeline

daviwil:
This basically copies the `publish-dev-release.yml` that you've already defined to add an `npm` publishing step.  Once this looks good I can create an Azure DevOps pipeline definition that uses it so that you can publish your first release.

### DIFF
--- a/eng/pipelines/publish-release.yml
+++ b/eng/pipelines/publish-release.yml
@@ -39,7 +39,7 @@ steps:
 
   - task: GoTool@0
     inputs:
-      version: '1.14'
+      version: '1.15.3'
     displayName: "Select Go Version"
 
   - script: |
--- a/eng/pipelines/publish-release.yml
+++ b/eng/pipelines/publish-release.yml
@@ -1,0 +1,107 @@
+trigger: none
+pr: none
+
+variables:
+  NodeVersion: '12.x'
+  goTestPath: '$(system.defaultWorkingDirectory)/test'
+  GOBIN: '$(system.defaultWorkingDirectory)/bin/go'
+
+pool:
+  vmImage: 'ubuntu-18.04'
+
+variables:
+  NodeVersion: '12.x'
+  TestFolder: '$(Build.SourcesDirectory)/test/'
+
+steps:
+  - task: NodeTool@0
+    displayName: 'Install Node $(NodeVersion)'
+    inputs:
+      versionSpec: '$(NodeVersion)'
+
+  - script: |
+      cd $(Build.SourcesDirectory)
+      npm install -g "@microsoft/rush"
+      rush update --debug
+      npm install -g autorest
+    displayName: 'Prepare Generator Environment'
+
+  - script: |
+      rush rebuild -v
+    displayName: 'Build Generator Sources'
+
+  - script: |
+      rush regenerate
+      git add -A ./test/.
+      git diff --staged -w 1>&2
+    displayName: 'Regenerate Autorest Tests'
+    failOnStderr: true
+
+  - task: GoTool@0
+    inputs:
+      version: '1.14'
+    displayName: "Select Go Version"
+
+  - script: |
+      set -e
+      go version
+      go get github.com/jstemmer/go-junit-report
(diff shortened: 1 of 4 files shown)


eng/pipelines/publish-release.yml
```diff
@@ -0,0 +1,143 @@
...
+  NodeVersion: '12.x'
+  TestFolder: '$(Build.SourcesDirectory)/test/'
+
+steps:
+  - task: NodeTool@0
+    displayName: 'Install Node.js $(NodeVersion)'
+    inputs:
+      versionSpec: '$(NodeVersion)'
+
+  - task: UsePythonVersion@0
```
jhendrixMSFT, line 22: I assume all the Python stuff can be removed yes?  Lines 22 through 51.

daviwil: Ouch, didn't mean to leave that in there.  Thanks for catching it!

daviwil:
Any idea what the CI failures are here?

jhendrixMSFT:
It looks like the generated tests are not current.  Is your fork current?

daviwil:
Should be, I just forked it yesterday.  I'll rebase and see if that fixes it.

daviwil:
Yep, apparently I'm up to date with the `track2` branch.  Is that the wrong one perhaps?

jhendrixMSFT:
No it's correct.  There are some packages we build from live links to the specs repo that aren't pinned to a specific commit.  Looks like those swaggers have been updated.  I will regenerate and push to your PR.

jhendrixMSFT:
Looks like this will publish a package version with the same format as our [dev releases](https://github.com/Azure/autorest.go/releases).

E.g. 4.0.0-20201029.01

Not sure we want this format.  Something like `4.0.0-preview-1` might be better, and we increment the `-N` suffix.  Once we hit GA we'd drop the suffix altogether.

What are other generators doing here?

eng/pipelines/publish-release.yml
```diff
@@ -0,0 +1,107 @@
...
+      testResultsFiles: $(goTestPath)/**/report.xml
+      failTaskOnFailedTests: true
+
+  - pwsh: |
+      $currentVersion = node -p -e "require('./src/package.json').version";
+      $currentVersion="$currentVersion-$(Build.BuildNumber)";
+      cd src
+      npm version --no-git-tag-version $currentVersion
+      npm pack;
+      npx publish-release --token $(package-write-token) --repo autorest.go --owner azure --name "Autorest for Go v$currentVersion" --tag v$currentVersion --notes='Preview version of Autorest for Go track 2' --prerelease --editRelease false --assets autorest-go-$currentVersion.tgz --target_commitish $(Build.SourceBranchName);
```
jhendrixMSFT, line 101: @daviwil what does the `--prerelease` switch do here, does it impact the package name in any way?

daviwil: Nope, it just marks the GitHub release itself as a prerelease, doesn't have any effect on the package itself.

daviwil:
> Looks like this will publish a package version with the same format as our [dev releases](https://github.com/Azure/autorest.go/releases).
> 
> E.g. 4.0.0-20201029.01
> 
> Not sure we want this format. Something like `4.0.0-preview-1` might be better, and we increment the `-N` suffix. Once we hit GA we'd drop the suffix altogether.
> 
> What are other generators doing here?

Yep, I missed that part, I'll update the code to use whatever version is stored in `package.json`.  When publishing new updates to `npm` it's better if we manage the version number in source control so that things do what we expect.

jhendrixMSFT:
@daviwil I've made this change based on what Python is doing and rebased on track2.

daviwil:
Sounds good!

jhendrixMSFT (approved):

jhendrixMSFT:
@daviwil thanks for your help on this.  Let me know when the pipeline is ready and I'll queue a build.

daviwil:
@jhendrixMSFT Here's the pipeline: https://dev.azure.com/azure-sdk/internal/_build?definitionId=2214

PR #511, opened 2020-10-28
Adding markdown comment support

### LINKED ISSUE #505: Add support for Markdown comments defined in swaggers
Example of what it looks like currently if we receive a comment defined in markdown: https://godoc.org/github.com/Azure/azure-sdk-for-go/services/cognitiveservices/v1.0/face#LargeFaceListClient.List

catalinaperalta:
Adding function to parse markdown comments to html and then to plain text. Fixes #505

### DIFF
--- a/src/generator/enums.ts
+++ b/src/generator/enums.ts
@@ -4,10 +4,11 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { Session } from '@azure-tools/autorest-extension-base';
-import { comment, joinComma } from '@azure-tools/codegen';
+import { comment } from '@azure-tools/codegen';
 import { CodeModel, ChoiceValue, Schemas } from '@azure-tools/codemodel';
 import { values } from '@azure-tools/linq';
 import { contentPreamble, hasDescription, sortAscending } from './helpers';
+import { commentLength } from '../common/helpers';
 
 // Creates the content in enums.go
 export async function generateEnums(session: Session<CodeModel>): Promise<string> {
@@ -19,14 +20,14 @@ export async function generateEnums(session: Session<CodeModel>): Promise<string
   let text = await contentPreamble(session);
   for (const enm of values(enums)) {
     if (enm.desc) {
-      text += `${comment(`${enm.name} - ${enm.desc}`, '// ')}\n`;
+      text += `${comment(`${enm.name} - ${enm.desc}`, '// ', undefined, commentLength)}\n`;
     }
     text += `type ${enm.name} ${enm.type}\n\n`;
     const vals = new Array<string>();
     text += 'const (\n';
     for (const val of values(enm.choices)) {
       if (hasDescription(val.language.go!)) {
-        text += `${comment(val.language.go!.description, "//", undefined, 150)}\n`;
+        text += `\t${comment(`${val.language.go!.name} - ${val.language.go!.description}`, "//", undefined, commentLength)}\n`;
       }
       let formatValue = `"${val.value}"`;
       if (enm.type !== 'string') {
--- a/src/generator/enums.ts
+++ b/src/generator/enums.ts
@@ -19,14 +19,14 @@ export async function generateEnums(session: Session<CodeModel>): Promise<string
   let text = await contentPreamble(session);
   for (const enm of values(enums)) {
     if (enm.desc) {
-      text += `${comment(enm.name, '// ')} - ${enm.desc}\n`;
+      text += `${comment(`${enm.name} - ${enm.desc}`, '// ')}\n`;
     }
     text += `type ${enm.name} ${enm.type}\n\n`;
     const vals = new Array<string>();
     text += 'const (\n';
     for (const val of values(enm.choices)) {
       if (hasDescription(val.language.go!)) {
-        text += comment(val.language.go!.description);
+        text += `${comment(val.language.go!.description, "//", undefined, 150)}\n`;
       }
       let formatValue = `"${val.value}"`;
       if (enm.type !== 'string') {
--- a/src/generator/enums.ts
+++ b/src/generator/enums.ts
@@ -26,7 +26,7 @@ export async function generateEnums(session: Session<CodeModel>): Promise<string
     text += 'const (\n';
     for (const val of values(enm.choices)) {
       if (hasDescription(val.language.go!)) {
-        text += `\t${comment(val.language.go!.name, '// ')} - ${val.language.go!.description}\n`;
+        text += comment(val.language.go!.description);
       }
       let formatValue = `"${val.value}"`;
       if (enm.type !== 'string') {
--- a/src/transform/transform.ts
+++ b/src/transform/transform.ts
@@ -41,14 +41,6 @@ async function process(session: Session<CodeModel>) {
   session.model.language.go!.headAsBoolean = headAsBoolean;
   processOperationRequests(session);
   processOperationResponses(session);
-  // parse operation comments that are in Markdown to convert them to plain text
-  for (const group of values(session.model.operationGroups)) {
-    for (const op of values(group.operations)) {
-      if (op.language.go!.description) {
-        op.language.go!.description = parseComments(op.language.go!.description);
-      }
-    }
-  }
   // fix up dictionary element types (additional properties)
   // this must happen before processing objects as we depend on the
   // schema type being an actual Go type.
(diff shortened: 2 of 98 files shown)


src/transform/transform.ts
```diff
@@ -100,9 +116,15 @@ async function process(session: Session<CodeModel>) {
   // fix up enum types
   for (const choice of values(session.model.schemas.choices)) {
     choice.choiceType.language.go!.name = schemaTypeToGoType(session.model, choice.choiceType, false);
+    // if (choice.language.go!.description) {
+    //   choice.language.go!.description = parseComments(choice.language.go!.description);
+    // }
```
catalinaperalta, line 121: This is broken for compute, I'm investigating why the comments arent formatting correctly.

src/generator/enums.ts
```diff
@@ -26,7 +26,7 @@ export async function generateEnums(session: Session<CodeModel>): Promise<string
     text += 'const (\n';
     for (const val of values(enm.choices)) {
       if (hasDescription(val.language.go!)) {
-        text += `\t${comment(val.language.go!.name, '// ')} - ${val.language.go!.description}\n`;
+        text += comment(val.language.go!.description);
```
catalinaperalta, line 29: Will change this once I fix the problem with the compute enums

test/compute/2019-12-01/armcompute/zz_generated_virtualmachines.go
```diff
@@ -21,45 +21,61 @@ import (
...
 	ResumeCapture(token string) (VirtualMachineCaptureResultPoller, error)
-	// BeginConvertToManagedDisks - Converts virtual machine disks from blob-based to managed disks. Virtual machine must be stop-deallocated before invoking this operation. <br>For Windows, please refer to [Convert a virtual machine from unmanaged disks to managed disks.](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/convert-unmanaged-to-managed-disks).<br>For Linux, please refer to [Convert a virtual machine from unmanaged disks to managed disks.](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/convert-unmanaged-to-managed-disks).
+	// BeginConvertToManagedDisks - Converts virtual machine disks from blob-based to managed disks. Virtual machine must be stop-deallocated
+	// before invoking this operation. For Windows, please refer to Convert a virtual machine from
+	// unmanaged disks to managed disks. [https://docs.microsoft.com/en-us/azure/virtual-machines/windows/convert-unmanaged-to-managed-disks
+	// [https://docs.microsoft.com/en-us/azure/virtual-machines/windows/convert-unmanaged-to-managed-disks]]. For Linux, please
```
catalinaperalta, line 32: This is not ideal, I think this is because of the module so I might switch to another, although it does format other html well.

catalinaperalta: Considering switching to the showdown module although the maintainer says he has ben unable to dedicate time to it, but hope to have more time in a couple of months.

test/autorest/arraygroup/zz_generated_array.go
```diff
@@ -24,7 +24,8 @@ type ArrayOperations interface {
 	GetArrayNull(ctx context.Context, options *ArrayGetArrayNullOptions) (*StringArrayArrayResponse, error)
 	// GetArrayValid - Get an array of array of strings [['1', '2', '3'], ['4', '5', '6'], ['7', '8', '9']]
 	GetArrayValid(ctx context.Context, options *ArrayGetArrayValidOptions) (*StringArrayArrayResponse, error)
-	// GetBase64URL - Get array value ['a string that gets encoded with base64url', 'test string' 'Lorem ipsum'] with the items base64url encoded
+	// GetBase64URL - Get array value ['a string that gets encoded with base64url', 'test string' 'Lorem ipsum'] with the items
```
jhendrixMSFT, line 27: I think this truncation happened due to the use of `comment()`.  I now recall that in track 1 we had an issue with `comment()` being somewhat non-deterministic and would sometimes change where the line-break would occur.  So you'd get a diff in the codegen even if the input didn't change.  I think we can fix this via the `factor` and/or `maxLength` params.

test/compute/2019-12-01/armcompute/zz_generated_dedicatedhostgroups.go
```diff
@@ -18,9 +18,8 @@ import (
 
 // DedicatedHostGroupsOperations contains the methods for the DedicatedHostGroups group.
 type DedicatedHostGroupsOperations interface {
-	// CreateOrUpdate - Create or update a dedicated host group. For details of Dedicated Host and Dedicated Host Groups please see [Dedicated Host Documentation]
-	// (https://go.microsoft.com/fwlink/?linkid=2082596
-	// [https://go.microsoft.com/fwlink/?linkid=2082596][https://go.microsoft.com/fwlink/?linkid=2082596 [https://go.microsoft.com/fwlink/?linkid=2082596]])
+	// CreateOrUpdate - Create or update a dedicated host group. For details of Dedicated Host and Dedicated Host Groups please see Dedicated Host Documentation
+	// [https://go.microsoft.com/fwlink/?linkid=2082596]
```
jhendrixMSFT, line 22: Are the `[]` supposed to be there?  I don't think it's a big deal, just curious.

test/network/2020-03-01/armnetwork/zz_generated_models.go
```diff
@@ -13470,8 +13469,8 @@ type VirtualNetworkPeeringPropertiesFormat struct {
 	RemoteAddressSpace *AddressSpace `json:"remoteAddressSpace,omitempty"`
 
 	// The reference to the remote virtual network. The remote virtual network can be in the same or different region (preview). See here to register for the
-	// preview and learn more (
-	// https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-create-peering [https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-create-peering]).
+	// preview and learn more
+	// (https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-create-peering).
```
jhendrixMSFT, line 13473: Similar question about surrounding `()`.

src/transform/transform.ts
```diff
@@ -39,14 +41,28 @@ async function process(session: Session<CodeModel>) {
   session.model.language.go!.headAsBoolean = headAsBoolean;
   processOperationRequests(session);
   processOperationResponses(session);
+  // parse operation comments that are in Markdown to convert them to plain text
+  for (const group of values(session.model.operationGroups)) {
```
jhendrixMSFT, line 45: Seems like this could be folded into `processOperationRequests()`.

As an aside, we could probably consolidate `processOperationRequests()` and `processOperationResponses()` to only loop once.

src/transform/transform.ts
```diff
@@ -1057,3 +1088,14 @@ function getEnumForDiscriminatorValue(discValue: string, enums: Array<ChoiceValu
   }
   throw console.error(`failed to find discriminator enum value for ${discValue}`);
 }
+
+// convert comments that are in Markdown to html and then to plain text
+export function parseComments(comment: string): string {
```
jhendrixMSFT, line 1093: Does this actually need to be exported?  Looks like it's used locally.

jhendrixMSFT (approved):
A few minor things that could be tweaked, but looks good otherwise.

test/compute/2019-12-01/armcompute/zz_generated_dedicatedhostgroups.go
catalinaperalta: Yes that's how the links are defined if they were specified as something like `[link name](https://link.com)`

test/network/2020-03-01/armnetwork/zz_generated_models.go
catalinaperalta: That's coming from the swagger

src/transform/transform.ts
catalinaperalta: Done. Parsing the comments on operations in `processOperationRequests`. About joining the processing of operation requests and responses, that's a good improvement. I could do it in a future PR.

src/transform/transform.ts
catalinaperalta: Nope, I removed the export.

PR #520, opened 2020-11-06
Switch uuid library to gofrs/uuid

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/3158


jhendrixMSFT:
You will need to run `gulp updatetestdep` to update the test dependencies with the new package.

ArcturusZhang:
Hi @jhendrixMSFT I just updated the dependencies. But please hold on merging, I am working to see how many packages are impacted and if not so many, we could regenerate all of them in the next major version.

I am opening a PR in go SDK to see whether the APIs in these two packages are identical

test/src/tests/Gopkg.lock
```diff
@@ -27,12 +27,12 @@
   version = "v3.2.0"
 
 [[projects]]
-  digest = "1:4cc7748b74453830951f95ddc3fd980387e7c13f77cc33d7cba54cd6b96dad26"
-  name = "github.com/kr/pretty"
+  digest = "1:73770491417b7cb646b28610eec6481187aba7b5c3703da0d42f8d590d016f5f"
+  name = "github.com/gofrs/uuid"
```
jhendrixMSFT, line 31: Hmm looks like `gopkg.toml` wasn't updated.  You might need to update it by hand and rerun `gulp updatetestdep`

ArcturusZhang: I have updated the toml, please have a look, thanks

jhendrixMSFT (approved):

jhendrixMSFT:
@ArcturusZhang when do you want to release with this change?  Also note the merge conflicts.

ArcturusZhang:
> @ArcturusZhang when do you want to release with this change? Also note the merge conflicts.

it may be better if we only have one big series of breaking changes in one major version.
Considering we just made the future mocking change, could we postpone this to the next next major version after we regenerate everything to pick up the future mocking fix?

ArcturusZhang:
Hi @jhendrixMSFT could you please merge this? We would like to release this change in this week's release.

Porges:
Thanks!

PR #521, opened 2020-11-06
Rename Client to Connection

jhendrixMSFT:
Removed operation group client method Do() and replaced it with
Pipeline() that returns the associated connection's pipeline.
Add connection methods for parameterized host values.


catalinaperalta (approved):
LGTM

PR #522, opened 2020-11-09
Add armcore-connection argument

jhendrixMSFT:
Specify "--armcore-connection=true" to use the Connection type in
armcore instead of generating one.


test/compute/2019-12-01/armcompute/zz_generated_connection.go
```diff
@@ -5,89 +5,4 @@
 
 package armcompute
 
-import (
-	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/armcore"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-)
-
-const scope = "https://management.azure.com//.default"
```
catalinaperalta, line 14: Did something break? The content in this file was removed.

jhendrixMSFT: This is expected, the only content in the file is the telemetry `const` (it probably got lost in the sea of red).

catalinaperalta (approved):
LGTM

PR #526, opened 2020-11-16
Fix constraint type issue

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/13604

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -538,23 +538,23 @@ namespace AutoRest.Go
             {
                 x.AddRange(p.Constraints
                     .Where(c => c.IsValidConstraint())
-                    .Select(c => GetConstraint(name, c.Key.ToString(), c.Value)).ToList());
+                    .Select(c => GetConstraint(name, p.ModelTypeName, c.Key.ToString(), c.Value)).ToList());
             }
 
             List<string> y = new List<string>();
             if (x.Count > 0)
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
Hi @jhendrixMSFT could you please take a look at this PR?

src/Extensions.cs
```diff
@@ -717,13 +717,17 @@ public static string GetConstraint(string name, string constraintName, string co
...
                 {
-                    value = $"float64({constraintValue})";
-                }
-                else
-                {
-                    value = $"int64({constraintValue})";
+                    case "float64":
+                        value = $"float64({constraintValue})";
+                        break;
+                    case "int32":
```
jhendrixMSFT, line 725: If you're adding `int32` you might as well add `float32` above.

ArcturusZhang: Well, in [this file](https://github.com/Azure/autorest.go/blob/cd02b097dc9a7ba3d77cdbc466882c7ceb2bc91c/src/Model/PrimaryTypeGo.cs#L60) we do not support `float32` in the SDK, therefore I suppose maybe we could also omit this one?

jhendrixMSFT: Good point, we can omit it here.

jhendrixMSFT (approved):

PR #529, opened 2020-11-18
Introduce a new constructor to avoid breaking change

ArcturusZhang:
No quite sure if the name of new constructor is appropriate.

And please do not merge this since we are still hesitating between this change or a full regenerate.


ArcturusZhang:
Close this since we are regenerate everything in the next major version of the SDK

PR #530, opened 2020-11-19
Remove operations interfaces from codegen

jhendrixMSFT:
Return clients by value and make their receivers by value.
Add method descriptions for Begin and Resume LRO APIs.


catalinaperalta (approved):
LGTM

PR #532, opened 2020-11-20
Return response envelopes by value

jhendrixMSFT:
We need to close on if response envelopes should return the raw response when available or not.  At present some code paths do include it which is why there are some test failures.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -110,7 +110,7 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
 }
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string): string {
   // dictionaries are handled slightly different so we do that first
   if (schema.type === SchemaType.Dictionary) {
     imports.add('strings');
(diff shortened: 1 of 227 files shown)


src/generator/operations.ts
```diff
@@ -652,11 +683,12 @@ function generateResponseUnmarshaller(op: Operation, response: Response, imports
     if ((mediaType === 'XML' && schemaResponse.schema.type === SchemaType.Array) || schemaResponse.schema.language.go!.discriminatorInterface) {
       target = 'result';
     }
-    unmarshallerText += `\treturn &result, resp.UnmarshalAs${getMediaFormat(response.schema, mediaType, `&${target}`)}\n`;
+    unmarshallerText += `\terr := resp.UnmarshalAs${getMediaFormat(response.schema, mediaType, `&${target}`)}\n`;
```
catalinaperalta, line 686: Why does this need to be separated now?

jhendrixMSFT: It broke the case for multiple responses, where the return type is `interface{}`.  Might be a bug in Go, I didn't investigate further.

src/generator/operations.ts
```diff
@@ -203,6 +203,36 @@ function generateMultiRespComment(op: Operation): string {
...
+function getZeroReturnValue(op: Operation, apiType: 'api' | 'op'): string {
+  if (!op.responses) {
+    // no responses return *http.Response
+    return 'nil';
+  }
+  let returnType = 'nil';
+  if (isLROOperation(op)) {
+    if (apiType === 'op') {
+      // the operation returns an *azcore.Response
+      returnType = 'nil';
```
catalinaperalta, line 215: Just curious, when does this case happen?

src/generator/operations.ts
catalinaperalta: Ok good to know 👍

catalinaperalta (approved):
LGTM

src/generator/operations.ts
jhendrixMSFT: For operations that don't return a model (the `httpinfrastructuregroup` has a lot of examples).

catalinaperalta: Ok I didnt see one with an LRO, but seems like a good idea just in case 👍

PR #536, opened 2020-12-04
Don't unmarshall into response envelopes

jhendrixMSFT:
To avoid allocating response envelopes on the heap, unmarshall into a
temporary variable.


jhendrixMSFT:
This the second half of the work introduced in https://github.com/Azure/autorest.go/pull/532

catalinaperalta (approved):
LGTM

PR #537, opened 2020-12-08
Improve error message

catalinaperalta:
Adding more detail to generated error messages.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -742,7 +742,7 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
   return text;
 }
 
-// differentiates between objects and basic types defined in Go. Returns false if it finds an object to expand for the error message. 
+// differentiates between objects and basic types defined in Go. Returns true if it finds an object to expand for the error message. 
 function isObjectType(schema: Schema): boolean {
   switch (schema.type) {
     case SchemaType.Object:
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -246,27 +246,25 @@ function generateStructs(objects?: ObjectSchema[]): StructDef[] {
         text += `\tif e.${prop.language.go!.name} != nil {\n`;
         // check if the property is an object or a basic type and output the corresponding error message.
         // this will only include details for the top level information in the corresponding type. 
-        if (!isObjectType(prop.schema)) {
+        if (!isObjectSchema(prop.schema)) {
           text += `\t\tmsg += fmt.Sprintf("${prop.language.go!.name}: %v\\n", *e.${prop.language.go!.name})\n`;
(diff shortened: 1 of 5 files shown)


catalinaperalta:
Example output with information from the service: 

```
InnerError:
        Code: ResourceGroupNotFound
        Message: Resource group '<testresources>' could not be found.
```

src/generator/models.ts
```diff
@@ -244,7 +244,32 @@ function generateStructs(objects?: ObjectSchema[]): StructDef[] {
...
       for (const prop of values(structDef.Properties)) {
         text += `\tif e.${prop.language.go!.name} != nil {\n`;
-        text += `\t\tmsg += fmt.Sprintf("${prop.language.go!.name}: %v\\n", *e.${prop.language.go!.name})\n`;
+        // check if the property is an object or a basic type and output the corresponding error message.
+        // this will only include details for the top level information in the corresponding type. 
+        if (!isObjectType(prop.schema)) {
+          text += `\t\tmsg += fmt.Sprintf("${prop.language.go!.name}: %v\\n", *e.${prop.language.go!.name})\n`;
+        } else {
+          // if the property is an object schema add the information in the Go struct to the error message 
+          if (prop.language.go!.errorType) {
```
jhendrixMSFT, line 253: You could condense this a bit by moving this `if` block up as an `else if` on line 251.

jhendrixMSFT: Also curious if we have an example of this in any swaggers?

src/generator/models.ts
```diff
@@ -716,3 +741,13 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
   text += `\t}\n`;
   return text;
 }
+
+// differentiates between objects and basic types defined in Go. Returns true if it finds an object to expand for the error message. 
+function isObjectType(schema: Schema): boolean {
```
jhendrixMSFT, line 746: This already exists in `common/helpers.ts`, see `isObjectSchema()`

src/generator/models.ts
catalinaperalta: I havent seen this in any of the swaggers. I added it in case it does occur. My thought was that these nicer error messages would be embedded into each other if it ever happened. But I could leave it unimplemented OR throw an error if it ever does happen so we can take a look at it then?

jhendrixMSFT: No you can leave it, will be handy if that ever happens.

src/generator/models.ts
catalinaperalta: Done. Removed the repeated function.

src/generator/models.ts
catalinaperalta: Done, fixed the if statement.

jhendrixMSFT (approved):

PR #540, opened 2020-12-10
add the missing return statement

ArcturusZhang:
Fixes https://github.com/ArcturusZhang/autorest.go/pull/new/fix-list

To avoid this issue I just add the return statement every time we get an error


ArcturusZhang:
Hi @jhendrixMSFT could you please have a review?

jhendrixMSFT (approved):

ArcturusZhang:
I also opened one PR here https://github.com/Azure/azure-sdk-for-go/pull/13944
to see if the linter would raise any objections on this fix

ArcturusZhang:
Hi @jhendrixMSFT it seems that the linter does not have any problems on this fix, could we merge this?

PR #541, opened 2020-12-10
skip the validations on polymorphic types

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/13821

### DIFF
--- a/src/Extensions.cs
+++ b/src/Extensions.cs
@@ -589,6 +589,12 @@ namespace AutoRest.Go
                     }
                     else if (composite != null)
                     {
+                        if (composite.IsPolymorphic)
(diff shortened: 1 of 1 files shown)


ArcturusZhang:
Hey @jhendrixMSFT could you please have a review on this? I am not sure the place I changed would get rid of all the places related with validation of polymorphic types, but just in the case of [this issue](https://github.com/Azure/azure-sdk-for-go/issues/13821), this would fix it.
You could find the result of the fix here: https://github.com/Azure/azure-sdk-for-go/pull/13945

jhendrixMSFT (approved):

PR #542, opened 2020-12-10
Return clients by reference

jhendrixMSFT:
Fixed up a headergroup test

### DIFF
--- a/test/autorest/arraygroup/arraygroup_test.go
+++ b/test/autorest/arraygroup/arraygroup_test.go
@@ -15,7 +15,7 @@ import (
 )
 
 func newArrayClient() ArrayClient {
-	return NewArrayClient(NewDefaultConnection(nil))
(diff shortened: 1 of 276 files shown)


test/autorest/arraygroup/arraygroup_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func newArrayClient() ArrayClient {
-	return NewArrayClient(NewDefaultConnection(nil))
+	return *NewArrayClient(NewDefaultConnection(nil))
```
catalinaperalta, line 18: Is this on purpose?

jhendrixMSFT: Good catch, copy/pasta error.

jhendrixMSFT: Fixed.

catalinaperalta (approved):
LGTM

PR #543, opened 2020-12-10
Add logging options to connection options

jhendrixMSFT:
Pipe the azcore.LogOptions through client connections.

### DIFF
--- a/src/generator/connection.ts
+++ b/src/generator/connection.ts
@@ -53,6 +53,8 @@ function generateContent(session: Session<CodeModel>, exportClient: boolean): st
   text += '\tRetry azcore.RetryOptions\n';
   text += '\t// Telemetry configures the built-in telemetry policy behavior.\n';
   text += '\tTelemetry azcore.TelemetryOptions\n';
+  text += '\t// Logging configures the built-in logging policy behavior.\n';
+  text += '\Logging azcore.LogOptions\n';
   if (isARM && session.model.security.authenticationRequired) {
     imports.add('github.com/Azure/azure-sdk-for-go/sdk/armcore');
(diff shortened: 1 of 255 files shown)


src/generator/connection.ts
```diff
@@ -53,6 +53,8 @@ function generateContent(session: Session<CodeModel>, exportClient: boolean): st
   text += '\tRetry azcore.RetryOptions\n';
   text += '\t// Telemetry configures the built-in telemetry policy behavior.\n';
   text += '\tTelemetry azcore.TelemetryOptions\n';
+  text += '\t// Logging configures the built-in logging policy behavior.\n';
+  text += '\Logging azcore.LogOptions\n';
```
catalinaperalta, line 57: I think this is missing a `\t`

src/generator/connection.ts
```diff
@@ -67,6 +69,7 @@ function generateContent(session: Session<CodeModel>, exportClient: boolean): st
     text += '\t\tRegisterRPOptions: armcore.DefaultRegistrationOptions(),\n';
   }
   text += '\t\tTelemetry: azcore.DefaultTelemetryOptions(),\n';
+  text += '\t\Logging: azcore.DefaultLogOptions(),\n';
```
catalinaperalta, line 72: Same thing here??

src/generator/connection.ts
jhendrixMSFT: Fixed.

src/generator/connection.ts
jhendrixMSFT: Fixed.

catalinaperalta (approved):
LGTM

PR #545, opened 2020-12-17
Future mocking

jhendrixMSFT:
Make the `Result()` method a field on the `WidgetFuture` type so it can be replaced by mock implementations.
Embed `azure.FutureAPI` so it can be replaced with a mock implementation.
Updated vendored dependencies to latest version.


jhendrixMSFT:
Codegen fix for https://github.com/Azure/azure-sdk-for-go/issues/14010

jhendrixMSFT:
@ArcturusZhang can you please test this out a bit to make sure everything is working?  I did test this using redis and it worked, might be good to generate the SDK to ensure nothing is broken before we merge this.

src/Templates/MethodTemplate.cshtml
```diff
@@ -283,13 +283,57 @@ else
     func (client @(Model.Owner)) @(Model.SenderMethodName)(req *http.Request) (@senderRetSig) {
     @if (Model.IsLongRunningOperation())
     {
+        var ftg = Model.ReturnType.Body.Cast<FutureTypeGo>();
```
jhendrixMSFT, line 286: The changes here are pretty much a copy/paste of the changes removed from ModelTemplate.cshtml with some slight tweaking.

ArcturusZhang:
Thanks @jhendrixMSFT , give me some time to generate more packages and test them - for instance in terraform? Eventually they will have to adopt this version

ArcturusZhang:
Hey @jhendrixMSFT the apidiff tool will report a lot of breaking changes regarding on this change. I suppose we should also update the apidiff tool to make a field with type of a function effective the same as it has a function with the same signature. Do you think this make sense?

ArcturusZhang:
> Hey @jhendrixMSFT the apidiff tool will report a lot of breaking changes regarding on this change. I suppose we should also update the apidiff tool to make a field with type of a function effective the same as it has a function with the same signature. Do you think this make sense?

Well, it may be quite difficult to implement - since there is also an anonymous field changed from `azure.Future` to `azure.FutureAPI`...

ArcturusZhang:
Hey @jhendrixMSFT I just done testing this new change in terraform, and I did not see any issues

src/Templates/MethodTemplate.cshtml
```diff
@@ -283,13 +283,57 @@ else
...
+		        }
+		        @if (ftg.IsDefaultReturnType)
+                {
+                    // for default return types (i.e. autorest.Response)
+                    // assign the raw HTTP response to the *http.Response field
+                    @:@(resultVarTarget).Response = future.Response()
+                }
+                else
+                {
+                    // modeled return types embed autorest.Resposne as an anonymous field
```
catalinaperalta, line 322: Nit: should be autorest.Response

catalinaperalta (approved):
LGTM

src/Templates/MethodTemplate.cshtml
jhendrixMSFT: Fixed

jhendrixMSFT:
@ArcturusZhang so the TF testing you did, you replaced their copy of the SDK with codegen using this change?

ArcturusZhang:
> @ArcturusZhang so the TF testing you did, you replaced their copy of the SDK with codegen using this change?

Yes, replaced by this release on my fork: https://github.com/ArcturusZhang/azure-sdk-for-go/releases/tag/v49.2.1

PR #549, opened 2021-01-11
Only check namespaces when flag is on

ArcturusZhang:
This PR introduces a new flag `namespace-chk` to enable the checking logic of `namespace`, which will be the generated package name. I will add the same logic in the scripts in the pipeline to provide more friendly error message.
@jhendrixMSFT

### DIFF
--- a/src/CodeGeneratorGo.cs
+++ b/src/CodeGeneratorGo.cs
@@ -46,11 +46,9 @@ namespace AutoRest.Go
             var folder = Path.GetFullPath(Settings.Instance.Host.GetValue<string>("output-folder").Result).Replace('\\', '/');
             // check if the namespace contains illegal characters
             var ns = Settings.Instance.Host.GetValue<string>("namespace").Result;
-            var r = new Regex(@"^[a-z][a-z0-9_]*[a-z0-9]?$");
-            // skip this check if namespace is not specified
-            if (!string.IsNullOrEmpty(ns) && !r.IsMatch(ns))
+            if (Settings.Instance.Host.GetValue<bool>("namespace-chk").Result)
             {
-                throw new InvalidOperationException($"namespace can only contains lower case letters, numbers and underscore");
+                NamespaceCheck(ns);
             }
 
             // if preview-chk:true is specified verify that preview swagger is output under a preview subdirectory.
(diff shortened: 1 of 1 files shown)


src/CodeGeneratorGo.cs
```diff
@@ -46,11 +46,9 @@ public override async Task Generate(CodeModel cm)
             var folder = Path.GetFullPath(Settings.Instance.Host.GetValue<string>("output-folder").Result).Replace('\\', '/');
             // check if the namespace contains illegal characters
             var ns = Settings.Instance.Host.GetValue<string>("namespace").Result;
-            var r = new Regex(@"^[a-z][a-z0-9_]*[a-z0-9]?$");
-            // skip this check if namespace is not specified
-            if (!string.IsNullOrEmpty(ns) && !r.IsMatch(ns))
+            if (Settings.Instance.Host.GetValue<bool>("namespace-chk").Result)
```
jhendrixMSFT, line 49: We're putting it behind a flag because the failure is a problem for swagger authors (similar to the preview thing)?

ArcturusZhang: well, I am not sure whether it is the code generator's obligation to verify the generated code could work (similar problem happens in the `nextLink`s), but this namespace works as a package name, wrong namespace will certainly cause the generated code not working.
Empty namespace or wrong namespace is truly a problem that the swagger author should resolve (usually something wrong in the readme).

jhendrixMSFT: Agreed.  I think this was another change I had made in the code generator when the POR was to work towards auto-generation and release of modules.

jhendrixMSFT (approved):

src/CodeGeneratorGo.cs
ArcturusZhang: I am sorry but what does "POR" stand for here?

ArcturusZhang:
Somehow my commit and I are considered as different persons on GitHub, interesting. I wonder how that happens...

src/CodeGeneratorGo.cs
jhendrixMSFT: Plan Of Record

PR #550, opened 2021-01-12
Add support to build autorest.go with NodeJS version > 10

tmeckel:
Currently it is not possible to build `autorest.go` on a machine were NodeJS with a higher version of 10 is installed (used) because `graceful-fs`, which is a dependency of  `gulp`, is using code which is no more supported by NodeJS > 10.

The following error is created in that case `ReferenceError: primordials is not defined`.

The pull request contains a NPM preinstall script which creates a `npm-shrinkwrap.json` file, which forces the dependency of `graceful-fs` to at least version 4.0.0 that does not have the issues with NodeJS what the older versions have and thus create the above mentioned error.

With that preinstall scritpt in place arbitrary versions of NodeJS can used to build, test and deploy `autorest.go` with having to update Gulp and it's corresponding task implementations used by `autorest.go`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -306,3 +306,5 @@ package/
 *.tgz
 src/obj/
 *.log
+
+**/npm-shrinkwrap.json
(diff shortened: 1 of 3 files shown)


jhendrixMSFT (approved):
Thanks!  Will be nice to have this irritant fixed.

PR #551, opened 2021-01-15
Remove DefaultOptions() method pattern

jhendrixMSFT:
Update dependencies which no longer use the pattern.


catalinaperalta (approved):
LGTM

PR #554, opened 2021-01-22
Consolidate unnecessary command line switches

jhendrixMSFT:
The armcore-connection and head-as-boolean switches have been removed in
favor of using azure-arm and openapi-type settings.
Pinned RP swaggers to specific versions to avoid random CI failures due
to swagger changes.
Added new status code to formatStatusCode() helper.

### DIFF
--- a/.scripts/regeneration.js
+++ b/.scripts/regeneration.js
@@ -53,22 +53,22 @@ const goMappings = {
 for (namespace in goMappings) {
     // for each swagger run the autorest command to generate code based on the swagger for the relevant namespace and output to the /generated directory
     const inputFile = swaggerDir + goMappings[namespace];
-    generate(inputFile, 'test/autorest/' + namespace, '--head-as-boolean=true');
+    generate(inputFile, 'test/autorest/' + namespace);
 }
 
-const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
+const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/1a7b2bbb84370bd96c0bf53f13f56f4fc2cafab9/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
 generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage" --openapi-type="data-plane"');
 
-const network = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/network/resource-manager/readme.md';
-generateFromReadme(network, 'package-2020-03', 'test/network/2020-03-01/armnetwork', '--credential-scope="https://management.azure.com//.default" --armcore-connection=true --module=armnetwork');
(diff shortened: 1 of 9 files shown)


src/generator/operations.ts
```diff
@@ -225,7 +224,7 @@ function getZeroReturnValue(op: Operation, apiType: 'api' | 'op' | 'handler'): s
   } else if (hasSchemaResponse(op)) {
     // simple schema response
     returnType = `${(<SchemaResponse>op.responses![0]).schema.language.go!.responseType.name}{}`;
-  } else if (op.language.go!.headAsBoolean) {
+  } else if (op.language.go!.headAsBoolean === true) {
```
catalinaperalta, line 227: Why does this have to explicitly check if it's equal to true?

.scripts/regeneration.js
```diff
@@ -53,22 +53,22 @@ const goMappings = {
...
+const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/1a7b2bbb84370bd96c0bf53f13f56f4fc2cafab9/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
 generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage" --openapi-type="data-plane"');
 
-const network = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/specification/network/resource-manager/readme.md';
-generateFromReadme(network, 'package-2020-03', 'test/network/2020-03-01/armnetwork', '--credential-scope="https://management.azure.com//.default" --armcore-connection=true --module=armnetwork');
+const network = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/1c100daabed6a008c76f3e4e0cae4e7eac5a5bf4/specification/network/resource-manager/readme.md';
+generateFromReadme(network, 'package-2020-03', 'test/network/2020-03-01/armnetwork', '--credential-scope="https://management.azure.com//.default" --module=armnetwork --azure-arm=true');
```
catalinaperalta, line 63: I understand this is to avoid CI failures, just wondering if this could lead us to not catch bugs in new features for generation as they come out? (Since we added these to test our generation against actual swaggers.) Not saying we absolutely shouldnt do this...just wondering if this will hide things we should be aware of?

src/generator/operations.ts
jhendrixMSFT: IIRC the former will return true if the key `headAsBoolean` exists, it doesn't check its value.  It's a bit pedantic since we only ever set this to `true`, but could lead to bugs in the odd case we decide to set it to `false`.

.scripts/regeneration.js
jhendrixMSFT: It's a trade-off of stability vs. being on the bleeding edge.  For reference, we do the same thing with the test server to avoid this same issue.  It does mean that we will need to update these periodically though (same with the test server).  Another benefit is that since the inputs are stable, it reduces the noise in the diff due to codegen changes.

src/generator/operations.ts
catalinaperalta: Ok, sounds good 👍

.scripts/regeneration.js
catalinaperalta: Ok it's fine. I think overall if we update every so often it shouldnt be a huge problem.

catalinaperalta (approved):
LGTM

PR #556, opened 2021-01-28
Fix error responses

ArcturusZhang:


ArcturusZhang (commented):
One of the file somehow has been all changed because of the line endings. I have marked the two key parts in that file.
But if we could find a way not to fix this, it would be the best

jhendrixMSFT:
Looks good, will approve and merge once you ack you're ok with the testing.

ArcturusZhang:
Well... the latest two commits break the polymorphism handling. I need to refactor the whole thing

src/TransformerGo.cs
```diff
@@ -21,11 +21,11 @@ public override CodeModelGo TransformCodeModel(CodeModel cm)
 
             SwaggerExtensions.ProcessGlobalParameters(cmg);
             TransformEnumTypes(cmg);
+            TransformMethods(cmg);
```
jhendrixMSFT, line 24: I'm not surprised this broke stuff.  The ordering here is really touchy as too many steps are interdependent.  Can you please put this back and add a comment that basically says "DON'T EVER CHANGE THIS"

ArcturusZhang: Yeah... I have to revert it and find some other way to fix this...

ArcturusZhang:
@jhendrixMSFT I fixed this issue... but I am kind of confused by the behavior of modeler. Check this out:
```
          "responses": {
            "OK": {
              "body": {
                "$ref": "6"
              },
              "isNullable": true
            },
            "Created": {
              "body": {
                "$ref": "9"
              },
              "isNullable": true
            }
          },
          "defaultResponse": {
            "body": {
              "$ref": "1"
            },
            "isNullable": true
          },
          "returnType": {
            "body": {
              "$ref": "6"
            },
            "isNullable": true
          },
```
this is the model generated by the modeler (before our transform). As you can see, 200 and 201 have different models. But its return type only has one model referenced, instead of creating a new `Object` that could be either of them. Why is this happening?
Thanks to this, after my fix, the generator now also adds a `autorest.Response` to the model of `201` (`Created`) too, despite no function will return that type.

test/src/tests/generated/httpinfrastructuregroup/models.go
```diff
@@ -21,8 +21,9 @@ type A struct {
 
 // B ...
 type B struct {
-	TextStatusCode *string `json:"textStatusCode,omitempty"`
-	StatusCode     *string `json:"statusCode,omitempty"`
+	autorest.Response `json:"-"`
```
ArcturusZhang, line 24: This is the return type of `Created` I mentioned in the comment above.
Despite no functions are returning this type, we still add a `autorest.Response` to it...

ArcturusZhang:
I just fixed the issue of extra response field in https://github.com/Azure/autorest.go/pull/556/commits/47a03ea008421f2317aa6be8c7162ff95eb9e21b

ArcturusZhang:
@jhendrixMSFT you could check my changes after your last review in these commits: 
https://github.com/Azure/autorest.go/pull/556/files/2b02ec2908b95667d99f26a38d7a77a55a6efb56..47a03ea008421f2317aa6be8c7162ff95eb9e21b

src/Model/MethodGo.cs
```diff
@@ -1,699 +1,752 @@
...
+            Logger.Instance.Log(Category.Debug, $"Method {SerializedName} has the following non-error & non-empty responses (total {nonErrorNonEmptyStatusCodes.Count()}): {string.Join(", ", nonErrorNonEmptyStatusCodes.Select(resp => resp.Value?.Body?.Name))}");
+            // categorize the models by its name
+            var nonErrorNonEmptyResponses = NonErrorNonEmptyResponses.ToList();
+            Logger.Instance.Log(Category.Debug, $"Non-error & non-empty models: {string.Join(", ", nonErrorNonEmptyResponses.Select(model => model.Body.Name))}");
+            // fix the original return type from the modeler
+            ReturnType = FixedReturnType();
+            Logger.Instance.Log(Category.Debug, $"Return Type of {SerializedName}: {ReturnType?.Body?.Name}");
+        }
+
+        public Response FixedReturnType()
```
jhendrixMSFT, line 104: Can't we move this into `ReturnValue()`?  Now we have two ways of getting the return type, how do you know the right one is being invoked?

ArcturusZhang: Hmm, reasonable... let me refactor...

jhendrixMSFT:
Fixes https://github.com/Azure/autorest/issues/3494

src/Model/MethodGo.cs
ArcturusZhang: Hi @jhendrixMSFT it turns out that I have to keep this function. instead, I removed the `ReturnValue` function because `ReturnType` could never be null and this makes it no difference from `ReturnType` (inherited from `Method`)
We also cannot override the property `ReturnType` using this function - otherwise this makes the setting of the `ReturnType` meaningless.
In the transform process, we need to keep the order of the operation unchanged, otherwise we might get unexpected result.
Previously, the transformation of the methods happens after the transformation of models. Therefore in the model transformation, it is better that we only use the transformed value for this transformation, and do not introduce the transformation prematurely.
Based on the above reasons, I believe we need to keep this.

ArcturusZhang: To clarify, we should never use this function, unless we are in the transformation process. In this way we can know which one is the right one to invoke - the right one is always the `ReturnType` inherited from `Method`

ArcturusZhang: and FYI I am regenerating every package in the track 1 sdk now to see if there is any bad impact

jhendrixMSFT: Ok, if we need to keep it and it's only used during transformation, can you please add a doc comment to this method explaining that?

I'll be watching for an update on your tests.

ArcturusZhang: I just added a comment on this function

jhendrixMSFT (approved):

PR #560, opened 2021-02-08
Use Pager and Poller interfaces in azcore

jhendrixMSFT:
Updated codegen to use interfaces in azcore.
Added doc comment to FinalResponse().


catalinaperalta (approved):
LGTM

PR #561, opened 2021-02-09
Sender: fixing a panic when the HTTP Response is nil

tombuildsstuff:
There's a panic in all of the new generated functions, where should `.Response.Response` be nil, accessing `.StatusCode` will panic - example:

https://github.com/Azure/azure-sdk-for-go/blob/dc4dfc8d80edd8c93454b7d3c15292b890b15f78/services/cdn/mgmt/2019-04-15/cdn/customdomains.go#L137

As such this PR nil-checks the response - since whilst the Azure Front Door may receive a request, it's not guaranteed to return a response (at which point `err` will be nil, causing this panic). Whilst a niche bug we see these responses often enough I believe this needs fixing here, in addition to the recent Azure/go-autorest patch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ bld/
 [Oo]bj/
 [Ll]og/
 
+# Jetbrains Rider
+.idea/
+
 # Visual Studio 2015 cache/options directory
 .vs/
 # Uncomment if you have tasks that create the project's static files in wwwroot
--- a/src/Templates/MethodTemplate.cshtml
+++ b/src/Templates/MethodTemplate.cshtml
@@ -324,7 +324,7 @@ else
                     // in it we need an extra ".Response" :(
                     <text>
                     sender := autorest.DecorateSender(client, autorest.DoRetryForStatusCodes(client.RetryAttempts, client.RetryDuration, autorest.StatusCodesForRetry...))
-                    if @(resultVarTarget).Response.Response, err = future.GetResult(sender); err == nil && @(resultVarTarget).Response.Response.StatusCode != http.StatusNoContent {
+                    if @(resultVarTarget).Response.Response, err = future.GetResult(sender); err == nil && @(resultVarTarget).Response.Response != nil && @(resultVarTarget).Response.Response.StatusCode != http.StatusNoContent {
                         @resultVar, err = client.@(ftg.ResponderMethodName)(@(resultVarTarget).Response.Response)
                         if err != nil {
                             err = autorest.NewErrorWithError(err, "@futureTypeName", "Result", @(resultVarTarget).Response.Response, "Failure responding to request")
--- a/src/Templates/MethodTemplate.cshtml
+++ b/src/Templates/MethodTemplate.cshtml
@@ -324,7 +324,11 @@ else
                     // in it we need an extra ".Response" :(
                     <text>
                     sender := autorest.DecorateSender(client, autorest.DoRetryForStatusCodes(client.RetryAttempts, client.RetryDuration, autorest.StatusCodesForRetry...))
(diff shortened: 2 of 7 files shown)


WodansSon (approved):
@tombuildsstuff,  LGTM! 🚀

ArcturusZhang:
Hi @tombuildsstuff could we also include the changes in the test files? You could run `npm run build && npm run regenerate` to refresh the test cases to fix the CI failure here.

ArcturusZhang:
Hi @jhendrixMSFT could you please take a look at this PR?

tombuildsstuff:
@ArcturusZhang I've spent the better part of 25m trying to get this working and there appear to be a large number of security vulnerabilities here, I've opened this issue to track it: https://github.com/Azure/autorest.go/issues/562

Since I can't get this setup locally - would you mind running the `npm` commands and pushing to this branch? I can grant you push access if required (or feel free to send a PR/tag me and I can merge it, if that's easier)

Thanks!

jhendrixMSFT:
We can certainly take this change, however I'm wondering if a better fix would be to modify our sender in `go-autorest` to return an error if both the `*http.Response` and the `error` are `nil` (which really is a violation of the contract stated for `http.Client.Do()`, would be nice to understand what exactly can cause this to happen).  @tombuildsstuff what are your thoughts?  With this change as-is, looks to me that in this case it will return `nil, nil` which seems undesirable.

tombuildsstuff:
Sounds like it should be fixed in both places tbh, as you mention `nil, nil` should throw some kind of error and that sounds reasonable to me. It's worth noting that this is a common enough issue that we're seeing this in from the Data Plane API's too: https://github.com/hashicorp/terraform/pull/27678 (albeit this may be fixed by upgrading AutoRest, we've not done that in TF Core just yet).

My understanding of this issue is that the Request is getting sent to the Azure Front Door/LB but isn't getting a Response sent back, I'm assuming due to network congestion - I've experienced this myself a handful of times (and we frequently see it from AWS -> Azure when running tests overnight) - but never been able to continually repro to debug it, the test failures we get are in the form of `Unable to communicate between [ip] and [ip]`, so perhaps it's a routing issue? I'll try and post some specific error messages the next time I see it.

jhendrixMSFT:
I slightly modified your fix and updated the codegen, PTAL.  Also working on updating the senders in `go-autorest` to return an error when there's no response or error returned from the client.

ArcturusZhang:
> I slightly modified your fix and updated the codegen, PTAL. Also working on updating the senders in `go-autorest` to return an error when there's no response or error returned from the client.

We once run into a similar issue before (if you recall this). But the result is that we have a standard that the `http.Response` will never be nil according to the documentation. But we did not come to a result of this then...

tombuildsstuff:
@ArcturusZhang whilst there may be a convention this isn't nil - it's a pointer so it can be by design (and likely should be, since as in the server-side issues above if the server doesn't send a response for whatever reason, there isn't one - so having an empty response would hide the issue which is arguably worse)

@jhendrixMSFT thanks for pushing those - this LGTM 👍

tombuildsstuff:
Since I've linked to a Data Plane example, here's today's ARM example: https://github.com/terraform-providers/terraform-provider-azurerm/issues/10522 - this instance is deleting a CosmosDB Account

jhendrixMSFT (approved):

jhendrixMSFT:
While I agree since the response is a pointer, the fact that both return values can be `nil` seems like a bug in the HTTP client.  My reading of the docs implies that you'll get one or the other (or sometimes both but the error always takes precedence).

https://pkg.go.dev/net/http#Client.Do

I will continue to dig around on this as I'm sure we're not the only people to have bumped into this.

PR #564, opened 2021-02-22
Fix issue that sometimes the unmarshal function is missing

ArcturusZhang:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/14336


ArcturusZhang:
Hi @jhendrixMSFT could you please take a look at this PR? We would like to include this fix in this week's release.

jhendrixMSFT (approved):

Issue #563, opened 2021-02-19
Panic after <foo>Sender returns nil, err


jim-minter:
In a function like the following in github.com/Azure/azure-sdk-for-go:

```golang
func (client PrivateZonesClient) Delete(ctx context.Context, resourceGroupName string, privateZoneName string, ifMatch string) (result PrivateZonesDeleteFuture, err error) {
	if tracing.IsEnabled() {
		ctx = tracing.StartSpan(ctx, fqdn+"/PrivateZonesClient.Delete")
		defer func() {
			sc := -1
			if result.Response() != nil {
				sc = result.Response().StatusCode
			}
			tracing.EndSpan(ctx, sc, err)
		}()
	}
	req, err := client.DeletePreparer(ctx, resourceGroupName, privateZoneName, ifMatch)
	if err != nil {
		err = autorest.NewErrorWithError(err, "privatedns.PrivateZonesClient", "Delete", nil, "Failure preparing request")
		return
	}

	result, err = client.DeleteSender(req)
	if err != nil {
		err = autorest.NewErrorWithError(err, "privatedns.PrivateZonesClient", "Delete", nil, "Failure sending request")
		return
	}

	return
}
```

A panic happens if DeleteSender returns a nil result and a non-nil error.  In that case, the `err != nil` branch is followed, then on the `return`, the `defer`red function is called.  It attempts to evaluate `result.Response()` but result is nil.  Presumably the `nil` check in the `defer`red function should be `if result != nil && result.Response() != nil {`?

This class of potential failure mode seems pretty widespread in github.com/Azure/azure-sdk-for-go right now, please can it be fixed?

jim-minter:
@jhendrixMSFT @ArcturusZhang ptal?

ArcturusZhang:
Hey @jim-minter in your case, `result` is of type `PrivateZonesDeleteFuture` which is a struct and can never be `nil`. The clause `result != nil` will give you a compiling error.

I believe the root cause is here: https://github.com/Azure/azure-sdk-for-go/blob/5f5b9952f37b2085f3ba6844f129f1d66fd6d7f9/services/privatedns/mgmt/2018-09-01/privatedns/privatezones.go#L229
[...]

PR #565, opened 2021-02-23
fix future panic

ArcturusZhang:
Fixes https://github.com/Azure/autorest.go/issues/563


src/Templates/MethodTemplate.cshtml
```diff
@@ -42,8 +42,8 @@ func (client @(Model.Owner)) @(Model.Name)(@Model.MethodParametersSignature()) (
 @if (Model.IsLongRunningOperation())
 {
     <text>
-        if result.Response() != nil {
-        sc = result.Response().StatusCode
+        if result.FutureAPI != nil && result.FutureAPI.Response() != nil {
+        sc = result.FutureAPI.Response().StatusCode
```
ArcturusZhang, line 46: These two lines are the only thing I changed in this PR.
Other changes are here because the VS is always auto-format this thing which is quite annoying

jhendrixMSFT (approved):

PR #566, opened 2021-02-24
Clean up test results comparison

jhendrixMSFT:
Removed helpers package to get accurate test failure line info.
Switched reflect.DeepEqual in favor of cmp.Diff which has vastly
superior pretty-printing of differences.


jhendrixMSFT:
This is test-only clean-up, no functional changes.

PR #567, opened 2021-02-24
Return arrays and maps by value

jhendrixMSFT:
Removed the extra indirection as it's not necessary and negatively
impacts useability.
Moved byValue off the schema and onto the property for better control of
where it can be applied.
Fixed impacted tests.


catalinaperalta (approved):
LGTM

PR #568, opened 2021-02-25
Don't crash on invalid pager definition

jhendrixMSFT:
Fail with a useful error message instead.
Cleaned up throwing errors to get stack traces.
Add 403 to list of errors.

### DIFF
--- a/src/transform/namer.ts
+++ b/src/transform/namer.ts
@@ -45,7 +45,7 @@ export async function namer(session: Session<CodeModel>) {
     // this looks like it already has data for this model.
     // send back an error
     session.error('bad flavor', ['go:1000', 'already-processed'], model.language.go);
-    throw Error('Go Namer Failed');
+    throw new Error('Go Namer Failed');
   }
 
   // copy all the .language.default data into .language.go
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/619

src/transform/namer.ts
```diff
@@ -45,7 +45,7 @@ export async function namer(session: Session<CodeModel>) {
     // this looks like it already has data for this model.
     // send back an error
     session.error('bad flavor', ['go:1000', 'already-processed'], model.language.go);
-    throw Error('Go Namer Failed');
```
catalinaperalta, line 48: Not saying to do this now, but could this be improved?

catalinaperalta (approved):
LGTM

src/transform/namer.ts
jhendrixMSFT: I think this is one of those "this should never happen but just in case" type of error condition, i.e. if we hit this then we broke something elsewhere (Garrett had added this in the template project).

PR #572, opened 2021-03-01
Add export-clients command line switch

jhendrixMSFT:
To force exporting of clients, mostly useful for data-plane where the
default is to not export the clients.
Add aztables to regeneration script that uses this option.


catalinaperalta:
Did we get a request for this? (Is tables going to use the exported client?)

jhendrixMSFT:
Yes it might, was discussing this with Mayuri this morning.

catalinaperalta (approved):
LGTM

PR #574, opened 2021-03-02
Fix handling of multi-request operations

jhendrixMSFT:
Modifying the operations collection while iterating over it caused some
operations to be skipped in processing.
Updated version of autorest.core to latest version.


catalinaperalta (approved):
LGTM

PR #575, opened 2021-03-03
Fold identical error schemas into one unmarshaller

jhendrixMSFT:
Don't emit separate case statements for status codes when they unmarshal
into the same type.
Include a default case statement if one isn't specified.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -780,13 +780,12 @@ function createProtocolErrHandler(op: Operation, imports: ImportManager): string
     return text;
   }
 
-  const generateUnmarshaller = function (exception: Response, prefix: string) {
+  const generateUnmarshaller = function (schemaError: Schema, prefix: string) {
     let unmarshaller = '';
-    if (exception.language.go!.genericError) {
+    if (schemaError.language.default.name === 'generic') {
       unmarshaller += `${prefix}${generateGenericError()}`;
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/go-autorest/issues/620

src/generator/operations.ts
```diff
@@ -808,22 +807,43 @@ function createProtocolErrHandler(op: Operation, imports: ImportManager): string
...
+  for (const exception of values(op.exceptions)) {
+    let errSchema = genericErr;
+    if (!exception.language.go!.genericError) {
+      errSchema = (<SchemaResponse>exception).schema;
+    }
+    if (!foldedMap.has(errSchema)) {
+      foldedMap.set(errSchema, new Array<string>());
+    }
+    for (const statusCode of values(<Array<string>>exception.protocol.http!.statusCodes)) {
+      foldedMap.get(errSchema)?.push(statusCode);
```
catalinaperalta, line 823: Why the `?` instead of `!`? Curious because of the flow I thought you could assume `!` this.

catalinaperalta (approved):
LGTM. Left a small question.

src/generator/operations.ts
jhendrixMSFT: Fixed (must have been an auto-complete thing).

PR #577, opened 2021-03-04
Handle some LRO corner-cases

jhendrixMSFT:
LROs are treated as single-response type APIs.  In the event of multiple
schema responses, select the applicable one or fail.
Update pinned swagger versions.


jhendrixMSFT:
Addresses https://github.com/Azure/autorest/issues/3951

jhendrixMSFT:
Note that none of the codegen was changed with this fix, however we cannot build synapse without it.

catalinaperalta (approved):
LGTM!

Issue #578, opened 2021-03-05
Creation of connection.go: NewConnectoin misses argument to armcore.NewRPRegistrationPolicy


scseely:
The track2 generator creates a file called connection.go. The file contains a method, NewConnection. The call to NewRPRegistrationPolicy is missing an argument (endpoint). 

Today, the line generated here reads:

policies = append(policies, armcore.NewRPRegistrationPolicy(cred, &options.RegisterRPOptions))

It should be

policies = append(policies, armcore.NewRPRegistrationPolicy(endpoint, cred, &options.RegisterRPOptions))

It appears that the change should only be to [src/generator/connection.ts:162](https://github.com/Azure/autorest.go/blob/track2/src/generator/connection.ts#L162)

jhendrixMSFT:
The `armcore.NewRPRegistrationPolicy()` is only applicable to ARM services and is enabled implicitly via the `armcore.Connection` (all this is enabled when `azure-arm:true` is set).  We forgot to remove this from codegen when we switched to using `armcore.Connection`.

ArcturusZhang:
@jhendrixMSFT this issue should have been fixed by #579 , maybe we need to close this issue?

jhendrixMSFT:
I will close once we have a new generator with the fix available (trying to wrap up the naming fixes).

jhendrixMSFT:
Fixed in v4.0.0-preview.14

PR #579, opened 2021-03-05
Use azure-arm to enable RP registration policy

jhendrixMSFT:
Don't relay on the OpenAPI type since it defaults to 'arm' which isn't
always correct.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/578

catalinaperalta (approved):
LGTM

PR #582, opened 2021-03-10
Add a flag to control whether we add the type name as enum value prefix

ArcturusZhang:


tombuildsstuff:
@ArcturusZhang is it worth always generating the `[Type]Name` and aliasing the `Name` values to those as a compatibility layer? This'd mean folks could move across gradually without it being a breaking change (and the aliased fields could be marked as Deprecated?)

ArcturusZhang:
> @ArcturusZhang is it worth always generating the `[Type]Name` and aliasing the `Name` values to those as a compatibility layer? This'd mean these are always unique and folks could move across gradually without it being a breaking change (and the aliased fields could be marked as Deprecated?)

I am planning to store the feature flags in the generation metadata (the first part of the changelog.md) to ensure only the new released package will take this change. And we only release packages by release requests, which means in one major version, only a limited number of the packages will receive breaking changes by the enum variables.

The deprecation might not solve this problem, because the issue is that we do not know which enum will not have the prefix. In previous implementations, the enums are ordered alphabetically and only the first enum is without the type name prefix - this might change in the next generation due to changes in swaggers, and the deprecation will then go to different enum entries.

jhendrixMSFT (approved):

PR #593, opened 2021-03-29
Add support for multi-part form data

jhendrixMSFT:
Replaced deprecated azure-tools/codemodel with autorest/codemodel which
was required as part of the fix.
Fixed a bug in transformation to make arrays of ReadSeekCloser by value.


src/package.json
```diff
@@ -71,4 +71,4 @@
     "showdown": "^1.9.1",
     "source-map-support": "0.5.13"
   }
-}
+}
```
catalinaperalta, line 74: Nit: newline

catalinaperalta (approved):
LGTM

PR #594, opened 2021-03-29
Add suffix to client file names

jhendrixMSFT:
To avoid file name collisions, add suffix _client to the client file
names; it also aligns with the contents of the file.
Fixed incorrect parameter name bug for required string bodies.
Use clone() as it handles circurlar references.
Fixed client param comparison to use the param name.


catalinaperalta (approved):
LGTM

PR #595, opened 2021-04-01
Simplify injecting custom pipeline policies

jhendrixMSFT:
Removed NewConnectionWithPipeline() in favor of policy slices in
ConnectionOptions.  The slices are to control where the policies are
injected in the pipeline, either before or after the retry policy.
Update armcore to latest version.


catalinaperalta (approved):
LGTM

PR #596, opened 2021-04-01
More build the world fixes

jhendrixMSFT:
Added clear() method to imports so it can be reset between batch builds.
Ensure synthetic options param doesn't collide with existing param.
Escape reserved names for internal operation method name.
Fixed param group check for optional params on the client.
Rename query var to reqQP to avoid collision with parameter name.
Treat empty nextLinkName as equivalent to null (lenient swagger
acceptance).
Make a copy of the operations collection so it's not modified while
iterating over it.
Ensure isLRO is set for all cases.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -13,6 +13,7 @@ import { ImportManager } from './imports';
 
 // Creates the content in models.go
 export async function generateModels(session: Session<CodeModel>): Promise<string> {
+  imports.clear();
   let text = await contentPreamble(session);
 
   // we do model generation first as it can add imports to the imports list
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -13,20 +13,21 @@ import { ImportManager } from './imports';
 
 // Creates the content in models.go
 export async function generateModels(session: Session<CodeModel>): Promise<string> {
-  imports.clear();
+  // this list of packages to import
+  const imports = new ImportManager();
   let text = await contentPreamble(session);
 
   // we do model generation first as it can add imports to the imports list
-  const structs = generateStructs(session.model.schemas.objects);
+  const structs = generateStructs(imports, session.model.schemas.objects);
   const responseEnvelopes = <Array<Schema>>session.model.language.go!.responseEnvelopes;
   for (const respEnv of values(responseEnvelopes)) {
-    const respType = generateStruct(respEnv.language.go!.responseType, respEnv.language.go!.properties);
+    const respType = generateStruct(imports, respEnv.language.go!.responseType, respEnv.language.go!.properties);
     generateUnmarshallerForResponseEnvelope(respType);
     structs.push(respType);
   }
   const paramGroups = <Array<GroupProperty>>session.model.language.go!.parameterGroups;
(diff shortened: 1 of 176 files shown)


src/generator/models.ts
```diff
@@ -13,6 +13,7 @@ import { ImportManager } from './imports';
 
 // Creates the content in models.go
 export async function generateModels(session: Session<CodeModel>): Promise<string> {
+  imports.clear();
```
catalinaperalta, line 16: It would be cool if we could consolidate this in one place. Not saying it's possible with the code as it is at the moment, but maybe later on. It might be easy to forget to clear it in two different places.

jhendrixMSFT: It's only this way because we use a package-local variable.  If we refactored to remove that, the need for `clear()` would go away.  However, it does mean we have to pass around the instance which is cumbersome.

jhendrixMSFT: I refactored the code to remove the need for `clear()`

src/generator/operations.ts
```diff
@@ -256,7 +256,8 @@ function generateOperation(op: Operation, imports: ImportManager): string {
   }
   let opName = op.language.go!.name;
   if (isLROOperation(op)) {
-    opName = opName[0].toLocaleLowerCase() + opName.substr(1);
+    // uncapitalizing runs the risk of reserved name collision, e.g. Import -> import
+    opName = getEscapedReservedName(opName.uncapitalize(), 'Operation');
```
catalinaperalta, line 260: Question, is this something we should do in the namer?

src/generator/operations.ts
```diff
@@ -1054,7 +1058,8 @@ function generateARMLROBeginMethod(op: Operation, imports: ImportManager): strin
   const zeroResp = getZeroReturnValue(op, 'api');
   text += `func (client *${clientName}) Begin${op.language.go!.name}(${params}) (${returns.join(', ')}) {\n`;
   let opName = op.language.go!.name;
-  opName = opName[0].toLocaleLowerCase() + opName.substr(1);
+  // uncapitalizing runs the risk of reserved name collision, e.g. Import -> import
+  opName = getEscapedReservedName(opName.uncapitalize(), 'Operation');
```
catalinaperalta, line 1062: Same question as above

catalinaperalta (approved):
LGTM, left a question about where some naming is which required exporting the getEspacedReservedName function

src/generator/operations.ts
jhendrixMSFT: We could, it gets a little complex.  In the example above, the namer properly calls this method `Import` so there's no collision.  The only reason we have the collision is for LROs, where we replace `Import` with `BeginImport` and unexport the `Import` method.  Hmm, maybe we could add this info to `protocolNaming`.  I will look into that.

src/generator/operations.ts
jhendrixMSFT: fixed

src/generator/operations.ts
jhendrixMSFT: fixed

PR #597, opened 2021-04-01
Autorest help content

timotheeguerin:
Added help content for autorest to show go specific help when loaded.

Used flags defined in the documentation here https://github.com/Azure/autorest/blob/master/docs/generate/flags.md

Result when loading this extension with `--help`
image


catalinaperalta (approved):
LGTM. Although, do we need to specify the `openapi-type` as well? If not, the should we change our wiki to reflect a more current generation command?

jhendrixMSFT:
Good question.  I believe that `openapi-type` should be part of the docs for `autorest` itself as it's not specific to any code generator.  @timotheeguerin is this switch doc'ed?

timotheeguerin:
@jhendrixMSFT I don't really know looking at [flags.md](https://github.com/Azure/autorest/blob/master/docs/generate/flags.md)  `--openapi-type` is only referenced in the go section. Can maybe bring this up in the autorest meeting to see if anybody else makes use of this flag and the docs/help needs to be updated

PR #598, opened 2021-04-01
Capture build and vet failures

jhendrixMSFT:
Some packages were failing to build but didn't fail CI.
Ensure any build/vet failures are propagated.

### DIFF
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -76,7 +76,7 @@ jobs:
               $exit = 1
             }
           }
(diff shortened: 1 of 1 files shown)


jhendrixMSFT:
See https://dev.azure.com/azure-sdk/public/_build/results?buildId=818338&view=logs&j=07798fbc-a58c-5d19-bd16-cf8bb3064434&t=ffdd06d4-05f6-5f16-74a3-ad6e3dcf76a0 for example of false positive.

PR #599, opened 2021-04-02
Use case-insensitive comparison for header prefix

jhendrixMSFT:
Canonicalization of header keys means a case-insensitive comparison must
be used.
Renamed variable t to res to avoid collision with receivers.
Fixed a case of missing internal marshaller.


jhendrixMSFT:
CC @adreed-msft

catalinaperalta (approved):
LGTM

PR #600, opened 2021-04-07
Bug fix: remove variables declared multiple times within the same context

catalinaperalta:
- fix a bug with multiple prefix variables being declared in the code. Make sure to only declare the prefix variable once per response handler. 
- update test blob storage version to 2020-06-12


src/generator/operations.ts
```diff
@@ -105,11 +105,15 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
...
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string, multPrefix: boolean): [string, boolean] {
   // dictionaries are handled slightly different so we do that first
   if (schema.type === SchemaType.Dictionary) {
     imports.add('strings');
-    let text = `\tprefix := strings.ToUpper("${schema.language.go!.headerCollectionPrefix}")\n`;
+    let firstPrefix = ':';
+    if (multPrefix === true) {
```
iscai-msft, line 113: nit: should this be `multiPrefix`?

iscai-msft (approved):

src/generator/operations.ts
deyaaeldeen: NIT:
```suggestion
    if (multPrefix) {
```

src/generator/operations.ts
```diff
@@ -118,7 +122,7 @@ function formatHeaderResponseValue(propName: string, header: string, schema: Sch
     text += `\t\t\t${respObj}.Metadata[hh[len(prefix):]] = resp.Header.Get(hh)\n`;
     text += '\t\t}\n';
     text += '\t}\n';
-    return text;
+    return [text, true];
```
deyaaeldeen, line 125: Please consider returning an object instead with well-named fields to describe the return value.

src/generator/operations.ts
deyaaeldeen: Also, perhaps it is out of scope for this PR but not sure why the function has that many parameters, I would expect related ones to be grouped in objects instead.

src/generator/operations.ts
deyaaeldeen: Also, consider adding unit tests to test the scenario you are improving in this PR.

deyaaeldeen (commented):
Thanks! I left a few comments

src/generator/operations.ts
```diff
@@ -105,11 +105,15 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
 }
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string, multPrefix: boolean): [string, boolean] {
   // dictionaries are handled slightly different so we do that first
   if (schema.type === SchemaType.Dictionary) {
     imports.add('strings');
-    let text = `\tprefix := strings.ToUpper("${schema.language.go!.headerCollectionPrefix}")\n`;
+    let firstPrefix = ':';
```
timotheeguerin, line 112:
```suggestion
   const prefix = multiPrefix? ":" : "";
```

or even changing to be more descriptive and updating below

```suggestion
   const operator = multiPrefix? ":=" : "=";
```

src/generator/operations.ts
catalinaperalta: Done! Updated to multiPrefix and fixed the if. I will bring up your comment about the function signature once Joel is back so that we can discuss that. We havent really used the model of grouping params into objects evenly throughout the whole generator.

src/generator/operations.ts
catalinaperalta: I'm not sure about adding an object for this specific case, since it will only ever be used for this function? And probably wont be useful elsewhere, but please let me know if you think it is better to include. 

The test for this specific fix is the fact that the code compiles. Without this fix the code doesnt compile and CI would not pass. This is why I updated the storage blobs api version so that this can be tested.

src/generator/operations.ts
catalinaperalta: Done! Thank you for the suggestion, I like it a lot more than what I had.

test/storage/2020-06-12/azblob/zz_generated_blob_client.go
```diff
@@ -1248,6 +1430,18 @@ func (client *blobClient) getPropertiesHandleResponse(resp *azcore.Response) (Bl
...
+	}
+	prefix = strings.ToUpper("x-ms-or-")
+	for hh := range resp.Header {
+		if strings.HasPrefix(strings.ToUpper(hh), prefix) {
+			if result.Metadata == nil {
+				result.Metadata = map[string]string{}
+			}
+			result.Metadata[hh[len(prefix):]] = resp.Header.Get(hh)
+		}
+	}
```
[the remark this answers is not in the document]
JeffreyRichter: In your call to HasPrefix you're calling ToUpper and then comparing the result to a lowercase string; this will always be false.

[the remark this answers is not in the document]
JeffreyRichter: Yes, you can change the := to just = in line 1436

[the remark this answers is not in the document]
catalinaperalta: > Why call ToUpper on a lowercase string literal (like x-ms-or)? Just make the string uppercase and avoid ToUpper. ToUpper creates a new string in the heap which has to be GC'd hurting performance/efficiency.

This change was added in this PR: https://github.com/Azure/autorest.go/pull/599. Let's discuss it on Tuesday in scrum, my impression is that Joel added this on purpose to make the comparison case insensitive. Maybe we can improve the case-insensitive comparison.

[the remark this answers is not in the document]
JeffreyRichter: Use https://golang.org/pkg/strings/#EqualFold instead fo refficient case-insensitive comparisons.

.scripts/regeneration.js
```diff
@@ -57,8 +57,8 @@ for (namespace in goMappings) {
     generate(inputFile, 'test/autorest/' + namespace);
 }
 
-const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/8c1f9b739165035d9ced32761cebb19125bce233/specification/storage/data-plane/Microsoft.BlobStorage/preview/2019-07-07/blob.json';
-generate(blobStorage, 'test/storage/2019-07-07/azblob', '--credential-scope="https://storage.azure.com/.default" --module="azstorage" --openapi-type="data-plane"');
+const blobStorage = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/storage-dataplane-preview/specification/storage/data-plane/Microsoft.BlobStorage/preview/2020-06-12/blob.json';
```
jhendrixMSFT, line 60: In the future, it would be ideal to make such changes in a separate PR.  The codegen changes get lost in the churn.

test/storage/2020-06-12/azblob/zz_generated_blob_client.go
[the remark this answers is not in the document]
catalinaperalta: I can add this in a separate PR, I think we could use EqualFold like this:
```
       for hh := range resp.Header {
		if len(hh) > len("x-ms-meta-") && strings.EqualFold(hh[:len("x-ms-meta-")], "x-ms-meta-") {
			if result.Metadata == nil {
				result.Metadata = map[string]string{}
			}
			result.Metadata[hh[len("x-ms-meta-"):]] = resp.Header.Get(hh)
		}
	}
```

I've tested this locally it works. If you think this is more efficient Jeff, then I can add this update in the generator in another PR.

src/generator/operations.ts
```diff
@@ -105,11 +105,12 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
...
 
 // use this to generate the code that will help process values returned in response headers
-function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string): string {
+function formatHeaderResponseValue(propName: string, header: string, schema: Schema, imports: ImportManager, respObj: string, zeroResp: string, multiPrefix: boolean): [string, boolean] {
   // dictionaries are handled slightly different so we do that first
   if (schema.type === SchemaType.Dictionary) {
     imports.add('strings');
-    let text = `\tprefix := strings.ToUpper("${schema.language.go!.headerCollectionPrefix}")\n`;
+    const operator = multiPrefix? '=' : ":=";
+    let text = `\tprefix ${operator} strings.ToUpper("${schema.language.go!.headerCollectionPrefix}")\n`;
```
jhendrixMSFT, line 113: You can replace the hard-coded `prefix` with `${camelCase(schema.language.go!.headerCollectionPrefix)}` then the name is unique.

test/storage/2020-06-12/azblob/zz_generated_blob_client.go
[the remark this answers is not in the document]
jhendrixMSFT: My vote would be to fix it now considering we're already touching this code.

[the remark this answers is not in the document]
jhendrixMSFT: Then we don't even need the `prefix` var and can simplify the fix.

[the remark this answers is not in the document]
catalinaperalta: Adding this fix in this PR will update several more files, just an FYI. Adding the change now 👍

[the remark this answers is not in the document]
catalinaperalta: Also, my code as is doesnt assign a variable, do we all agree that that is the way to go?

jhendrixMSFT (approved):

PR #601, opened 2021-04-08
Fix unmarshalling for additionalProperties

catalinaperalta:
Unmarshalling for BlobMetadata additional properties was being skipped during unmarshalling because no custom unmarshaller was being generated for the BlobMetadata struct. 

Make sure key and value pairs are both included in the map. 
 
Fix this case for all additional properties.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -793,6 +793,9 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
     if (prop.schema.type === SchemaType.DateTime) {
       text += `\t\t${prop.language.go!.name} *${prop.schema.language.go!.internalTimeType} \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
     } else if (prop.language.go!.isAdditionalProperties || prop.language.go!.needsXMLDictionaryUnmarshalling) {
+      if (prop.language.go!.isAdditionalProperties && prop.language.go!.needsXMLDictionaryUnmarshalling) {
+        sn = ',any'
+      }
       text += `\t\t${prop.language.go!.name} *additionalProperties \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
     }
   }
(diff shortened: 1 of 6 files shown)


src/generator/xmlAdditionalProps.ts
```diff
@@ -26,22 +26,31 @@ type additionalProperties map[string]*string
 // UnmarshalXML implements the xml.Unmarshaler interface for additionalProperties.
 func (ap *additionalProperties) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
```
jhendrixMSFT, line 27: We should adopt the implementation from track 1.5 (unsure why I didn't do this to begin with).

https://github.com/Azure/azure-storage-blob-go/blob/master/azblob/zz_generated_models.go#L38

src/generator/models.ts
```diff
@@ -793,6 +793,9 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
     if (prop.schema.type === SchemaType.DateTime) {
       text += `\t\t${prop.language.go!.name} *${prop.schema.language.go!.internalTimeType} \`${structDef.Language.marshallingFormat}:"${sn}"\`\n`;
     } else if (prop.language.go!.isAdditionalProperties || prop.language.go!.needsXMLDictionaryUnmarshalling) {
+      if (prop.language.go!.isAdditionalProperties && prop.language.go!.needsXMLDictionaryUnmarshalling) {
+        sn = ',any'
```
jhendrixMSFT, line 797: Why is this needed?

src/generator/xmlAdditionalProps.ts
catalinaperalta: I did take a look at the implementation for track 1.5 it doesnt unmarshal the content for BlobMetadata properly. The XML could be malformed though. Since it for the additional properties the startElement is empty for some reason.

jhendrixMSFT:
Closing as this is an issue with the swagger and will be fixed at that level.

PR #603, opened 2021-04-14
Cast discriminator to string if it's an enum type

catalinaperalta:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/14519

### DIFF
--- a/src/generator/polymorphics.ts
+++ b/src/generator/polymorphics.ts
@@ -57,7 +57,12 @@ export async function generatePolymorphicHelpers(session: Session<CodeModel>): P
     text += `\tswitch m["${disc.discriminator!.property.serializedName}"] {\n`;
     for (const val of values(disc.discriminator!.all)) {
       const objSchema = <ObjectSchema>val;
-      text += `\tcase ${objSchema.discriminatorValue}:\n`;
+      let disc = objSchema.discriminatorValue;
+      // when the discriminator value is an enum, cast the const as a string
+      if (disc![0] !== '"') {
+        disc = `string(${disc})`;
+      }
+      text += `\tcase ${disc}:\n`;
       text += `\t\tb = &${val.language.go!.name}{}\n`;
     }
     text += '\tdefault:\n';
--- a/test/autorest/complexgroup/zz_generated_polymorphic_helpers.go
+++ b/test/autorest/complexgroup/zz_generated_polymorphic_helpers.go
@@ -103,7 +103,7 @@ func unmarshalMyBaseTypeClassification(rawMsg *json.RawMessage) (MyBaseTypeClass
 	}
 	var b MyBaseTypeClassification
 	switch m["kind"] {
-	case MyKindKind1:
+	case string(MyKindKind1):
 		b = &MyDerivedType{}
 	default:
 		b = &MyBaseType{}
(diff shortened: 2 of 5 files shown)


catalinaperalta:
Havent tested this with any network operation yet

catalinaperalta:
Tested with a network call, seems to be working fine now.

test/autorest/complexgroup/zz_generated_polymorphic_helpers.go
```diff
@@ -103,7 +103,7 @@ func unmarshalMyBaseTypeClassification(rawMsg *json.RawMessage) (MyBaseTypeClass
 	}
 	var b MyBaseTypeClassification
 	switch m["kind"] {
-	case MyKindKind1:
+	case string(MyKindKind1):
```
jhendrixMSFT, line 106: So it looks like we have tests in the test server that would have caught this.  This looks to be used in the `FlattenComplexClient`.  Can you please add those tests?

src/generator/polymorphics.ts
```diff
@@ -57,7 +57,12 @@ export async function generatePolymorphicHelpers(session: Session<CodeModel>): P
     text += `\tswitch m["${disc.discriminator!.property.serializedName}"] {\n`;
     for (const val of values(disc.discriminator!.all)) {
       const objSchema = <ObjectSchema>val;
-      text += `\tcase ${objSchema.discriminatorValue}:\n`;
+      let disc = objSchema.discriminatorValue;
+      // when the discriminator value is an enum, cast the const as a string
+      if (disc![0] !== '"') {
```
jhendrixMSFT, line 62: This assumes the discriminator value is a string, which to be frank is likely always the case.  You could look at `disc.discriminator!.property.schema` to determine the type to cast it to although it gets a little verbose between `Choice` and `SealedChoice`.  Up to you.

catalinaperalta: I was going to do something like what you suggested but then I saw that in the transform we're doing a check similar to the one I have above when adding the `disciminatorValue`...so I decided to leave it the same. I can switch it to the other check though, let me know if you think it's better. I also agree that I think it's basically always a string.

test/autorest/complexgroup/zz_generated_polymorphic_helpers.go
catalinaperalta: The route is returning a 404 in the testserver, asking about it. Will see why that happened.

src/generator/polymorphics.ts
jhendrixMSFT: Ah ok if we're doing it elsewhere then leave it as-is.

test/autorest/complexgroup/zz_generated_polymorphic_helpers.go
catalinaperalta: I believe this operation was added to the swagger back in 2018 but never had a corresponding route added in the testserver. I think since 1. We arent really focusing on flattening here and 2. Our test is about enums, we should add another test in the testserver for this specific scenario. 

I've added an issue here: https://github.com/Azure/autorest.go/issues/606

jhendrixMSFT: Sounds good.

jhendrixMSFT (approved):

PR #604, opened 2021-04-15
Refactor poller implementation

jhendrixMSFT:
To prep for data-plane poller implementation, no functional changes.
Consolidated poller implementations into pollers.ts.
Refactored ARM implementation.
Removed comments on unexported types and methods.
Fixed poller doc comment generation.
Pinned blob storage swagger to prevent inadvertent changes.


test/synapse/2019-06-01/azartifacts/zz_generated_pollers.go
```diff
@@ -7,74 +7,135 @@
...
 package azartifacts
 
-import "github.com/Azure/azure-sdk-for-go/sdk/azcore"
-
-// CreateDataFlowDebugSessionResponsePoller provides polling facilities until the operation completes
+import (
+	"context"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"net/http"
+	"time"
```
catalinaperalta, line 14: This isnt used here.

catalinaperalta:
So you think data-plane and arm poller implementations will be more similar than we previously envisioned? I guess my main question is, will there be a big benefit in a shared poller file? Or could it get complicated with more future divisions between ARM and data-plane implementations? Not against it by the way, just want to understand the change.

test/synapse/2019-06-01/azartifacts/zz_generated_pollers.go
jhendrixMSFT: Fixed

jhendrixMSFT:
It seems likely.  We'll have a basic `azcore.Poller`, similar to what Python has, that will be used by data-plane.  Both cases need to generate the type-specific poller interfaces and their implementations.  They just have different guts.

catalinaperalta (approved):
LGTM

PR #605, opened 2021-04-15
Fixing an issue when the Future fails but no response is set

tombuildsstuff:
The latest "Future API" changes have broken Future's when a Future fails, such that the "err" condition doesn't set the HTTP Response:

image

This means calls to `future.Response()` will panic because this is nil - ergo this needs updating to use the same approach as before these changes were implemented, by parsing the HTTP response up regardless of what happens

An aside, the Typed Errors present within this function aren't used elsewhere in the codebase, so I'm wondering if the line immediately below should be updated to follow the Track1 convention:

> err = autorest.NewErrorWithError(err, "@futureTypeName", "Result", @(resultVarTarget).Response.Response, "Failure sending request")

WDYT?

As it stands, unfortunately it appears that SDK v53.0.0 is broken if Future's are present (and has been subtly broken in v52.x as well)

### DIFF
--- a/src/Templates/ModelTemplate.cshtml
+++ b/src/Templates/ModelTemplate.cshtml
@@ -516,6 +516,7 @@ if (modelType is DictionaryTypeGo dictionaryType && dictionaryType.SupportsAddit
                 return
             }
             if !done {
+                @:@(resultVarTarget).Response = future.Response()
                 err = azure.NewAsyncOpIncompleteError("@futureTypeName")
                 return
             }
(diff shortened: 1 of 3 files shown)


ArcturusZhang (approved):
LGTM but I suppose our old implementation also have this issue?

Also the test cases need to be regenerated so that the CI will pass. @jhendrixMSFT could you also take a look?

jhendrixMSFT (approved):

PR #608, opened 2021-04-15
Refine doc comments

jhendrixMSFT:
Also removed extra indirection in unmarshalling of polymorphic types.


catalinaperalta (approved):
LGTM

PR #610, opened 2021-04-22
Add support for data-plane LROs

jhendrixMSFT:
Refactored codegen for LROs to support ARM and data-plane.
Fixed bug in internal populate() func.
Added more types to schemaTypeToGoType() and fixed default case to throw
and error instead of bad codegen.
Fixed tracking of host params to not use reference-equality (fixes
potential for duplicate params).
Shorted some var names in codegen.
Reordered some poller fields as a side-effect of refactor.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -58,9 +58,10 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
   }
   if (needsJSONPopulate) {
     text += 'func populate(m map[string]interface{}, k string, v interface{}) {\n';
+    text += '\tif v == nil {\n\t\treturn\n\t}\n';
     text += '\tif azcore.IsNullValue(v) {\n';
     text += '\t\tm[k] = nil\n';
-    text += '\t} else if !reflect.ValueOf(v).IsNil() {\n';
+    text += '\t} else {\n';
     text += '\t\tm[k] = v\n';
     text += '\t}\n';
     text += '}\n\n';
@@ -399,7 +400,6 @@ function generateStructs(imports: ImportManager, objects?: ObjectSchema[]): Stru
       generateInternalUnmarshaller(obj, structDef, parentType);
     }
     if (needsM) {
-      imports.add('reflect');
       imports.add('github.com/Azure/azure-sdk-for-go/sdk/azcore');
       structDef.HasJSONMarshaller = true;
       generateJSONMarshaller(imports, obj, structDef, parentType);
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -61,7 +61,7 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
     text += '\tif v == nil {\n\t\treturn\n\t}\n';
     text += '\tif azcore.IsNullValue(v) {\n';
     text += '\t\tm[k] = nil\n';
-    text += '\t} else {\n';
+    text += '\t} else if !reflect.ValueOf(v).IsNil() {\n';
     text += '\t\tm[k] = v\n';
     text += '\t}\n';
     text += '}\n\n';
@@ -400,6 +400,7 @@ function generateStructs(imports: ImportManager, objects?: ObjectSchema[]): Stru
       generateInternalUnmarshaller(obj, structDef, parentType);
     }
     if (needsM) {
+      imports.add('reflect');
       imports.add('github.com/Azure/azure-sdk-for-go/sdk/azcore');
       structDef.HasJSONMarshaller = true;
       generateJSONMarshaller(imports, obj, structDef, parentType);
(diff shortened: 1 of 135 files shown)


jhendrixMSFT:
Still need the change in `azcore` for the `LROPoller` type and supporting funcs so CI will fail for now.

src/generator/operations.ts
```diff
@@ -1078,34 +1080,15 @@ function generateARMLROBeginMethod(op: Operation, imports: ImportManager): strin
...
-    text += `\t\t\treturn client.${info.protocolNaming.errorMethod}(resp)\n`;
-    text += '\t\t},\n';
-    text += `\t\trespHandler: func(resp *azcore.Response) (${schemaResponse!.schema.language.go!.responseType.name}, error) {\n`;
-    text += generateResponseUnmarshaller(op, schemaResponse!, imports);
-    text += '\t\t},\n';
-    text += `\t\tstatusCodes: []int{${formatStatusCodes(statusCodes)}},\n`;
-  }
-  text += '\t\tpipeline: client.con.Pipeline(),\n';
-  text += '\t}\n';
+  text += emitPoller(op, isARM, imports);
```
catalinaperalta, line 1091: A true nit 😆 but I would switch the param order to end with `isARM` to match the other function signatures.

test/autorest/lrogroup/lrosadsheader_test.go
```diff
@@ -181,17 +151,10 @@ func TestLROSADSBeginPost202NonRetry400(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	poller, err = op.ResumePost202NonRetry400(rt)
-	if err != nil {
-		t.Fatal(err)
-	}
-	res, err := resp.PollUntilDone(context.Background(), 1*time.Millisecond)
```
catalinaperalta, line 188: Why not include all the PollUntilDone methods?

src/generator/operations.ts
jhendrixMSFT: Fixed

test/autorest/lrogroup/lrosadsheader_test.go
jhendrixMSFT: These operations are intended to fail when you first poll the polling endpoint.  So what happens is, since we now poll in the Resume methods, the Resume calls fail (expected) and return an error.

src/generator/models.ts
```diff
@@ -58,6 +58,7 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
   }
   if (needsJSONPopulate) {
     text += 'func populate(m map[string]interface{}, k string, v interface{}) {\n';
+    text += '\tif v == nil {\n\t\treturn\n\t}\n';
     text += '\tif azcore.IsNullValue(v) {\n';
```
catalinaperalta, line 62: Would like to understand why both of these checks are necessary?

src/generator/models.ts
```diff
@@ -58,6 +58,7 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
...
     text += '\t\tm[k] = nil\n';
     text += '\t} else if !reflect.ValueOf(v).IsNil() {\n';
```
catalinaperalta, line 64: Also, is this else inside of the `if v== nil` check?

jhendrixMSFT: No, the `v == nil` check is first.  I probably could condense it like so.

```go
if v == nil {
  return
} else if azcore.IsNullValue(v) {
  tm[k] = nil
} else if !reflect.ValueOf(v).IsNil() {
  tm[k] = v
}
```

src/generator/models.ts
jhendrixMSFT: The `v == nil` check is for types that have an `interface{}` field type(s).  In that case, if the field is `nil` then `v` will also be `nil` which causes `azcore.IsNullValue()` to panic.

test/synapse/2019-06-01/azartifacts/zz_generated_pollers.go
```diff
@@ -22,6 +23,42 @@ type CreateDataFlowDebugSessionResponsePoller interface {
...
+func (p *createDataFlowDebugSessionResponsePoller) Done() bool {
+	return p.pt.Done()
+}
+
+func (p *createDataFlowDebugSessionResponsePoller) Poll(ctx context.Context) (*http.Response, error) {
+	return p.pt.Poll(ctx)
+}
+
+func (p *createDataFlowDebugSessionResponsePoller) FinalResponse(ctx context.Context) (CreateDataFlowDebugSessionResponseResponse, error) {
+	respType := CreateDataFlowDebugSessionResponseResponse{CreateDataFlowDebugSessionResponse: &CreateDataFlowDebugSessionResponse{}}
```
catalinaperalta, line 39: A comment for a separate discussion at a different time...but I dont like the envelope ending with ResponseResponse.

catalinaperalta (approved):
LGTM

test/synapse/2019-06-01/azartifacts/zz_generated_pollers.go
jhendrixMSFT: Yes I agree.  I had brought this up with Jeff last week I think.

Issue #607, opened 2021-04-15
Use custom marshaller/unmarshaller to remove pointer-to-map/slice


jhendrixMSFT:
It adds more complexity to the code generator but improves the user experience.

Sample - https://play.golang.org/p/jZk-0WXUxCB

jhendrixMSFT:
Fixed

PR #615, opened 2021-05-03
Pass maps and slices by value

jhendrixMSFT:
This requires custom marshallers to distinguish between a nil value
(don't send anything) and an empty value (send empty map/slice).
Refactored marshalling detection logic which also fixed some broken
corner-cases.
Added custom XML marshaller to handle nested elements for slices.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -7,7 +7,7 @@ import { Session } from '@autorest/extension-base';
 import { comment } from '@azure-tools/codegen';
 import { CodeModel, ComplexSchema, ConstantSchema, DictionarySchema, GroupProperty, ImplementationLocation, ObjectSchema, Language, Schema, SchemaType, Parameter, Property } from '@autorest/codemodel';
 import { values } from '@azure-tools/linq';
-import { isArraySchema, isObjectSchema, getRelationship, hasAdditionalProperties, hasPolymorphicField, commentLength } from '../common/helpers';
+import { isArraySchema, isObjectSchema, hasAdditionalProperties, hasPolymorphicField, commentLength } from '../common/helpers';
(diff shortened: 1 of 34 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/607

src/generator/models.ts
```diff
@@ -382,6 +382,16 @@ function determineMarshallersForObj(obj: ObjectSchema): Marshallers {
     obj.language.go!.needsPatchMarshaller) {
     needsM = true;
   }
+  // now correct for how the type is used.  if the length is 1 it
+  // means the obj is only input or output.  2 is both, 0 means the
+  // type isn't referenced anywhere so we treat it as 2
+  if (obj.usage?.length === 1) {
```
jhendrixMSFT, line 388: @catalinaperalta this cuts down on the number of marshallers/unmarshallers required to make the code work, i.e. types used as input will only have custom marshallers (no custom unmarshallers), and vice-versa.  However I'm wondering if we should always generate both when any custom marshalling is needed.  E.g. if a type needs custom marshalling and it only has a custom unmarshaller, then you couldn't marshal it into JSON for say logging purposes.  So I'm tempted to remove this change.  What do you think?

catalinaperalta: I think you're right. If it needs custom marshalling...I can see the need for custom unmarshalling arising and given that this code is meant to be used as is for mgmt plane as is I think it's better to provide this than not. For both logging purposes and use by customers in the many many maaaany scenarios that exist out there.

catalinaperalta (approved):
LGTM. Really liked the comments in the marshaller/unmarshaller section

PR #617, opened 2021-05-04
Remove extra indirection

jhendrixMSFT:
json.RawMessage is a type alias for []byte which can implicitly be set
to nil, so the extra indirection was just clutter (no functional changes).
Fixed regeneration script due to change in behavior in Go 1.16.

### DIFF
--- a/.scripts/regeneration.js
+++ b/.scripts/regeneration.js
@@ -118,7 +118,7 @@ function autorestCallback(outputDir, inputFile) {
         // format the output on success
         // print any output or error from go fmt
         if (stderr === '' && error === null) {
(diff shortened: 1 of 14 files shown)


catalinaperalta (approved):
LGTM...do you know why the dependency kept failing to install?

jhendrixMSFT:
Turns out it wasn't failing to install, that's what I thought at first too.  Using `go fmt .` will cause dependencies to be downloaded, and they are written to `stderr` which was causing the step to fail.  Switching to `gofmt` does not trigger dependencies to be downloaded.  Not sure why this started showing up now...

PR #618, opened 2021-05-05
Change Error() to return raw response body

jhendrixMSFT:
It provides the complete error text and helps in the case where
unmarshalling into the target error type fails.
Added disclaimer about the content returned from Error().


test/autorest/additionalpropsgroup/zz_generated_models.go
```diff
@@ -56,23 +55,15 @@ type CatAPTrueResponse struct {
...
-		msg += fmt.Sprintf("Message: %v\n", *e.Message)
-	}
-	if e.Status != nil {
-		msg += fmt.Sprintf("Status: %v\n", *e.Status)
-	}
-	if msg == "" {
-		msg = "missing error info"
-	}
-	return msg
+	return e.raw
```
catalinaperalta, line 66: Wondering whether we'd need to ever let customers know that they could potentially access the error fields on the error type? Should we differentiate between when we fail to unmarshal and not?

catalinaperalta: Just because it's a bit of a waste of unmarshalling if nobody ever uses the fields....why would we keep unmarshalling then? Thinking out loud here...

jhendrixMSFT: You bring up good points.  We need to update the doc comments indicating the error type(s) returned from the APIs so that callers can get to the concrete types.

I think the usefulness of the concrete error types is going to be service and context dependent (e.g. blob storage seems to care about this).

catalinaperalta: I agree, to clarify that the structured info wont be lost. Let me know if you want to add that in this PR.

jhendrixMSFT: Done.

catalinaperalta (approved):
LGTM

PR #622, opened 2021-05-11
Add REQUIRED comment to struct fields

jhendrixMSFT:
For the properties marked as required, add a REQUIRED prefix to the doc
comment.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -132,6 +132,16 @@ class StructDef {
     if (this.Language.errorType) {
       text += '\traw string\n';
     }
+    // group fields by required/optional/read-only in that order
+    this.Properties?.sort((lhs: Property, rhs: Property): number => {
+      if ((lhs.required && !rhs.required) || (!lhs.readOnly && rhs.readOnly)) {
+        return -1;
+      } else if ((rhs.readOnly && !lhs.readOnly) || (!rhs.readOnly && lhs.readOnly)) {
+        return 1;
+      } else {
+        return 0;
+      }
+    });
     for (const prop of values(this.Properties)) {
       if (hasDescription(prop.language.go!)) {
         if (!first) {
--- a/test/autorest/additionalpropsgroup/zz_generated_models.go
+++ b/test/autorest/additionalpropsgroup/zz_generated_models.go
@@ -16,7 +16,6 @@ import (
 
 type CatAPTrue struct {
 	PetAPTrue
-	// Friendly - OPTIONAL
 	Friendly *bool `json:"friendly,omitempty"`
 }
 
@@ -57,12 +56,9 @@ type CatAPTrueResponse struct {
 
 // Implements the error and azcore.HTTPResponse interfaces.
 type Error struct {
-	raw string
-	// Message - OPTIONAL
+	raw     string
 	Message *string `json:"message,omitempty"`
-
-	// Status - OPTIONAL
-	Status *int32 `json:"status,omitempty"`
+	Status  *int32  `json:"status,omitempty"`
 }
 
 // Error implements the error interface for type Error.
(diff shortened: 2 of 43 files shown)


tombuildsstuff:
@jhendrixMSFT perhaps a weird thought, but is it worth grouping these properties into Required and then Optional (and then Read Only) so that it's clearer which are required/optional?

jhendrixMSFT:
Great suggestion, I'll make this change.

src/generator/models.ts
```diff
@@ -132,6 +132,16 @@ class StructDef {
     if (this.Language.errorType) {
       text += '\traw string\n';
     }
+    // group fields by required/optional/read-only in that order
```
catalinaperalta, line 135: Just confirming the below if is also checking in this order?

test/autorest/additionalpropsgroup/zz_generated_models.go
```diff
@@ -96,14 +104,20 @@ type PetAPInPropertiesResponse struct {
...
 	AdditionalProperties map[string]*string
 
-	// Dictionary of
+	// OPTIONAL; Dictionary of
 	AdditionalProperties1 map[string]*float32 `json:"additionalProperties,omitempty"`
-	ID                    *int32              `json:"id,omitempty"`
-	Name                  *string             `json:"name,omitempty"`
-	OdataLocation         *string             `json:"@odata.location,omitempty"`
+
+	// Name - OPTIONAL
```
catalinaperalta, line 119: What's the difference between some having the OPTIONAL/REQUIRED/READ-ONLY at the beginning vs some after the name?  Should this be more homogenous? It looks a little weird that the Optional and required switch places in some cases.

catalinaperalta: Also, unrelated to this PR but the `Dictionary of` is missing something?

src/generator/models.ts
jhendrixMSFT: Correct.  The properties are already lexicographically sorted.  This sort groups them by their required/optional/read-only status.

test/autorest/additionalpropsgroup/zz_generated_models.go
jhendrixMSFT: The `Dictionary of` looks to be an issue with M4.  We don't generate this, nor is it in the swagger.  I'll follow up on that.

The modifier showing up as a suffix is for cases where the property has no doc comment.  We did it like this in track 1 as simply having `// READ-ONLY` would make the linter unhappy.  That no longer appears to be the case.  For consistency, I can remove this.

catalinaperalta: hmm, yeah at least for now, if the linter doesnt complain about it, then I vote for more consistency in the comments.

jhendrixMSFT: Also I think I misspoke about the `Dictionary of` thing, it looks like we're somehow munging the description.  I'll fix that too.

jhendrixMSFT: Fixing the `Dictionary of ` looks to be more involved than I thought.  I've opened https://github.com/Azure/autorest.go/issues/623 to track it.

jhendrixMSFT: The suffix thing has been fixed.

test/autorest/additionalpropsgroup/zz_generated_models.go
```diff
@@ -176,12 +182,14 @@ type PetAPInPropertiesWithAPStringResponse struct {
 }
 
 type PetAPObject struct {
-	// Contains additional key/value pairs not defined in the schema.
+	// REQUIRED
+	ID *int32 `json:"id,omitempty"`
+
+	// OPTIONAL; Contains additional key/value pairs not defined in the schema.
```
jhendrixMSFT, line 188: I intentionally left `OPTIONAL` for additional properties as they are by definition optional.

catalinaperalta (approved):
Not sure if the // REQUIRED comment over a group of fields that dont have a comment over them will ever cause confusion (about whether those subsequent fields are also required), but we can get feedback on that if it becomes a problem.

PR #624, opened 2021-05-13
Move response envelopes to their own file

jhendrixMSFT:
To reduce clutter with API modeled types.  No functional changes, just
moved existing code to other source files.


catalinaperalta (approved):
lgtm

PR #625, opened 2021-05-17
Pass optional map and slice params by value

jhendrixMSFT:
Expand earlier fix to include optional parameters.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/607

PR #627, opened 2021-05-21
Update to latest autorest.core and M4

jhendrixMSFT:
The update brings changes to model Any and AnyObject schema types.  The
former maps to an interface{}, the latter is map[string]interface{}.
Also supported are the security and security-scopes switches which
replace the credential-scope switch.
The telemetry const has been moved to constants.go, which means that
connection.go can be omitted for packages that don't need it (ARM).
Updated test server to v3 which uses recordings instead of a local node
server.  This slightly changes the command used to start the sevice.
Moved autorest.core version to the configuration file and removed some
legacy stuff that's not needed.


catalinaperalta (approved):
LGTM

ArcturusZhang:
> Moved autorest.core version to the configuration file

Does this mean we no longer need the `--version:3.2.1` option in the autorest command?

PR #629, opened 2021-05-25
Add support for method-only parameterized host

jhendrixMSFT:
Added keyvault to the build which uses this pattern.


catalinaperalta (approved):
lgtm

PR #630, opened 2021-05-26
Fix handling of byte arrays

jhendrixMSFT:
Byte arrays in URL format require special marshallers and unmarshallers.
Consolidate duplicate code for JSON marshaller and internal marshallers.
Fixed scope for azkeyvault.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -5,7 +5,7 @@
 
 import { Session } from '@autorest/extension-base';
 import { comment } from '@azure-tools/codegen';
-import { CodeModel, ComplexSchema, DictionarySchema, GroupProperty, ObjectSchema, Language, SchemaType, Parameter, Property } from '@autorest/codemodel';
+import { ByteArraySchema, CodeModel, ComplexSchema, DictionarySchema, GroupProperty, ObjectSchema, Language, SchemaType, Parameter, Property } from '@autorest/codemodel';
 import { values } from '@azure-tools/linq';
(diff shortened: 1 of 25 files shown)


src/generator/models.ts
```diff
@@ -62,6 +66,17 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
     text += '\t}\n';
     text += '}\n\n';
   }
+  if (needsJSONPopulateByteArray) {
+    text += 'func populateByteArray(m map[string]interface{}, k string, b []byte, f azcore.Base64Encoding) {\n';
+    text += '\tif len(b) == 0 {\n';
+    text += '\t\treturn\n';
+    text += '\t} else if azcore.IsNullValue(b) {\n';
+    text += '\t\tm[k] = nil\n';
+    text += '\t} else if !reflect.ValueOf(b).IsNil() {\n';
```
catalinaperalta, line 75: Just checking do we need to do the IsNullValue and IsNil with those azcore methods for a byte array just like we did for the interfaces?

jhendrixMSFT: D'oh I think you're right, this can be simplified.  I also need to test the JSON `null` scenario too.

jhendrixMSFT: Fixed with test.  We do need the `IsNullValue()` to handle the JSON null scenario.

catalinaperalta (approved):
LGTM

PR #631, opened 2021-06-08
Walk the list for supported security schemes

jhendrixMSFT:
Throw a helpful error message one wasn't found.

### DIFF
--- a/src/generator/connection.ts
+++ b/src/generator/connection.ts
@@ -26,17 +26,22 @@ export async function generateConnection(session: Session<CodeModel>): Promise<s
   const content = generateContent(session, imports);
   text += imports.text();
   if (session.model.security.authenticationRequired) {
-    // we only support AADToken scheme at present
-    const scheme = session.model.security.schemes[0];
-    if (scheme.type !== 'AADToken') {
-      throw new Error(`unsupported security scheme ${scheme.type}`);
+    let tokenScheme: AADTokenSecurityScheme | undefined;
+    for (const scheme of values(session.model.security.schemes)) {
+      // we only support AADToken scheme at present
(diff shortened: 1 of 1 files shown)


src/generator/connection.ts
```diff
@@ -26,17 +26,22 @@ export async function generateConnection(session: Session<CodeModel>): Promise<s
   const content = generateContent(session, imports);
   text += imports.text();
   if (session.model.security.authenticationRequired) {
-    // we only support AADToken scheme at present
-    const scheme = session.model.security.schemes[0];
-    if (scheme.type !== 'AADToken') {
-      throw new Error(`unsupported security scheme ${scheme.type}`);
+    let tokenScheme: AADTokenSecurityScheme | undefined;
+    for (const scheme of values(session.model.security.schemes)) {
```
catalinaperalta, line 30: Trying to understand, if for any reason there are multiple schemes is it ok to simply accept the right one and ignore the others?

jhendrixMSFT: We will likely need to support both.  The problem here is we don't have an abstraction in `azcore` that works with key credential types.  Until we do, we can't support anything other than `AADToken`.

catalinaperalta: Hmmm I see, yeah makes sense. Ok, although should we have something somewhere (anywhere 😝) that tells people this? Like when people are generating so that they know other schemes arent going to be automatically included? Or would they just know that from looking at our code?

jhendrixMSFT: I added an entry to the wiki about this.  In addition, the code generator will throw an error describing that only `AADToken` is supported.

catalinaperalta (approved):
LGTM!

PR #635, opened 2021-06-15
handle mixed environments

jhendrixMSFT:


catalinaperalta (approved):
LGTM. Just two question:
1. Why didnt `sep` work?
2. I cant remember if there would ever be a case where the package name ends in a /?

jhendrixMSFT:
If there's a path in a config file like `/foo/bar/baz` and you run this on Windows, it won't trim off the `baz` for the package name.  By default we author everything in  config files using `/` as it's accepted on all OS

PR #641, opened 2021-06-28
Fix intersection of arrays/additional properties and time

jhendrixMSFT:
Ensure properly marshalling and unmarshalling of custom time formats
when in a slice or map.

### DIFF
--- a/src/transform/transform.ts
+++ b/src/transform/transform.ts
@@ -41,7 +41,7 @@ async function process(session: Session<CodeModel>) {
   // this must happen before processing objects as we depend on the
   // schema type being an actual Go type.
   for (const dictionary of values(session.model.schemas.dictionaries)) {
-    dictionary.elementType.language.go!.name = schemaTypeToGoType(session.model, dictionary.elementType, false);
+    dictionary.elementType.language.go!.name = schemaTypeToGoType(session.model, dictionary.elementType, true);
     dictionary.language.go!.elementIsPtr = !isTypePassedByValue(dictionary.elementType);
     if (dictionary.language.go!.description) {
       dictionary.language.go!.description = parseComments(dictionary.language.go!.description);
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -417,6 +417,13 @@ function generateJSONMarshallerBody(obj: ObjectSchema, structDef: StructDef): st
         base64Format = 'URL';
       }
       marshaller += `\tpopulateByteArray(objectMap, "${prop.serializedName}", ${receiver}.${prop.language.go!.name}, azcore.Base64${base64Format}Format)\n`;
+    } else if (isArraySchema(prop.schema) && prop.schema.elementType.language.go!.internalTimeType) {
+      const source = `${receiver}.${prop.language.go!.name}`;
+      marshaller += `\taux := make([]*${prop.schema.elementType.language.go!.internalTimeType}, len(${source}), len(${source}))\n`;
+      marshaller += `\tfor i := 0; i < len(${source}); i++ {\n`;
+      marshaller += `\t\taux[i] = (*${prop.schema.elementType.language.go!.internalTimeType})(${source}[i])\n`;
+      marshaller += '\t}\n';
+      marshaller += `\tpopulate(objectMap, "${prop.serializedName}", aux)\n`;
     } else {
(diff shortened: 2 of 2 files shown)


jhendrixMSFT:
Unfortunately we don't have any tests in the test server that cover this case.  I had to hack an existing swagger to test the changes.  Snippets from the diff.

```
 type DatetimeWrapper struct {
+	// OPTIONAL; Contains additional key/value pairs not defined in the schema.
+	AdditionalProperties map[string]*time.Time
 	Field *time.Time `json:"field,omitempty"`
 	Now   *time.Time `json:"now,omitempty"`
+
+	// List of schedule times for backup
+	Schedules []*time.Time `json:"schedules,omitempty"`
 }
...
+	aux := make([]*timeRFC3339, len(d.Schedules), len(d.Schedules))
+	for i := 0; i < len(d.Schedules); i++ {
+		aux[i] = (*timeRFC3339)(d.Schedules[i])
+	}
+	populate(objectMap, "schedules", aux)
+	if d.AdditionalProperties != nil {
+		for key, val := range d.AdditionalProperties {
+			objectMap[key] = (*timeRFC3339)(val)
+		}
+	}
...
+		case "schedules":
+			var aux []*timeRFC3339
+			err = unpopulate(val, &aux)
+			for _, au := range aux {
+				d.Schedules = append(d.Schedules, (*time.Time)(au))
+			}
+			delete(rawMsg, key)
+		default:
+			if d.AdditionalProperties == nil {
+				d.AdditionalProperties = map[string]*time.Time{}
+			}
+			if val != nil {
+				var aux timeRFC3339
+				err = json.Unmarshal(val, &aux)
+				d.AdditionalProperties[key] = (*time.Time)(&aux)
+			}
+			delete(rawMsg, key)

**Comment by jhendrixMSFT** [comment]
Fixes https://github.com/Azure/autorest.go/issues/640

File: src/transform/transform.ts [hunk 1]
```diff
@@ -41,7 +41,7 @@ async function process(session: Session<CodeModel>) {
   // this must happen before processing objects as we depend on the
   // schema type being an actual Go type.
   for (const dictionary of values(session.model.schemas.dictionaries)) {
-    dictionary.elementType.language.go!.name = schemaTypeToGoType(session.model, dictionary.elementType, false);
+    dictionary.elementType.language.go!.name = schemaTypeToGoType(session.model, dictionary.elementType, true);
```
catalinaperalta, line 44: Trying to understand what case we weren't covering before so that this switched to true....were we generating incorrectly for all in body params? Or are there cases where this should be false?

jhendrixMSFT: The `inBody` param indicates if the type is passed via HTTP body or not (the "not" being a header).  When this is `true`, we add some extra annotations to the schema which is required for the intersection of additional properties and custom time times.
However, I think this parameter could possibly be removed as I don't think it matters if the extra annotations are there if we don't use them.

jhendrixMSFT: Removing the param will erroneously add custom time types where they aren't needed, so it will remain for now.

catalinaperalta: Ah got it, it's for additional props. Thanks!

catalinaperalta (approved):
LGTM. Sad there isnt a case to cover this in the testserver, maybe we should open an issue to add a test for it?

jhendrixMSFT:
I was thinking about this.  Test holes like these are pretty unique to Go, given the way our marshallers/unmarshallers work (I believe there are other, similar corner-cases).  We could add them to the test server, however I'm wondering if it would be better if we had our own, generator-specific swagger to test all of our odd cases.  Thoughts?

catalinaperalta:
> I was thinking about this. Test holes like these are pretty unique to Go, given the way our marshallers/unmarshallers work (I believe there are other, similar corner-cases). We could add them to the test server, however I'm wondering if it would be better if we had our own, generator-specific swagger to test all of our odd cases. Thoughts?

Yeah I agree. I think it would help us to maintain a consistent set of test cases for cases that are specifically important for Go to test. I think we've run across these a few times since developing the track2 generator, so having our own generator specific swagger would be a great idea.

jhendrixMSFT:
Opened https://github.com/Azure/autorest.go/issues/643

PR #682, opened 2021-08-24
Don't assume exported connection type in method generation

jhendrixMSFT:
Updated the condition to match that when generating the connection.
This fixes an issue where the connection wasn't exported but the method
group clients assumed it to be.


catalinaperalta (approved):
LGTM

PR #683, opened 2021-08-26
Update codegen based on recent azcore refactor

jhendrixMSFT:
Rename package prefixes based on the refactor.
Updated telemetry policy codegen to reflect the recent changes.
Changed telemetry const generation per policy changes.
Updated tests to use azcore/to.


test/autorest/azurespecialsgroup/zz_generated_constants.go
```diff
@@ -9,6 +9,6 @@
 package azurespecialsgroup
 
 const (
-	telemetryInfo = "azsdk-go-azurespecialsgroup/" + version
-	version       = "v0.1"
+	module  = "azurespecialsgroup"
```
RickWinter, line 12: Should this be azspecialsgroup for the package?

test/autorest/bytegroup/zz_generated_constants.go
```diff
@@ -9,6 +9,6 @@
 package bytegroup
 
 const (
-	telemetryInfo = "azsdk-go-bytegroup/" + version
-	version       = "v0.1"
+	module  = "bytegroup"
```
RickWinter, line 12: Should these have a package name that makes them unique to tests?  aztestbytegroup, etc...?

test/synapse/2019-06-01/azspark/zz_generated_constants.go
```diff
@@ -9,8 +9,8 @@
 package azspark
 
 const (
-	telemetryInfo = "azsdk-go-azspark/" + version
-	version       = "v0.1"
+	module  = "azspark"
+	version = "v0.1"
```
RickWinter, line 13: We should standardize our versions v0.1.0 keeping the Major.Minor.Rev

RickWinter (approved):

test/autorest/bytegroup/zz_generated_constants.go
jhendrixMSFT: We kept the package naming consistent with track 1 so it was easier (at least in the beginning) to use track 1 as a baseline.  While we could go and change it, it's a bit of busy work that doesn't bring much value IMO.

test/synapse/2019-06-01/azspark/zz_generated_constants.go
jhendrixMSFT: fixed

Issue #686, opened 2021-08-27
Include HTTP status code for failed LROs on initial response


jhendrixMSFT:
For reference.

https://github.com/Azure/azure-sdk-for-go/blob/main/services/cosmos-db/mgmt/2021-06-15/documentdb/mongodbresources.go#L370

We should be able to safely call `result.Response()` to get the raw `*http.Response` so that the status code can be included in the error.  In the event that there was no initial response, the return value is `nil` which `autorest.NewErrorWithError()` can tolerate.

While it's an anti-pattern to touch `result` if `err` is not `nil`, we've already violated this elsewhere for futures and have to support it.

jhendrixMSFT:
Fixed in 2.1.187 of the generator.

PR #687, opened 2021-08-30
Return the future's HTTP response if available

jhendrixMSFT:
This ensures that the DetailedError is populated with the HTTP response
if the initial LRO response returns a failed status code.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/686

ArcturusZhang (approved):

Issue #690, opened 2021-09-08
Missing type conversion for query params that are enums


jhendrixMSFT:
In log analytics.

https://github.com/Azure/azure-sdk-for-go/pull/15344

```go
	if options != nil && options.Continents != nil {
		for _, qv := range options.Continents {
			reqQP.Add("continents", qv)
		}
	}
```
Here, `qv` is an enum so it  needs a type conversion to `string`.  We also need to handle the case where enums are not strings.

jhendrixMSFT:
Fixed in preview.28

PR #691, opened 2021-09-09
Handle non-string exploded query params

jhendrixMSFT:
Exploded query params were assumed to be of type string which can lead
to bad codegen for cases where the query param element is either a const
or a non-string type.  Perform the necessary type conversion.

### DIFF
--- a/test/maps/azalias/zz_generated_alias_client.go
+++ b/test/maps/azalias/zz_generated_alias_client.go
@@ -11,6 +11,7 @@ package azalias
 import (
 	"context"
 	"errors"
+	"fmt"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"net/http"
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/690

test/maps/azalias/zz_generated_alias_client.go
```diff
@@ -129,6 +135,11 @@ func (client *aliasClient) listCreateRequest(ctx context.Context, options *Alias
 	}
 	reqQP := req.Raw().URL.Query()
 	reqQP.Set("api-version", "2.0")
+	if options != nil && options.GroupBy != nil {
+		for _, qv := range options.GroupBy {
+			reqQP.Add("groupBy", string(qv))
```
jhendrixMSFT, line 140: This is an example of the fix.  Before this you'd get `reqQP.Add("groupBy", qv)` which wouldn't compile.

chlowell (approved):

PR #692, opened 2021-09-10
feat: upgrade modelerfour version

tadelesh:
Upgrade modelerfour version to 4.21.0 to supports x-ms-original-file which will be used in test generation.


jhendrixMSFT (changes requested):
I looked into migrating to 4.20.0 earlier this week however it introduces various codegen changes which need to be understood first.

tadelesh:
> I looked into migrating to 4.20.0 earlier this week however it introduces various codegen changes which need to be understood first.

Closed as codegen changes needed.

PR #708, opened 2021-10-15
Add support for polymorphic additionalProperties

tcnghia:
When `additionalProperties: { "$ref": "/some/PolymorphicType" }` is used the resulted struct field will be `map[string]PolymorphicType`, but we are still using regular parsing helpers. 

This PR adds polymorphic handling for such case.


tcnghia:
@jhendrixMSFT  I've added code gen test and also some tests for the new polymorphic deserialization.

Issue #717, opened 2021-10-23
Refactor data plane constructors


chlowell:
- [x] upgrade to azcore v0.20.0
- [ ] remove Connection, merging it into the client
- [ ] new client constructor should take a pipeline

So the simple case, a client with no host params, would look like this:
```go
type FooClient struct {
    ep string
    pl runtime.Pipeline
}

func NewFooClient(endpoint string, pipeline runtime.Pipeline) *FooClient
```

jhendrixMSFT:
Fixed in upcoming `preview.32` release.

PR #723, opened 2021-10-29
Remove Connection type

jhendrixMSFT:
This brings all of the constructors into alignment.
Data-plane client constructors now simply take any host or other client
parameters and a pipeline parameter.
Vanilla ARM client constructors are similar to data-plane, however they
take an azcore.ClientOptions parameter instead of a pipeline.
The parameter logic was ported from the now-deleted connection.ts file
with some slight modifications to simplify it.
Emit a constant for swaggers that specify a host.
Endpoint was renamed to host in places where it made sense.
The client's pipeline field is accessed directly now.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/717

test/autorest/custombaseurlgroup/zz_generated_paths_client.go
```diff
@@ -20,12 +21,24 @@ import (
...
 type PathsClient struct {
-	con *Connection
+	host string
+	pl   runtime.Pipeline
 }
 
 // NewPathsClient creates a new instance of PathsClient with the specified values.
-func NewPathsClient(con *Connection) *PathsClient {
-	return &PathsClient{con: con}
+func NewPathsClient(host *string, options *azcore.ClientOptions) *PathsClient {
```
jhendrixMSFT, line 29: This is an oddity I wanted to discuss.  The swagger defines this as required, however it has a default value, so we consider it optional.  This is how we used to generate the Connection params, however I wonder if in this case we should actually generate a `PathClientOptions` that embeds the `azcore.ClientOptions` and contains the `host` param.  It's more work in the code generator, but I feel like it's the correct thing to do.  Thoughts?

test/autorest/custombaseurlgroup/zz_generated_paths_client.go
```diff
@@ -20,12 +21,30 @@ import (
 // PathsClient contains the methods for the Paths group.
 // Don't use this type directly, use NewPathsClient() instead.
 type PathsClient struct {
-	con *Connection
+	host string
+	pl   runtime.Pipeline
+}
+
+// PathsClientOptions contains the optional parameters for NewPathsClient.
+type PathsClientOptions struct {
```
jhendrixMSFT, line 29: New client options struct.

test/autorest/custombaseurlgroup/zz_generated_paths_client.go
jhendrixMSFT: Moved optional params to an options struct for non-data plane scenarios.

chlowell:
Why do we have two flavors of ARM?

chlowell:
This may be a bigger question than this PR, but do we need to distinguish `arm` from `data-plane`? Naively, it seems like we could reasonably generate the same API for both.

test/consumption/2019-10-01/armconsumption/zz_generated_credits_client.go
```diff
@@ -36,7 +36,11 @@ func NewCreditsClient(credential azcore.TokenCredential, options *arm.ClientOpti
 	if len(cp.Host) == 0 {
 		cp.Host = arm.AzurePublicCloud
 	}
-	return &CreditsClient{ep: string(cp.Host), pl: armruntime.NewPipeline(module, version, credential, &cp)}
+	client := &CreditsClient{
+		host: string(cp.Host),
```
chlowell, line 40: Ugh, reading this has me thinking the `arm.ClientOptions` field shouldn't be "Host" because the value includes the scheme 😩

jhendrixMSFT: Good point, it probably should have been `HostURL`.  Given the last round of pain from breaking changes, I don't think we want to fix it now.

jhendrixMSFT:
There is somewhat of an unstated requirement that the "vanilla" generated code should be useable, and we see this play out in the autorest test server tests.  If we were to treat vanilla the same as data-plane, we'd need to write client constructors that handle client options and create a pipeline.

We also have some customers (internal and external) that use our code generator on their non-Azure swaggers, so we'd be impacting their work.  I don't think we need to go to great lengths to support this scenario, but changing the vanilla clients so they no longer work as-is seems like a step backwards.

chlowell:
You mean, asking consumers of vanilla clients to create a pipeline is too high a bar? I'm just looking at these thinking either could reasonably converge with the other:

```go
// arm ("vanilla")
func NewPetsClient(options *azcore.ClientOptions)

// data-plane
func NewKeyVaultClient(pl runtime.Pipeline)
```

jhendrixMSFT:
You are right, we could conceivably replace the vanilla client (as you cited above) with the data-plane variant which would simplify the code generator a bit.  I've opened https://github.com/Azure/autorest.go/issues/724 to track this.

chlowell (approved):
Thanks for doing this!

Issue #699, opened 2021-09-23
Generated methods reference missing client fields


chlowell:
Track 2 data plane clients generated from [this Service Bus spec](
https://raw.githubusercontent.com/Azure/azure-rest-api-specs/sb_dataplane_namespace/specification/servicebus/data-plane/servicebus-swagger.json) lack `apiVersion` fields referenced by methods:

```go
type entityClient struct {
	con *connection
}

func (client *entityClient) deleteCreateRequest(...) (*policy.Request, error) {
	...
	if client.apiVersion != nil {
		reqQP.Set("api-version", "2017_04")
	}
	...
}
```

jhendrixMSFT:
I looked into this, we have a weird corner-case here.

This is similar to single-value enums that are `x-ms-enum modelAsString:false`.  The value is considered a constant by M4, so it should generate no parameter and have its value be embedded in the code.

Where it gets weird is for constants that are marked as `required:false`.  What this means is that the value doesn't _have_ to be sent, but if it is, it can _only_ be the constant value.

You can see an instance of this correctly working for method params [here](https://github.com/Azure/autorest.go/blob/track2/test/compute/2019-12-01/armcompute/zz_generated_dedicatedhosts_client.go#L247-L249).  Unfortunately, we don't properly handle this for client params.

jhendrixMSFT:
Fixed in upcoming `preview.32`

PR #725, opened 2021-11-03
Fix codegen for optional client constant parameters

jhendrixMSFT:
Apply the same pattern we use for optional constant method parameters.

### DIFF
--- a/test/maps/azalias/zz_generated_alias_client.go
+++ b/test/maps/azalias/zz_generated_alias_client.go
@@ -19,12 +19,13 @@ import (
 )
 
 type aliasClient struct {
-	endpoint string
-	pl       runtime.Pipeline
+	endpoint   string
+	apiVersion *string
+	pl         runtime.Pipeline
 }
 
 // newAliasClient creates a new instance of aliasClient with the specified values.
-func newAliasClient(geography *Geography, pl runtime.Pipeline) *aliasClient {
+func newAliasClient(geography *Geography, apiVersion *string, pl runtime.Pipeline) *aliasClient {
 	hostURL := "https://{geography}.atlas.microsoft.com"
 	if geography == nil {
 		defaultValue := GeographyUs
@@ -32,8 +33,9 @@ func newAliasClient(geography *Geography, pl runtime.Pipeline) *aliasClient {
 	}
 	hostURL = strings.ReplaceAll(hostURL, "{geography}", string(*geography))
 	client := &aliasClient{
-		endpoint: hostURL,
-		pl:       pl,
+		endpoint:   hostURL,
+		apiVersion: apiVersion,
+		pl:         pl,
 	}
 	return client
 }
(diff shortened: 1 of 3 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/699

test/maps/azalias/zz_generated_alias_client.go
```diff
@@ -19,21 +19,23 @@ import (
...
-	endpoint string
-	pl       runtime.Pipeline
+	endpoint   string
+	apiVersion *string
+	pl         runtime.Pipeline
 }
 
 // newAliasClient creates a new instance of aliasClient with the specified values.
-func newAliasClient(geography *Geography, pl runtime.Pipeline) *aliasClient {
+func newAliasClient(geography *Geography, apiVersion *string, pl runtime.Pipeline) *aliasClient {
```
jhendrixMSFT, line 28: I will say this is clunky as-is.  Since it behaves more like a boolean, I was thinking this should really look like `includeAPIVersion bool` instead.

chlowell: That does look better, so long as there's only one version. I wonder whether we really want a client parameter for this case though. If I understand correctly, in this case the swagger specifies one value for api-version but doesn't declare that param required. That implies the client needn't set api-version but if it does, there's only one valid value. It seems odd to make api-version configurable then--there's no reason to configure it. Should the client just use the known value?

jhendrixMSFT: I agree that the behavior doesn't really make sense and is likely a bug in the swagger.  However, I don't think we want to have the code generator treat any params as "special" and bypass the authoring.  I believe some track 1 code generators did this with `api-version` which became problematic for some resource manager APIs that also have a parameter named `api-version`.

jhendrixMSFT: Here's the description from the swagger which makes the intent a bit more clear

| "Specifies the version of the REST protocol used for processing the request. This is required when using shared key authorization."

So perhaps it's not a bug, however the API is a bit clunky (I suppose a hand-written convenience layer would simplify this).

This does make me realize that we aren't emitting the doc comments for global parameters though.

chlowell: I see that the swagger doesn't constrain `api-version` values, so strictly speaking we don't know that only one valid value exists (unless that's formally implied somehow; I don't know much about swagger syntax). Given that, I think we do want `*string` here.

It's my experience updating azblob's generated code that has me looking sideways at single-value parameters. Passing method parameters with only one acceptable value was awkward (I haven't checked whether the blob swagger constrains the parameters we generate "enums" for). I couldn't help but think, "there's only one value and you know what it is, why do I have to pass it to you" 😆

chlowell (approved):

test/maps/azalias/zz_generated_alias_client.go
jhendrixMSFT: For the linked service bus swagger, the value comes from the `default` clause.

PR #734, opened 2021-11-29
Use IsZero instead of IsNil in populate function

jhendrixMSFT:
This is to handle the case where a field is the interface type.  The
change is safe as all fields are either pointer-to-type or implicitly
nilable.


.scripts/regeneration.js
```diff
@@ -98,8 +98,8 @@ generate("azstorage", blobStorage, 'test/storage/2020-06-12/azblob', '--security
 const network = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/228cf296647f6e41182cee7d1a403990e6a8fe3c/specification/network/resource-manager/readme.md';
 generateFromReadme("armnetwork", network, 'package-2020-03', 'test/network/2020-03-01/armnetwork', '--module=armnetwork --azure-arm=true');
 
-const compute = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/228cf296647f6e41182cee7d1a403990e6a8fe3c/specification/compute/resource-manager/readme.md';
-generateFromReadme("armcompute", compute, 'package-2019-12-01', 'test/compute/2019-12-01/armcompute', '--module=armcompute --azure-arm=true');
+//const compute = 'https://raw.githubusercontent.com/Azure/azure-rest-api-specs/228cf296647f6e41182cee7d1a403990e6a8fe3c/specification/compute/resource-manager/readme.md';
```
jhendrixMSFT, line 101: Temporary until https://github.com/Azure/autorest/pull/4381 is merged and a new release is published.

src/generator/models.ts
```diff
@@ -61,7 +61,7 @@ export async function generateModels(session: Session<CodeModel>): Promise<strin
     text += '\t\treturn\n';
     text += '\t} else if azcore.IsNullValue(v) {\n';
     text += '\t\tm[k] = nil\n';
-    text += '\t} else if !reflect.ValueOf(v).IsNil() {\n';
+    text += '\t} else if !reflect.ValueOf(v).IsZero() {\n';
```
jhendrixMSFT, line 64: This is the change you'll see in the codegen.

test/swagger/alias.json
```diff
@@ -430,6 +430,31 @@
           "$ref": "#/definitions/GeoJsonFeatureData"
         }
       ]
+    },
+    "ScheduleCreateOrUpdateProperties": {
```
jhendrixMSFT, line 434: This is used to generate a type that's been fixed so I can write a test for it.

seankane-msft:
Is there a situation where we have a non-nillable type that would not be represented on the response? I believe gen code normally uses a *string instead of a string for this reason but that's the only situation I can think of where this breaks down.

jhendrixMSFT:
Kind of.  This was discovered in Microsoft.Automation RP, when a field is an object thus we generate an `interface{}`.  If you set this to something like `5`, the `IsNil()` will panic since `5` can't be nil.

I recreated this in the updated swagger, it generates the following type.

```go
type ScheduleCreateOrUpdateProperties struct {
	// A list of all the previously created aliases.
	Aliases []*string `json:"aliases,omitempty"`

	// Gets or sets the description of the schedule.
	Description *string `json:"description,omitempty"`

	// Gets or sets the interval of the schedule.
	Interval interface{} `json:"interval,omitempty"`

	// Gets or sets the start time of the schedule.
	StartTime *time.Time `json:"startTime,omitempty"`
}
```

I then wrote the round-trip test to marshall/unmarshall to test this code path.

src/generator/models.ts
tadelesh: @jhendrixMSFT , I don't think IsZero is a good way to be used for empty judgement here. If we set the param with false/0/"", this param will be ignored with unmarshall which may lose some useful info.

jhendrixMSFT: Good catch.  This is because we're looking at the value, not the type.  I'll have an update later today.

jhendrixMSFT: Fixed

seankane-msft (approved):
LGTM

seankane-msft:
> Kind of. This was discovered in Microsoft.Automation RP, when a field is an object thus we generate an `interface{}`. If you set this to something like `5`, the `IsNil()` will panic since `5` can't be nil.

I thought when a field was an object we generated `map[string]interface{}`, is that a change or am I misremembering? 
In Tables, the entity is an object in the swagger [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cosmos-db/data-plane/Microsoft.Tables/preview/2019-02-02/table.json#L1689-L1705) and when it is generated it is a `map[string]interface{}` [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/data/aztables/internal/models.go#L236-L243).

Issue #738, opened 2021-11-30
Unmarshall  of LRO operation response problem


tadelesh:
If an operation is LRO and the response is a polymorphism type directly, the unmarshall type passed into the `PollUntilDone` function is an inner `xxxClassification`, not the `xxxResult` which will cause failure.
The [swagger](https://github.com/Azure/azure-rest-api-specs/blob/37d690709ce4639bd0aafe7f4a5f5856f2bef624/specification/azure-kusto/resource-manager/Microsoft.Kusto/stable/2021-08-27/kusto.json#L3150-L3155) which just shows the LRO response type. The generated code for operation: `DataConnections_CreateOrUpdate` will fail when poll result. It seems `&respType.DataConnectionClassification` should be `&respType.DataConnectionCreateOrUpdateResult`.

jhendrixMSFT:
Fixed in preview.32

PR #739, opened 2021-12-01
Unmarshal polymorphic LRO final response into result envelope

jhendrixMSFT:
The result envelope contains the customer unmarshaller for its
polymorphic return type.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/738

seankane-msft (approved):

PR #741, opened 2021-12-02
Update dev and runtime dependencies

jhendrixMSFT:
No functional changes.


src/autorest-configuration.md
```diff
@@ -8,9 +8,9 @@ AutoRest needs the below config to pick this up as a plug-in - see https://githu
 
 # Pipeline Configuration
 ``` yaml
-version: 3.6.2
+version: 3.7.2
```
jhendrixMSFT, line 11: This update fixes the issue we had with generating `armcompute`.

src/generator/operations.ts
```diff
@@ -651,7 +651,7 @@ function createProtocolRequest(group: OperationGroup, op: Operation, imports: Im
       text += `\turlPath = strings.ReplaceAll(urlPath, "{${pp.language.go!.serializedName}}", ${paramValue})\n`;
     }
   }
-  text += `\treq, err := runtime.NewRequest(ctx, http.Method${(<string>op.requests![0].protocol.http!.method).capitalize()}, ${hostParam})\n`;
+  text += `\treq, err := runtime.NewRequest(ctx, http.Method${capitalize(<string>op.requests![0].protocol.http!.method)}, ${hostParam})\n`;
```
seankane-msft, line 654: in other places the `<string>` gets pulled off, why is it necessary here? (for my own curiosity)

seankane-msft (approved):

src/generator/operations.ts
jhendrixMSFT: Good catch, these casts are also now unnecessary, I've removed them.

Issue #732, opened 2021-11-17
Aggregate types that are composed via allOf


jhendrixMSFT:
Both examples should generate the following type.

```go
type ExtendedProductProperties struct {
    Version *string
}
```

Example 1
```json
    "ExtendedProductProperties": {
      "description": "Product information.",
      "type": "object",
      "properties": {},
      "allOf": [
        {
          "$ref": "#/definitions/VirtualMachineExtensionProductProperties"
        },
        {
          "$ref": "#/definitions/VirtualMachineProductProperties"
        }
      ]
    },
    "VirtualMachineExtensionProductProperties": {
      "description": "Product information.",
      "type": "object",
      "properties": {
        "version": {
          "description": "Specifies product version.",
          "type": "string",
          "readOnly": true
        }
      }
    },
    "VirtualMachineProductProperties": {
      "description": "Product information.",
      "type": "object",
      "properties": {
        "version": {
          "description": "Specifies product version.",
          "type": "string",
          "readOnly": true
        }
      }
    },
```
Example 2
```json
    "ExtendedProductProperties": {
      "description": "Product information.",
      "type": "object",
      "properties": {
        "version": {
          "description": "Specifies product version.",
          "type": "string",
          "readOnly": true
        }
      },
      "allOf": [
        {
          "$ref": "#/definitions/VirtualMachineProductProperties"
        }
      ]
    },
    "VirtualMachineProductProperties": {
      "description": "Product information.",
      "type": "object",
      "properties": {
        "version": {
          "description": "Specifies product version.",
          "type": "string",
          "readOnly": true
        }
      }
    },
```

jhendrixMSFT:
In the cases where the type for `version` is mismatched, an error must be thrown.

jhendrixMSFT:
Fixed in preview.33

Issue #709, opened 2021-10-18
Cleanup: only generate required polymorphic helpers


jhendrixMSFT:
We generate singular/map/etc polymorphic helpers regardless of what ones are actually required.  No functional impact, just code bloat.

jhendrixMSFT:
Fixed in preview.33

PR #744, opened 2021-12-07
Aggregrate properties for composite allOf types

jhendrixMSFT:
Don't anonymously embed composite types, promote their properties
into the containing type.  This avoids unnecessary levels of nesting
and represents what's sent over the wire.  It also simplifies the
handling of custom marshallers and unmarshallers.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/732

src/transform/transform.ts
```diff
@@ -72,15 +72,22 @@ async function process(session: Session<CodeModel>) {
...
-          const asObj = <ObjectSchema>child;
-          let discValue = getEnumForDiscriminatorValue(asObj.discriminatorValue!, discriminatorEnums);
+        if (obj.discriminatorValue) {
+          let discValue = getEnumForDiscriminatorValue(obj.discriminatorValue!, discriminatorEnums);
           if (!discValue) {
             // this happens when there are intermediate types in the hierarchy that aren't discriminators.
             // e.g. BaseType -> GenericProps -> DerivedType
             // here, GenericProps is just a collection of standard properties common to various objects.
             // M4 gives it a discriminatorValue of "GenericProps" that won't exist in the enum defined
             // in BaseType.  unclear if this is desirable, but for now we just embed that value directly.
```
jhendrixMSFT, line 82: Turns out this is a bug in M4.  We will be able to remove this in a future version.

jhendrixMSFT: M4 has been fixed and incorporated.

jhendrixMSFT:
Also fixes https://github.com/Azure/autorest.go/issues/709

seankane-msft (approved):
This removes embedded structs, will this be a breaking change for management plane?

jhendrixMSFT:
It will be.  However, we think the change is worth it as it simplifies the usage and is the same that we generate in track 1 which will ease migration.

seankane-msft:
> It will be. However, we think the change is worth it as it simplifies the usage and is the same that we generate in track 1 which will ease migration.

I would agree with you, in that case LGTM

PR #746, opened 2021-12-14
Skip duplicate entry when aggregating allOf

jhendrixMSFT:
Some cases M4 does this for us.  In this case, we do it.
Added test to ensure coverage.
Bump version for next release.

### DIFF
--- a/src/generator/models.ts
+++ b/src/generator/models.ts
@@ -239,6 +239,8 @@ function aggregateProperties(obj: ObjectSchema): Array<Property> {
             const msg = `type ${obj.language.go!.name} contains duplicate property ${exists.language.go!.name} with mismatched types`;
             throw new Error(msg);
           }
+          // don't add the duplicate
+          continue;
         }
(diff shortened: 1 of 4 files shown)


seankane-msft (commented):
When are there duplicates ? Is there an example of how this changes generated code ?

jhendrixMSFT:
There are two cases for duplicates.  First is parent-child which M4 handles for us.  These already exist in various swaggers.  Second is siblings which M4 does not handle.  I added this in alias.json; `GeoJsonObject` and `GeoJsonFeatureData` now both have a duplicate `id` property.

jhendrixMSFT:
As to how it changes the generated code.  Without this fix, `GeoJsonFeature` ended up with two fields `ID *string` which is broken.

seankane-msft:
Ahhh gotcha. Thanks for the explanation, LGTM.

seankane-msft (approved):

Issue #730, opened 2021-11-09
Stop generating modeled error types


jhendrixMSFT:
The Azure contract for error responses is the HTTP status code and an error code, usually provided via `x-ms-error-code` in the response, however in some cases it's in the body.  Any content in the body is for extra context but is subject to change.  Given this, we should stop generating models for the error types specified in swagger.

jhendrixMSFT:
Ok this will work something like this.

`azcore` will contain an `ErrorResponse` interface (maybe a type, will see).

```go
type ErrorResponse interface {
  ErrorCode() string
  StatusCode() int
  RawResponse() *http.Response
}
```

We will stop generating error models.  Instead, we'll call into `azcore` to return an error that conforms to the `ErrorResponse` interface.

The `Error` method on the error will return rich information, basically the same info that you'd see for request/response logging.

jhendrixMSFT:
Some M4 bugs prevent emitting error types in some cases.

jhendrixMSFT:
Fixed in upcoming preview.36

ArcturusZhang:
Hey @jhendrixMSFT in some circumstances, the error definitions in swagger might cause duplicate schema errors, since they are recursive types (at least one property is of its own type). I reported this to the modeler here: https://github.com/Azure/autorest/issues/4397

Do you think this change could resolve this? (I think not)

jhendrixMSFT:
No but it shouldn't be a problem.  I believe we already have such generated code.  Or are you saying it's still an issue?

jhendrixMSFT:
This was backed out in https://github.com/Azure/autorest.go/pull/773

PR #747, opened 2021-12-17
Update codegen with azcore breaking changes

jhendrixMSFT:
Modeled error types are no longer used so the Error() method has been
removed along with the raw field.  In addition the *HandleError()
methods for API have also been removed.
LROs no longer require an error handler func during construction.
The arguments to NewPipeline() and NewResponseError() have slightly
changed.
The client endpoint field in the options struct was renamed to Endpoint.
Removal of error types is pending M4 fixes however the plubming is in
place.


jhendrixMSFT:
Partially fixes https://github.com/Azure/autorest.go/issues/730.  The modeled errors are no longer emitted as error types or used, however M4 bugs prevent removal of them.

src/generator/operations.ts
```diff
@@ -946,15 +942,15 @@ function generateResponseUnmarshaller(op: Operation, response: SchemaResponse, u
     // use the designated time type for unmarshalling
     unmarshallerText += `\tvar aux *${response.schema.language.go!.internalTimeType}\n`;
     unmarshallerText += `\tif err := runtime.UnmarshalAs${getMediaType(response.protocol)}(resp, &aux); err != nil {\n`;
-    unmarshallerText += `\t\treturn ${zeroValue}, runtime.NewResponseError(err, resp)\n`;
+    unmarshallerText += `\t\treturn ${zeroValue}, runtime.NewResponseError(resp)\n`;
```
jhendrixMSFT, line 945: This is incorrect as the unmarshalling error is lost.

jhendrixMSFT: This has been fixed.

src/transform/transform.ts
```diff
@@ -518,65 +517,51 @@ function processOperationResponses(session: Session<CodeModel>) {
...
-        const schemaError = (<SchemaResponse>ex).schema;
-        if (isObjectSchema(schemaError)) {
-          for (const prop of values(schemaError.properties)) {
-            // adding the Inner prefix on error types, since errors in Go have an Error() method
-            // in order to implement the error interface. This causes errors to not be able to have
-            // an Error field as well, since it would cause confusion
-            if (prop.language.go!.name === 'Error') {
-              prop.language.go!.name = 'Inner' + prop.language.go!.name;
+      // TODO: make this work once relevant M4 bugs have been fixed
+      // annotate all exception types as errors so we can skip generating them
```
chlowell, line 521: Do you want to link the issues here?

chlowell (approved):
Does M4 also prevent a half measure like unexporting the unwanted types?

src/transform/transform.ts
jhendrixMSFT: I haven't dug into this fully yet and opened them as Tim was OOF at the time.  I'm tracking this overall work in https://github.com/Azure/autorest.go/issues/730

jhendrixMSFT:
There's also https://github.com/Azure/autorest.go/issues/757 which talks about dropping unused types.

PR #751, opened 2022-01-03
Fix reading of text responses

jhendrixMSFT:
For operations that return text media type, read the response and
populate the Value field in the result envelope.


test/autorest/mediatypesgroup/zz_generated_mediatypes_client.go
```diff
@@ -278,6 +278,12 @@ func (client *MediaTypesClient) binaryBodyWithThreeContentTypesCreateRequest(ctx
 // binaryBodyWithThreeContentTypesHandleResponse handles the BinaryBodyWithThreeContentTypes response.
 func (client *MediaTypesClient) binaryBodyWithThreeContentTypesHandleResponse(resp *http.Response) (MediaTypesClientBinaryBodyWithThreeContentTypesResponse, error) {
 	result := MediaTypesClientBinaryBodyWithThreeContentTypesResponse{RawResponse: resp}
+	body, err := runtime.Payload(resp)
```
jhendrixMSFT, line 281: I think there's a larger issue here with multi media-types.  I've opened https://github.com/Azure/autorest.go/issues/752 to track this.

chlowell (approved):

Issue #750, opened 2021-12-22
Cannot read property 'parameters' of undefined


Alancere:
### What happened?
track2 generator policyinsights issue：
```
error   | PreCheck/DuplicateSchema | Duplicate Schema named 'ErrorResponse' (1 differences):
  - properties.error.$ref: "#/components/schemas/schemas:65" => "#/components/schemas/schemas:298"; This error can be *temporarily* avoided by using the 'modelerfour.lenient-model-deduplication' setting.  NOTE: This setting will be removed in a future version of @autorest/modelerfour; schemas should be updated to fix this issue sooner than that.
error   | PreCheck/DuplicateSchema | Duplicate Schema named 'ErrorResponse' (1 differences):
  - properties.error.$ref: "#/components/schemas/schemas:65" => "#/components/schemas/schemas:357"; This error can be *temporarily* avoided by using the 'modelerfour.lenient-model-deduplication' setting.  NOTE: This setting will be removed in a future version of @autorest/modelerfour; schemas should be updated to fix this issue sooner than that.
error   | PreCheck/DuplicateSchema | Duplicate Schema named 'ErrorDefinition' (1 differences):
  - properties.details.$ref: "#/components/schemas/schemas:69" => "#/components/schemas/schemas:302"; This error can be *temporarily* avoided by using the 'modelerfour.lenient-model-deduplication' setting.  NOTE: This setting will be removed in a future version of @autorest/modelerfour; schemas should be updated to fix this issue sooner than that.
error   | PreCheck/DuplicateSchema | Duplicate Schema named 'ErrorDefinition' (1 differences):
  - properties.details.$ref: "#/components/schemas/schemas:69" => "#/components/schemas/schemas:361"; This error can be *temporarily* avoided by using the 'modelerfour.lenient-model-deduplication' setting.  NOTE: This setting will be removed in a future version of @autorest/modelerfour; schemas should be updated to fix this issue sooner than that.
```
When I use the directive modelerfour.lenient-model-deduplication, the following issue occurs again:
```
warning | PreNamer/DeduplicateName | Deduplicating schema name: 'ErrorDefinition' -> 'ErrorDefinitionAutoGenerated'
warning | PreNamer/DeduplicateName | Deduplicating schema name: 'ErrorDefinition' -> 'ErrorDefinitionAutoGenerated2'
warning | PreNamer/DeduplicateName | Deduplicating schema name: 'ErrorResponse' -> 'ErrorResponseAutoGenerated'
[...]

Issue #748, opened 2021-12-20
Throw exception when stuttering collision for models


tadelesh:
When doing stuttering, if collision happens for two models, hope to throw exception about the two collision models' name instead of leave it to compiling stage.

jhendrixMSFT:
Fixed in upcoming preview.35

Issue #731, opened 2021-11-12
Decide how to handle operation group name collisions


jhendrixMSFT:
In cases where swagger mixes named/unnamed operation groups, we calculate a name for the unnamed group but don't ensure it doesn't collide with an existing name.  This can lead to bad codegen if the names collide.  We can either.

- Have the generator try to be smart and avoid collisions
- Throw an error, forcing either the swagger to be updated or an autorest directive be used to fix the name

jhendrixMSFT:
Fixed in upcoming preview.35

PR #753, opened 2022-01-04
Fix codegen failure for obscure paging corner-case

jhendrixMSFT:
Predicate using the next page operation on the existence of
nextLinkOperation instead of member as it might be removed when the
underlying operation has been removed via a transform.
Throw an Error if stuttering type name fix-up causes a collision.
Throw an Error if the fallback operation group name causes a collision.
Replaced some deprecated methods with their corresponding updates.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -511,7 +511,8 @@ function generateOperation(op: Operation, imports: ImportManager): string {
     text += `\t\tadvancer: func(ctx context.Context, resp ${getResponseEnvelopeName(op)}) (*policy.Request, error) {\n`;
     const nextLink = op.language.go!.paging.nextLinkName;
     const response = getResultFieldName(op);
-    if (op.language.go!.paging.member) {
+    // nextLinkOperation might be absent in some cases, see https://github.com/Azure/autorest/issues/4393
+    if (op.language.go!.paging.nextLinkOperation) {
       const nextOpParams = getCreateRequestParametersSig(op.language.go!.paging.nextLinkOperation).split(',');
       // keep the parameter names from the name/type tuples and find nextLink param
       for (let i = 0; i < nextOpParams.length; ++i) {
--- a/src/transform/namer.ts
+++ b/src/transform/namer.ts
@@ -99,7 +99,7 @@ export async function namer(session: Session<CodeModel>) {
     structNames.add(obj.language.go!.name);
   }
 
-  // fix stuttering type names, object fields, and other bits
+  // fix stuttering type names
   for (const obj of values(model.schemas.objects)) {
     const details = <Language>obj.language.go;
     const originalName = details.name;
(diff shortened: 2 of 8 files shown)


jhendrixMSFT:
- https://github.com/Azure/autorest.go/issues/731
- https://github.com/Azure/autorest.go/issues/748
- https://github.com/Azure/autorest.go/issues/750

src/generator/operations.ts
```diff
@@ -511,7 +511,8 @@ function generateOperation(op: Operation, imports: ImportManager): string {
     text += `\t\tadvancer: func(ctx context.Context, resp ${getResponseEnvelopeName(op)}) (*policy.Request, error) {\n`;
     const nextLink = op.language.go!.paging.nextLinkName;
     const response = getResultFieldName(op);
-    if (op.language.go!.paging.member) {
+    // nextLinkOperation might be absent in some cases, see https://github.com/Azure/autorest/issues/4393
+    if (op.language.go!.paging.nextLinkOperation) {
```
jhendrixMSFT, line 515: Paging fix.

test/maps/azalias/zz_generated_models.go
```diff
@@ -137,7 +137,7 @@ type GeoJSONFeatureData struct {
 // GeoJSONObjectClassification provides polymorphic access to related types.
 // Call the interface's GetGeoJSONObject() method to access the common type.
 // Use a type switch to determine the concrete type.  The possible types are:
-// - *GeoJSONObject, *GeoJsonFeature
+// - *GeoJSONFeature, *GeoJSONObject
```
jhendrixMSFT, line 140: For discriminated types, we were generating the comment before the type names were fixed up.  So this is an added bonus.

jhendrixMSFT: I further refined this fix to include names that were changed due to stuttering.

src/transform/namer.ts
```diff
@@ -92,10 +92,27 @@ export async function namer(session: Session<CodeModel>) {
...
+    structNames.add(obj.language.go!.name);
+  }
+
+  // fix stuttering type names
+  for (const obj of values(model.schemas.objects)) {
+    const details = <Language>obj.language.go;
+    const originalName = details.name;
+    details.name = trimPackagePrefix(stutteringPrefix, originalName);
+    // if the type was renamed to remove stuttering, check if it collides with an existing type name
+    if (details.name !== originalName && structNames.has(details.name)) {
```
chlowell, line 108: I guess this prevents combining the two new loops, because in principal we need to generate all the struct names before we can be certain none collide?

jhendrixMSFT: Sadly yes.  It's also why some of the code comments had the old type names.

chlowell (approved):

Issue #756, opened 2022-01-10
Logic for additional properties: dictionary type with `object` type value


tadelesh:
We found many swagger examples (e.g. [swagger](https://github.com/Azure/azure-rest-api-specs/blob/d8242e83bad59e45d5bb1dad769a185504f5fde3/specification/appplatform/resource-manager/Microsoft.AppPlatform/stable/2020-07-01/appplatform.json#L3543-L3549), [example](https://github.com/Azure/azure-rest-api-specs/blob/d8242e83bad59e45d5bb1dad769a185504f5fde3/specification/appplatform/resource-manager/Microsoft.AppPlatform/stable/2020-07-01/examples/Bindings_CreateOrUpdate.json#L7-L10)) set value with primary type like string or integer for the `dictionary` type with `object` type value. The latest codegen will set such dictionary type with map[string][string]interface{} as m4 model it with any-object. It will cause marshal/unmarshall failure. The root cause is the wrong swagger definition. But other language's SDK use special logic for such type's code generation. We should decide how we can handle such issues in GO SDK: 1. change codegen logic or 2. use directive to rewrite all the wrong definition and then fix swagger one by one.

jhendrixMSFT:
The current thinking is that we'll require the service teams to fix their affected swagger.  Will update as more info is available.

jhendrixMSFT:
Fixed in upcoming preview.36 release.

PR #759, opened 2022-01-13
Default to Any instead of AnyObject

jhendrixMSFT:
Too many broken specs depended on this track 1 behavior.
Moved generator mod definition under the autorest dir where it should
have been from the start.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/756

chlowell (approved):

Issue #757, opened 2022-01-11
Skip orphan models


ArcturusZhang:
Since we are currently aggregate the models together instead of using anonymous properties to mimic inheritance, the models that work only for being inherited will be never referenced in our SDK.
For instance, a model `VirtualMachine` will inherit from model `Resource`, and in generated code, all the properties of Resource will be added to VirtualMachine, and it no longer holds a reference to the model `Resource`.

Therefore could we skip the generation of those orphan models? This could reduce the size of our SDK, and increase the stability of the SDK.

jhendrixMSFT:
There's no way at present to tell if a model is unused.  See https://github.com/Azure/autorest/issues/3664 for M4 feature request to make this happen.

ArcturusZhang:
Hi @jhendrixMSFT what does it mean when you said this is "partially resolved" by #760?
I assume this issue contains two parts:
1. Some schemas are not used in the swagger
2. Some schemas are used in the swagger, but their generated types are no longer used in the SDK. This might happen since SDK is doing some kind of optimization.

Which part does #760 resolve? If it is resolving part 2, I believe this issue is good to merge. The unused schemas in the swagger should be issues of the swaggers instead of something we should resolve during SDK generation.

jhendrixMSFT:
Unfortunately, it's more complicated than that.  M4 defines a type's usage as `input`, `output`, and `exception`.  Note that these can be logically OR'ed together, so a type can have all three usages applied to it.
A type's usage can also be empty, which means that the type isn't used.  This is common for discriminated types, as the base type is referenced but the child types are not.

The "partially resolves" for this issue is the case where a type's usage is empty, and it's not a discriminated type.  You can see an example of this in file `test/autorest/azurespecialsgroup/zz_generated_models.go` as part of #760 (i.e. the removal of the `ODataFilter` type).

We don't have a way, at present, to handle the case where a type is used in an `allOf` but not actually returned (the `Resource` example in ARM).

jhendrixMSFT:
It's also been pointed out that some specs will define various unreferenced types as part of emulating `oneOf` behavior.

jhendrixMSFT:
Given that we can't reliably determine if a model is really orphaned, we won't be making this change.  The namespace pollution is very minimal and hasn't been a problem in track 1 which has the same issue.

PR #760, opened 2022-01-18
Omit unused model types from code generation

jhendrixMSFT:
Slightly tweaked test swagger so that needed models are referenced.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/730
Partially addresses https://github.com/Azure/autorest.go/issues/757

chlowell (approved):

PR #762, opened 2022-01-20
Clean up some linter issues

jhendrixMSFT:
Removed redundant capacity parameter when calling make().
Omit time helper funcs when they aren't needed.

### DIFF
--- a/src/generator/time.ts
+++ b/src/generator/time.ts
@@ -32,7 +32,7 @@ export async function generateTimeHelpers(session: Session<CodeModel>): Promise<
   }
   let needsPopulate = false;
   for (const obj of values(session.model.schemas.objects)) {
-    if (obj.language.go!.marshallingFormat === 'xml') {
+    if (obj.language.go!.marshallingFormat !== 'json') {
       // population helpers are for JSON only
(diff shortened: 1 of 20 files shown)


src/generator/time.ts
```diff
@@ -21,33 +23,57 @@ export class Content {
...
   const content = new Array<Content>();
+  if (!session.model.language.go!.hasTimeRFC1123 &&
+    !session.model.language.go!.hasTimeRFC3339 &&
+    !session.model.language.go!.hasUnixTime &&
+    !session.model.language.go!.hasDate) {
+    return content;
+  }
+  let needsPopulate = false;
+  for (const obj of values(session.model.schemas.objects)) {
+    if (obj.language.go!.marshallingFormat === 'xml') {
```
chlowell, line 35: Nitpicking:
```suggestion
    if (obj.language.go!.marshallingFormat !== 'json') {
```

chlowell (approved):

src/generator/time.ts
jhendrixMSFT: Fixed

PR #763, opened 2022-01-20
Split custom marshallers and unmarshallers into their own file

jhendrixMSFT:
It just clutters up the models file and is uninteresting to customers.


jhendrixMSFT:
@chlowell I meant to ask you if you think this is an improvement.  Outside of people working on codegen, I don't think that the marshallers/unmarshallers are particularly interesting and I find myself having to scroll past them when I just want to look at types.  I could be convinced though that having them together like we do now is valuable though.

CC @ArcturusZhang

chlowell:
I could go either way as well but lean toward doing it. You've done enough maintenance here to see the benefit. I expect users to prefer the package docs to the code, especially when they just want to see models, and therefore not care about the separation. I wouldn't want to separate un/marshaling from the rest of a handwritten type, for the sake of maintenance, but in this case maintenance is running the generator again and inspecting the diff, so I'm less worried about introducing a bug by forgetting to make corresponding changes in two files.

PR #767, opened 2022-01-28
Generate pager APIs for single-page operations

jhendrixMSFT:
When a pager's nextLinkName is null, it means that the API returns a
single page now but might return multiple in the future which shouldn't
be considered a breaking change.

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -505,7 +505,7 @@ function generateOperation(op: Operation, imports: ImportManager): string {
     text += `\t\trequester: func(ctx context.Context) (*policy.Request, error) {\n`;
     text += `\t\t\treturn client.${info.protocolNaming.requestMethod}(${reqParams})\n`;
     text += '\t\t},\n';
-    // this is no advancer for single-page pagers
+    // there is no advancer for single-page pagers
     if (op.language.go!.paging.nextLinkName) {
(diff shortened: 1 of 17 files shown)


src/generator/operations.ts
```diff
@@ -505,27 +505,30 @@ function generateOperation(op: Operation, imports: ImportManager): string {
...
-      const nextOpParams = getCreateRequestParametersSig(op.language.go!.paging.nextLinkOperation).split(',');
-      // keep the parameter names from the name/type tuples and find nextLink param
-      for (let i = 0; i < nextOpParams.length; ++i) {
-        const paramName = nextOpParams[i].trim().split(' ')[0];
-        const paramType = nextOpParams[i].trim().split(' ')[1];
-        if (paramName.startsWith('next') && paramType === 'string') {
-          nextOpParams[i] = `*resp.${response}.${nextLink}`;
-        } else {
-          nextOpParams[i] = paramName;
+    // this is no advancer for single-page pagers
```
chlowell, line 508: True, but an odd turn of phrase. Do you mean
```suggestion
    // there is no advancer for single-page pagers
```

chlowell (approved):

src/generator/operations.ts
jhendrixMSFT: Yes good catch, updated.

PR #772, opened 2022-02-07
Fold result types into response types

jhendrixMSFT:
Due to inability to anonymously embed a generic type parameter, we will
continue to generate API-specific response types.
The result types have been removed, their fields promoted to their
matching response types.


chlowell (approved):

jhendrixMSFT:
There is some debate on this approach, so needs a bit more discussion first.

Issue #771, opened 2022-02-07
Omit unused model will cause model misssing


tadelesh:
When using preview.36 codegen, some models will missing due to the new `omit unused model` feature. Swagger is [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/datafactory/resource-manager/readme.go.md). Struct `TypeConversionSettings` will missing.

jhendrixMSFT:
This feature has only been partially implemented, specifically just for error types.  Also, I can't find anything with the name `TypeConversionSettings` in the linked swagger.  Can you please clarify?

tadelesh:
@jhendrixMSFT `TypeConversionSettings` is defined in a separated [file](https://github.com/Azure/azure-rest-api-specs/blob/6a6928f0212c20965ea1faf538cbec8ca6e2e31e/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/entityTypes/Pipeline.json#L5138-L5167) and is used by `TabularTranslator` defined [here](https://github.com/Azure/azure-rest-api-specs/blob/6a6928f0212c20965ea1faf538cbec8ca6e2e31e/specification/datafactory/resource-manager/Microsoft.DataFactory/stable/2018-06-01/entityTypes/Pipeline.json#L5099-L5137). All the two models will be generated with v35. But `TypeConversionSettings` is missing with v36.

jhendrixMSFT:
Thanks for clarifying.  This type is a field in `TabularTranslator` which is a discriminated type.  While we have logic to preserve them, we need to propagate that to all child types.

jhendrixMSFT:
Fixed in upcoming preview.38 release.

PR #773, opened 2022-02-08
Revert exclusion of unused models

jhendrixMSFT:
Due to limitations of OpenAPI, some cases can't be accurately described
like support for oneOf.  As a result, we can't reliably determine if a
model type really isn't used as some specs emulate the oneOf behavior.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/771

chlowell:
The few of these models I checked truly aren't used by their respective clients. Naively, it seems like ModelerFour could know a type is never used, but I take it that's impossible in general because ModelerFour is restricted to OpenAPI semantics? How can we be certain an error type is unused?

jhendrixMSFT:
After the `oneOf` conversation I've been wondering the same thing.  While I suppose it seems unlikely that `oneOf` would be used for the error case, there's no reason why it couldn't be.  Makes me think that we should back out that portion of the change as well.
And frankly, we're talking about maybe a handful of types to omit out of dozens, if not more.  I don't think anybody will notice these vestigial types (they exist in track 1 and nobody has complained about them).

chlowell:
Right, the impact is namespace pollution and wasted bytes. Worst case scenario is that one of these types disappears in a future version after being deleted from the swagger (which seems like the right fix) but an application somewhere took a dependency on that type despite not being able to use it with the associated client--seems pretty unlikely 😆

It would be nice for autorest to identify unused types though, if only to warn about swagger cruft.

chlowell (approved):

Issue #778, opened 2022-02-16
Lack of `strings` input in some case


tadelesh:
I happened to meet an compiling issue when generating sdk code with this [swagger](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/authorization/resource-manager/readme.md). It seems we need to add `strings` to the `imports` before this [line](https://github.com/Azure/autorest.go/blob/2a1bcf3b856e580849ca731efa4c9b2fcdc0965a/src/generator/operations.ts#L799).

jhendrixMSFT:
Fixed in upcoming preview.37

PR #779, opened 2022-02-16
Add missing strings import

jhendrixMSFT:
Fixes case when the import isn't added elsewhere.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/778

chlowell (approved):

PR #783, opened 2022-02-25
Upgrade autorest.testserver

chlowell:
A few small fixes I made while investigating test coverage. Upgrading autorest.testserver fixes a bug responsible for underestimating our coverage. Looks like this PR is doomed to fail CI because the new version adds routes (I guess this is why the testserver version is pinned), creating a harmless diff in the regeneration step. Is there a workaround for that or do we just merge a red PR when upgrading like this?


jhendrixMSFT:
You need to run `rush regenerate` to rebuild the tests.

chlowell:
🤦 thanks for reminding me!

jhendrixMSFT (approved):

PR #790, opened 2022-03-11
Add more LRO test coverage

chlowell:
Adding a few new tests and unskipping ones that pass.

### DIFF
--- a/test/autorest/lrogroup/lros_test.go
+++ b/test/autorest/lrogroup/lros_test.go
@@ -15,7 +15,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
-	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


test/autorest/lrogroup/lros_test.go
```diff
@@ -236,6 +236,40 @@ func TestLROBeginDeleteProvisioning202Deletingcanceled200(t *testing.T) {
...
+	require.NoError(t, err)
+	res, err := resp.PollUntilDone(context.Background(), time.Second)
+	require.NoError(t, err)
+	if r := cmp.Diff(res.Product, Product{
+		ID:   to.StringPtr("/lro/patch/201/retry/onlyAsyncHeader"),
+		Name: to.StringPtr("foo"),
+		Properties: &ProductProperties{
+			ProvisioningState: to.StringPtr("Succeeded"),
+		},
+	}); r != "" {
```
jhendrixMSFT, line 251: Do you think we should use something like `require.Equal()` here instead?  I know we have this lingering pattern in the other tests, I just haven't been able to update them yet (it might be a difficult thing to do).  Minor issue.

test/autorest/lrogroup/lros_test.go
```diff
@@ -776,7 +827,6 @@ func TestLROBeginPutAsyncSubResource(t *testing.T) {
 }
 
 func TestLROBeginPutNoHeaderInRetry(t *testing.T) {
-	t.Skip("problem with put flow")
```
jhendrixMSFT, line 779: Cool that these now work.  Presumably it's due to the LRO rewrite from last year.

jhendrixMSFT (approved):

test/autorest/lrogroup/lros_test.go
chlowell: May as well. I wrote this before you rolled out `require` and only did the easy transforms after my rebase. The old pattern is regular enough that replacing it wasn't so bad (I did the rest of the file).

chlowell:
Planning to merge this after upgrading autorest.testserver (https://github.com/Azure/autorest.testserver/pull/369)

jhendrixMSFT:
@chlowell can we merge this while we sort out the issue in the test server?

PR #791, opened 2022-03-16
Use generic pager and poller types from azcore

jhendrixMSFT:
Stop synthesizing per-method pager and poller types.
Add an unmarshaller to LRO response envelopes that return a non-embedded
result so that the JSON is propagated to the correct field.
Select Go version before code generation as it's dependent on it.
Execute modtidy after regeneration.


tadelesh:
It seems request options for LRO will miss the `ResumeToken` definition if there is a header param in request. ([swagger example](https://github.com/Azure/azure-rest-api-specs/blob/a3be4ceeb6654b64bf9250cbf3903b3e3246dde7/specification/search/resource-manager/Microsoft.Search/stable/2020-08-01/search.json#L932-L934))

jhendrixMSFT:
Also, what if an optional param was added in the future that doesn't belong to any existing params group?  That would result in a breaking change.

tadelesh:
> @tadelesh good catch. This is due to parameter grouping. As you can see, this parameter belongs to group `search-management-request-options`.
> 
> If the swagger defines custom parameter groups, then we [mistakenly](https://github.com/Azure/autorest.go/blob/track2/src/transform/transform.ts#L443) skip synthesizing the optional params groups. I believe the original idea was that if the swagger defines a custom parameter group with one or more optional params, that we'd use that as the "optional params bucket" instead of creating our own. Clearly we didn't actually do this, however I'm wondering if it's actually a good idea.
> 
> The pro I see is that it's one less type/param for the API. The con is that we add our own param (params in the future?) to this type. Also, if there are multiple, optional, param groups, which one gets the resume token?
> 
> Given that param groups are somewhat uncommon, and the above ambiguity, I'm leaning towards synthesizing the optional params group type. I could be convinced otherwise though, let me know your thoughts.
> 
> CC @chlowell

I agree with your decision. Separating the user defined param group with our SDK level option is truly more clear and also won't cause any mix up situation if swagger change.

jhendrixMSFT:
@tadelesh I've rebased the changes on latest `track2` branch so the fix for optional params group is available.  Can you please resume your testing?

tadelesh:
@jhendrixMSFT I've finished the first round of testing and all RPs' mgmt. plane SDK can be successfully generated and compiled. If no further SDK signature change, I'll change the testgen and do the full cover mock test. As there is not much change for the request and response handling, I think everything will work fine.

src/generator/gomod.ts
```diff
@@ -13,11 +13,11 @@ export async function generateGoModFile(session: Session<CodeModel>): Promise<st
     return '';
   }
   let text = `module ${modName}\n\n`;
-  text += 'go 1.16\n\n';
+  text += 'go 1.18\n\n';
   // here we specify the minimum version of azcore as required by the code generator.
   // the version can be overwritten by passing the --azcore-version switch during generation.
-  const version = await session.getValue('azcore-version', '0.21.0');
-  if (!version.match(/^\d+\.\d+\.\d+$/) && !version.match(/^\d+\.\d+\.\d+-beta\.\d+$/)) {
+  const version = await session.getValue('azcore-version', '0.22.1-0.20220315231014-ed309e73db6b');
```
jhendrixMSFT, line 19: Waiting for official release.

chlowell (approved):

PR #792, opened 2022-03-16
Add azcore-version optional command line switch

jhendrixMSFT:


jhendrixMSFT:
We're automatically rolling forward to autorest core 3.8.0 even though I thought it was pinned in the config file.  Following up on this, would prefer to fix the other things separately.

chlowell (approved):

PR #802, opened 2022-04-12
add version to the reserved word list

ArcturusZhang:
Related issue: https://github.com/Azure/azure-sdk-for-go/issues/17470

This RP has an enum type of `Version`, which is conflicting with the built-in method `Version()` and causing compiling errors.

### DIFF
--- a/src/CodeNamerGo.cs
+++ b/src/CodeNamerGo.cs
@@ -242,7 +242,7 @@ namespace AutoRest.Go
                     "autorest", "client", "date", "err", "req", "resp", "result", "sender", "to", "validation", "m", "v", "k", "objectMap",
 
                     // reserved method names
-                    "Send"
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
I believe the test is failing due to a newer version of TypeScript being used.  I tried moving to an older version but was unsuccessful.  If you can't figure out a way to do this, feel free to skip the test for now (we can fix it later).

ArcturusZhang:
Hi @jhendrixMSFT I just locally verified that this fix works, but with a side effect that now every property `Version` will be renamed to `VersionProperty`.
I am finding a way to only change the type names

ArcturusZhang:
I pushed a new iteration - now we only take the `Version` and `UserAgent` when escaping names of types

PR #806, opened 2022-04-20
Fix codegen with transforms

jhendrixMSFT:
The emitter was taking the union of go-protocol and go/text-transform as
its outputs.  If a transform was used which resulted in shortening the
code, you'd end up with trailing garbage characters.
Added test to cover this case.


test/databoxedge/2021-02-01/armdataboxedge/zz_generated_addons_client.go
```diff
@@ -95,9 +95,6 @@ func (client *AddonsClient) createOrUpdate(ctx context.Context, deviceName strin
 // createOrUpdateCreateRequest creates the CreateOrUpdate request.
 func (client *AddonsClient) createOrUpdateCreateRequest(ctx context.Context, deviceName string, roleName string, addonName string, resourceGroupName string, addon AddonClassification, options *AddonsClientBeginCreateOrUpdateOptions) (*policy.Request, error) {
 	urlPath := "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.DataBoxEdge/dataBoxEdgeDevices/{deviceName}/roles/{roleName}/addons/{addonName}"
-	if deviceName == "" {
```
chlowell, line 98: Is this arbitrary or is an empty path segment somehow okay here?

chlowell (approved):

test/databoxedge/2021-02-01/armdataboxedge/zz_generated_addons_client.go
jhendrixMSFT: This is just arbitrary to test the change.  I wanted to pick something that was sure to decrease the length of the codegen.

chlowell: Could you comment about that in the markdown? I imagine the transform will puzzle future generations otherwise

jhendrixMSFT: Good point, fixed.

PR #807, opened 2022-04-21
Add a per-method doc comment with the API version

jhendrixMSFT:


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17679

src/generator/operations.ts
```diff
@@ -587,6 +587,17 @@ function emitPagerDefinition(op: Operation, imports: ImportManager): string {
...
 
+function genApiVersionDoc(apiVersions?: ApiVersions): string {
+  if (!apiVersions) {
+    return '';
+  }
+  const versions = new Array<string>();
+  apiVersions.forEach((val) => {
+    versions.push(val.version);
+  })
+  return `// Uses API version ${versions.join(',')}\n`;
```
RickWinter, line 598: Is there defensive coding or are there really instances multiple versions are the basis for code gen?

jhendrixMSFT: I asked @timotheeguerin about when we'd have more than one API version for an operation.

RickWinter: nit, "//  Generated from API version ${versions.join(',')}, See {hyperlink to swagger}\n"

Might be a lot of redundant text

jhendrixMSFT: I don't know that we have the swagger information available.

I can change the text as specified above if you think that's better.

chlowell (approved):

RickWinter (approved):

PR #809, opened 2022-04-22
Use LRO machinery from runtime instead of arm/runtime

jhendrixMSFT:
Fixed renaming of AzurePublicCloud to AzurePublic.
Fixed renaming of PageProcessor to PagingHandler.
Update to released version of azcore.


jhendrixMSFT:
Pending a release of `azcore`, but all the breaking changes are in place.

RickWinter (approved):

Issue #818, opened 2022-05-09
Error handling for HEAD operation


tadelesh:
For now, if HEAD operation returns 200-300 response, `result.success` will be `true`, otherwise `false`, regardless of service error like `405 Method Not Allowed`. It will cause mis-leading of `false` return. I think we should use error return to handle some HTTP error code.

jhendrixMSFT:
In response to https://github.com/Azure/azure-sdk-for-go/issues/17809

PR #819, opened 2022-05-09
Fix for head as boolean

jhendrixMSFT:
Use the list of specified HTTP status codes from the swagger instead of
a hard-coded 4xx list for false.  This was masking some endpoints that
don't support HEAD and were returning a 405.

### DIFF
--- a/test/autorest/azurespecialsgroup/zz_generated_header_client.go
+++ b/test/autorest/azurespecialsgroup/zz_generated_header_client.go
@@ -86,6 +86,9 @@ func (client *HeaderClient) CustomNamedRequestIDHead(ctx context.Context, fooCli
 	if err != nil {
 		return HeaderClientCustomNamedRequestIDHeadResponse{}, err
 	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusNotFound) {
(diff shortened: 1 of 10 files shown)


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/818

test/autorest/azurespecialsgroup/zz_generated_header_client.go
```diff
@@ -86,6 +86,9 @@ func (client *HeaderClient) CustomNamedRequestIDHead(ctx context.Context, fooCli
 	if err != nil {
 		return HeaderClientCustomNamedRequestIDHeadResponse{}, err
 	}
+	if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusNotFound) {
+		return HeaderClientCustomNamedRequestIDHeadResponse{}, runtime.NewResponseError(resp)
```
RickWinter, line 90: Nit, HeaderCleintCustomRequestIDHeaderResponse... that's a mouthful :)

jhendrixMSFT: This comes from the autorest test server swagger, the names aren't all that great.

Issue #797, opened 2022-03-30
Don't generate a JSON marshaler for output only types


jhendrixMSFT:
https://github.com/Azure/azure-sdk-for-go/blob/e4f7f8464a2ab317d3bf2f0e3caf584f1f63632e/sdk/resourcemanager/containerservice/armcontainerservice/zz_generated_models.go#L2078

This type is used for output so it should never need a JSON marshaler.

RickWinter:
This is an optimization and not required for Stable release

jhendrixMSFT:
Fixed in upcoming `preview.41`

PR #820, opened 2022-05-09
Omit marshaller for output-only types

jhendrixMSFT:


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/797

PR #821, opened 2022-05-10
Avoid canonicalizing HTTP headers

jhendrixMSFT:
Some services (e.g. storage) use HTTP headers for metadata and should be
case-preserving.  Calling Header.Set canonicalizes the HTTP header name
to avoid relying on specific casing which breaks some storage scenarios.


jhendrixMSFT:
Fixes https://github.com/Azure/azure-sdk-for-go/issues/17850

RickWinter (approved):

chlowell (approved):

PR #824, opened 2022-05-16
Prep for 4.0.0 release

jhendrixMSFT:


jhendrixMSFT:
This is identical to `preview.41` which is our GA release.

RickWinter:
Do we need to do this?  As long as generated code doesn't break, we can still change internals

jhendrixMSFT:
IMO yes as it signals our code generator is baked and that we won't be introducing any breaking changes to the generated code.

chlowell (approved):

Issue #825, opened 2022-05-18
Stop generating module name and version constants for data-plane


chlowell:
`ModuleName` and `ModuleVersion` aren't needed anymore because data-plane clients don't create their own pipelines. We can stop generating these constants and leave user-agent configuration up to handwritten code. Doing this in a non-breaking fashion might take some thought. I suppose we could make `module-version` optional?

RickWinter:
I see there is a fix for this.  I'm wondering though if we really want this behavior.  Having the generator handle it would make it less likely that individuals get creative in naming or forget to write the two values.

RickWinter:
Good points.  I don't want to be hand editing generated code.  We could have the version.go file have only these two constants and require that to be written by hand (or process) seperate from the autorest generator?

jhendrixMSFT:
Yeah we can do that.

jhendrixMSFT:
Fixed in upcoming `v4.0.0` release.

PR #826, opened 2022-05-18
Remove moduleName and moduleVersion constants for data-plane

jhendrixMSFT:
They are no longer used by the generated code.  SDK authors must
maintain their own constants of these values which is better than
editing generated code anyways.
Added 'rush buildvet' command to 'go build' and 'go vet' the things.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/825

chlowell (approved):

PR #827, opened 2022-05-24
Export response types when clients aren't

jhendrixMSFT:
Response types were inadvertently not being exported when exporting
clients is disabled.

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/go",
-  "version": "4.0.0",
+  "version": "4.0.0-preview.42",
   "description": "AutoRest Go Generator",
   "main": "dist/exports.js",
   "typings": "dist/exports.d.ts",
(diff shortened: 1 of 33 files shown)


chlowell:
I feel like I'm missing something obvious here--why would someone want to export these types but not the client?

chlowell:
I feel like I'm missing something obvious here--why would someone want to export these types but not the client?

jhendrixMSFT:
For data-plane, the default is to not export the clients but export everything else.  The idea was that there would be hand-written clients that leverage the internal, generated ones, and all of the other types would be exported as-is.

The response types were being inadvertently not exported due to the clients being renamed before they were synthesized.

chlowell:
Ah, I see. A handwritten client would be able to use the generated one provided they were in the same package, and the generated models would be exported to applications. Would we really ship a module like that? Sounds like DPG 😄

chlowell (approved):

jhendrixMSFT:
It's currently what we're doing for `azblob` with limited success.  The path we took for `aztables` and key vault works much better.

PR #830, opened 2022-05-31
Add DO NOT EDIT comment to generated code

jhendrixMSFT:
The language server will issue a warning when attempting to modify a
source file that includes this comment.


jhendrixMSFT:
See https://github.com/golang/go/issues/33397 for `gopls` support.

serbrech:
linters tools also automatically ignore files based on these heuristics, so that's nice

Issue #834, opened 2022-06-06
Do not overwrite `go.mod` file if there exists one


tadelesh:
This will remove any extra info in go.mod like `retract`, `exclude` or `replace`. If there exists a `go.mod` file, we can do some replacement for the `azcore` version and `go` version.

jhendrixMSFT:
Fix will be in `preview.43`

PR #835, opened 2022-06-06
Don't overwrite an existing go.mod file

jhendrixMSFT:
If there's a preexisting go.mod file, update its version of azcore if
its semver is less than the generator or specified version.

Fixes https://github.com/Azure/autorest.go/issues/834


jhendrixMSFT:
That's a good point.  Reading the issue a bit closer, there was a suggestion about making modifications to the existing `go.mod` for things like this instead of overwriting it.  Let me take a look at that.

chlowell (approved):

Issue #833, opened 2022-06-01
Single value enum property with modalAsString=false


tadelesh:
For single value enum property with modalAsString=false, it is regarded as constants property in m4. It's OK to expose the property to the end user as it could be a return type, but could we set the value directly when marshalling to avoid any wrong usage?

jhendrixMSFT:
Fixed in upcoming `preview.43`

PR #836, opened 2022-06-07
Embed constant values in JSON marshallers

jhendrixMSFT:
Create a JSON marshaller for structs that contain fields with constant
values.  The marshaller embeds the constant value instead of the field.
Add a doc comment for constant fields where the value is ignored.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/833

chlowell (approved):

Issue #847, opened 2022-07-07
Add remove non-reference type feature


tadelesh:
Related issue: https://github.com/Azure/autorest.go/issues/757
We found the following case will cause breaking change in mgmt. plane SDK:
1. Discriminator model change from user-defined model to swagger common model
2. Orphan model removal
3. Add new orphan model
We'd like to add feature to remove all the non-reference type to the codegen. For now, only mgmt. new packages will enable such feature, existed packages will enable such feature when swagger breaking happens. Data plane will not use this as non-reference types have other usage in some RPs.

PR #849, opened 2022-07-12
label unreferenced types and skip them when generating code

tadelesh:
Fix: #847

1. Label unreferenced types after `namer` and `transform`
2. Iterator all operations and label all request param and response schema/header schema
3. For discriminator, label all the schema for all parents and children


src/common/helpers.ts
```diff
@@ -199,3 +209,29 @@ export function formatConstantValue(schema: ConstantSchema): string {
...
+          }
+          // don't add the duplicate
+          continue;
+        }
+        allProps.push(parentProp);
+      }
+    }
+  }
+  return allProps;
+}
```
jhendrixMSFT, line 237: Missing new-line.

src/transform/transform.ts
```diff
@@ -938,3 +939,95 @@ function parseComments(comment: string): string {
...
+      if (isObjectSchema(parent)) {
+        dfsSchema(parent, visited, choices, sealedChoices, objects);
+      }
+    }
+  } else if (isArraySchema(schema)) {
+    dfsSchema(schema.elementType, visited, choices, sealedChoices, objects);
+  } else if (isDictionarySchema(schema)) {
+    dfsSchema(schema.elementType, visited, choices, sealedChoices, objects);
+  }
+}
```
jhendrixMSFT, line 1033: Missing new-line.

.scripts/regeneration.js
```diff
@@ -10,47 +10,47 @@ const sem = require('./semaphore')(8);
...
-    'filegroup': ['body-file.json'],
-    'formdatagroup': ['body-formdata.json'],
-    'headergroup': ['header.json'],
-    'headgroup': ['head.json', '--head-as-boolean'],
-    'httpinfrastructuregroup': ['httpInfrastructure.json', '--head-as-boolean'],
-    'integergroup': ['body-integer.json'],
-    'lrogroup': ['lro.json'],
-    'mediatypesgroup': ['media_types.json'],
-    'migroup': ['multiple-inheritance.json'],
+    'additionalpropsgroup': ['additionalProperties.json', '--remove-non-reference-schema'],
```
jhendrixMSFT, line 13: Maybe we can shorten this a bit to `--remove-unreferenced-types`.

src/transform/transform.ts
```diff
@@ -938,3 +939,95 @@ function parseComments(comment: string): string {
...
+  if (modelSchemas){
+    for (const schema of modelSchemas){
+      if (!existedSet.has(schema)){
+        schema.language.go!.omitType = true;
+      }
+    }
+  }
+}
+
+function iterateOperations(model: CodeModel, choices: Set<ChoiceSchema>, sealedChoices: Set<SealedChoiceSchema>, objects: Set<ObjectSchema> ) {
```
jhendrixMSFT, line 973: Could you please add a comment explaining what this does.

src/transform/transform.ts
```diff
@@ -938,3 +939,95 @@ function parseComments(comment: string): string {
...
+        const schema = values(model.schemas.objects).where((o) => o.language.default.name === schemaParts[schemaParts.length - 1]).first();
+        if (schema) {
+          dfsSchema(schema, visited, choices, sealedChoices, objects);
+        }
+      }
+    }
+  }
+}
+
+function dfsSchema(schema: Schema, visited: Set<Schema>, choices: Set<ChoiceSchema>, sealedChoices: Set<SealedChoiceSchema>, objects: Set<ObjectSchema>) {
```
jhendrixMSFT, line 1003: Could you please add a comment explaining what this does.

test/autorest/errorsgroup/zz_generated_models.go
```diff
@@ -13,63 +13,6 @@ type Animal struct {
...
-		WhatNotFound: l.WhatNotFound,
-		SomeBaseProp: l.SomeBaseProp,
-	}
-}
-
-// NotFoundErrorBaseClassification provides polymorphic access to related types.
-// Call the interface's GetNotFoundErrorBase() method to access the common type.
-// Use a type switch to determine the concrete type.  The possible types are:
-// - *AnimalNotFound, *LinkNotFound, *NotFoundErrorBase
-type NotFoundErrorBaseClassification interface {
```
jhendrixMSFT, line 58: We cannot remove a discriminator or any of its child types.  When encountering a discriminated type, you'll need to mark its entire hierarchy as required.

https://github.com/Azure/autorest.testserver/blob/main/swagger/xms-error-responses.json#L148

jhendrixMSFT (changes requested):

test/autorest/errorsgroup/zz_generated_models.go
tadelesh: All the error types have been removed because we do not use the error type in our SDK directly. If you think we should reserve them, I can change the logic. For all the discriminators, all the type chain will be reserved if one type has been referenced.

jhendrixMSFT (approved):

src/transform/transform.ts
```diff
@@ -938,3 +939,96 @@ function parseComments(comment: string): string {
...
+        }
+      }
+      // Such odata definition is not used directly in operation. But it seems the model is a detailed definition for some string param. Reserve for now.
+      if (op.extensions?.['x-ms-odata']) {
+        const schemaParts = op.extensions['x-ms-odata'].split('/');
+        const schema = values(model.schemas.objects).where((o) => o.language.default.name === schemaParts[schemaParts.length - 1]).first();
+        if (schema) {
+          dfsSchema(schema, referencedTypes);
+        }
+      }
```
ArcturusZhang, line 1001: if we have to parse the odata extension in this way by splitting the path, does this mean the modeler v4 does not do anything to this extension?

src/transform/transform.ts
```diff
@@ -938,3 +939,96 @@ function parseComments(comment: string): string {
...
+      for (const child of values(schema.children?.all)) {
+        dfsSchema(child, referencedTypes);
+      }
+    }
+    // If schema is has allOf, then reserve all the parents
+    for (const parent of values(schema.parents?.immediate)) {
+      if (isObjectSchema(parent)) {
+        dfsSchema(parent, referencedTypes);
+      }
+    }
```
ArcturusZhang, line 1028: Should we preserve the parent schema or omit the parent schema? I believe the parent schemas are one of the biggest source of unused models since we are flattening the properties from its parent into itself

test/compute/2019-12-01/armcompute/zz_generated_models.go
```diff
@@ -268,12 +268,6 @@ type BootDiagnosticsInstanceView struct {
 	Status *InstanceViewStatus `json:"status,omitempty" azure:"ro"`
 }
 
-// CloudError - An error response from the Compute service.
-type CloudError struct {
```
ArcturusZhang, line 272: Take this file as an example: we are removing:
- errors
- models that are not referenced in the swagger (I know that they have one in gallery `GalleryArtifactSource` and all its siblings)

But I think we should have more to omit in this RP, such like, `Resource` or `TrackedResource` and other kind of `allOf` models. They are never directly used by our generated code, and if they switch to the base type in common-types, and when they do that, there usually some slightly name difference - we will have troubles.
Like when they are defining swaggers by themselves, for simplification consideration, they would just name an entity with id, name, type, systemData, location and tag as a `Resource`, but in fact, in common-types, this thing is called a `TrackedResource` or `ProxyResource`. When they switch to this, we will see the Resource is removed and we have a TrackedResource, both of them are usually never referenced in our SDK

ArcturusZhang (commented):
I have some concerns that we did not cover all cases of "unreferenced types" in this PR.
One of them in top of my head is the base types. And we also had quite a few troubles from them already.

src/transform/transform.ts
tadelesh: It seems yes.

test/compute/2019-12-01/armcompute/zz_generated_models.go
tadelesh: I add the constrains for the parents. If the type is not a discriminator, it will not be iterated.

ArcturusZhang (approved):

tadelesh:
Test result for all mgmt. RPs: [result.log](https://github.com/Azure/autorest.go/files/9139421/result.log)

PR #850, opened 2022-07-12
fix doc broken url

tadelesh:


tadelesh:
@jhendrixMSFT, can we set the branch rule according to the source code folder? I just changed something with test generator, but `Azure.autorest.go` always is required which blocks the merge.

lirenhe (approved):

tadelesh:
> @jhendrixMSFT @tadelesh I believe for the pipeline in question the branch includes/excludes can be updated [here](https://github.com/Azure/autorest.go/blob/main/eng/pipelines/ci-gotest.yml#L1-L17).

@benbp Thanks for the quick reply. We've already separated the two pipelines. But the question here is about the branch rule in GitHub. We set `Azure.autorest.go` as required check before. Now we want such check only required for some paths. Is that possible?

benbp:
@tadelesh this is not possible with github, you can only set branch protection rules global to the repository. We work around this constraint in the other repositories with check-enforcer (see [docs](https://github.com/Azure/azure-sdk-actions/blob/main/docs/check-enforcer.md)). The check-enforcer status is the primary branch protection rule, and it in turn enforces a minimum number of azure pipeline runs.

You can set this up for this repository by copying over [this file](https://github.com/Azure/azure-sdk-tools/blob/main/.github/workflows/event.yml). Once that's merged, it will show up as a status check and I can go in and add it as the main branch protection rule.

tadelesh:
@benbp Check enforcer enabled with this [PR](https://github.com/Azure/autorest.go/pull/852). Could you help to update the check policy? Thanks.

Issue #858, opened 2022-08-01
Prevent breaking change for adding a new request body type


tadelesh:
### Problem description
For current codegen, an operation has only one JSON request at first, we will generate this operation with no naming suffix. Then new request body type has been added to this operation. According to the current logic, if an operation with more than one request body types, we create a new method with the media type name as a suffix for the non-binary types. This causes the breaking changes for the original operation name from XXX to XXXWithJSON.
### Example
- [original swagger](https://github.com/Azure/autorest.testserver/blob/e66f072497164e23886e40ac687774333b8d3671/swagger/dpg-initial.json#L75)
- [updated swagger](https://github.com/Azure/autorest.testserver/blob/e66f072497164e23886e40ac687774333b8d3671/swagger/dpg-update1.json#L93)
- [original generated code](https://github.com/tadelesh/go-dpg-poc/blob/8773e9124d045b666f33f7bb48c4c81490707a11/evolution/service/initial/params_client.go#L166)
- [updated generated code](https://github.com/tadelesh/go-dpg-poc/blob/8773e9124d045b666f33f7bb48c4c81490707a11/evolution/service/update/params_client.go#L291)
### Solution
For operation with more than one response body types, we refer the sequence of the consumes in swagger and add WithXXX suffix for the second consumes and so on. It strongly relies on the swagger change, but it could minimize the influence for current generated control plane and data plane code.

tadelesh:
@JeffreyRichter @jhendrixMSFT @chlowell What do you think?

JeffreyRichter:
I agree that we have a problem here. The proposal requires that NEW response Content-Types always be added after OLD content types in th swagger. We'd need a linter rule to verify this at least.
But also, it is a bit more nuanced. If a service could return the SAME response in JSON or XML, there is no need to expose this distinction in the SDK at all because the wire format is suppoed to be an implementation detail to SDK customers. The SDK should just pick the most efficient content type to use automatically. In other words, if the models used for the 2+ content types is the same, we only need to expose one method i nthe SDK.

tadelesh:
[...]

PR #859, opened 2022-08-02
feat: change the operation name with more than one response body type

tadelesh:
Fix: https://github.com/Azure/autorest.go/issues/858


chlowell:
Is it safe to rely on the order of body types in the swagger? For an operation consuming only JSON, we would generate a `Foo()` method the JSON body type. With this code, if a future version of the swagger adds `"consumes": ["image/jpeg", "application/json"]`, we would generate `Foo()` for image/jpeg and `FooWithJSON()`.

Similarly, isn't this a non-breaking change for the swagger?
```diff
---"consumes": ["application/json", "image/jpeg"]
+++"consumes": ["image/jpeg", "application/json"]
```

tadelesh:
> Is it safe to rely on the order of body types in the swagger? For an operation consuming only JSON, we would generate a `Foo()` method the JSON body type. With this code, if a future version of the swagger adds `"consumes": ["image/jpeg", "application/json"]`, we would generate `Foo()` for image/jpeg and `FooWithJSON()`.
> 
> Similarly, isn't this a non-breaking change for the swagger?
> 
> ```diff
> ---"consumes": ["application/json", "image/jpeg"]
> +++"consumes": ["image/jpeg", "application/json"]
> ```

Because we already have shipped some packages with two consumes, I could only rely on the order to prevent breaking change for released packages. Regarding we must select one for non-suffix name operation, we could not relay on any sort from codegen side, so just leave it to the swagger author. Any sequence change will be detected by SDK breaking change tools, but yes, it is not a swagger breaking change.

jhendrixMSFT (changes requested):
Relying on the ordering is fragile so we shouldn't be going down that path.  We can discuss possible solutions in the linked issue.

test/autorest/optionalgroup/zz_generated_explicit_client.go
```diff
@@ -341,28 +341,28 @@ func (client *ExplicitClient) postOptionalStringHeaderCreateRequest(ctx context.
 	return req, nil
 }
 
-// PostOptionalStringParameter - Test explicitly optional string. Please put null.
+// PostOptionalStringParameterWithText - Test explicitly optional string. Please put null.
```
jhendrixMSFT, line 344: This feature should be off by default.  We shouldn't be changing any existing names.

jhendrixMSFT (changes requested):

test/autorest/optionalgroup/zz_generated_explicit_client.go
tadelesh: Change default to `false`.

src/transform/transform.ts
```diff
@@ -293,34 +293,60 @@ function recursiveAddMarshallingFormat(schema: Schema, marshallingFormat: 'json'
...
+            newOp.language.go!.name = name;
+            newOp.language.go!.protocolNaming = new protocolMethods(newOp.language.go!.name);
+            group.addOperation(newOp);
+            if (req.language.go!.description) {
+              req.language.go!.description = parseComments(req.language.go!.description);
+            }
+          }
+          group.operations.splice(group.operations.indexOf(op), 1);
+        }
+      } else {
```
jhendrixMSFT, line 340: A lot of this code looks duplicated from the above branch.  Can this be consolidated?  The algorithm itself looks good.

tadelesh: Resolved.

jhendrixMSFT (approved):

PR #862, opened 2022-08-16
Add support for x-ms-external extension

jhendrixMSFT:
This allows types to be defined external to codegen.


test/swagger/alias.json
```diff
@@ -392,6 +392,7 @@
       }
     },
     "GeoJsonObjectNamedCollection": {
+      "x-ms-external": true,
```
chlowell, line 395: I suppose you would do this instead of simply deleting the definition when the definition is referenced by something you do want to generate?

chlowell (approved):

test/swagger/alias.json
jhendrixMSFT: Correct.  This way, you can have a hand-written model used by generated code.

Issue #864, opened 2022-08-17
Always generate 64 precision number to prevent breaking changes


tadelesh:
For DPG, there exists a rule of always generating 64 precision number for LLC code to prevent possible breaking changes when service change some param/property type from 32 precision to 64 precision.
Current codegen generates code according to the user defined precision. We could opt-in always generate int64/float64.

tadelesh:
Add @JeffreyRichter @jhendrixMSFT for discussion.

jhendrixMSFT:
Do we have any data on how often this kind of breaking change happens?

JeffreyRichter:
It is extremely rare. The best example fo when it did happen is when the max size of a blob changed.
64-bit integers can be painful to work with: more size, slower, as an array index, etc. I personally feel that this is rare enough that it is better to keep the 32-bit ints.

tadelesh:
This choice will also influence CADL support (CADL has `int32`, `int64`, `float32`, `float64`). I've confirmed with other compiled languages: .NET decided to follow the defined precision while Java will always use `long` for `int` and double for `float`. Since it should be a rare situation (I can't even find an example), and I agree that 32-bit number could be benefit for memory and speed, I think we will follow user-defined precision for Go codegen. cc @mikekistler

PR #865, opened 2022-08-17
opt-in ignore precision for int and float type

tadelesh:
Resolve: #864


tadelesh:
WIP: need add logic to gotest

PR #868, opened 2022-08-23
Add support for client-side default values for models

jhendrixMSFT:
Sets the specified default value during marshalling when no value is
specified.


jhendrixMSFT:
https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-client-default for reference

chlowell (approved):

RickWinter (approved):

PR #871, opened 2022-08-24
Continue the pipeline if publishing autorest coverage fails

jhendrixMSFT:
Fixing incorrect change from previous commit.


tadelesh:
Replaced by #873.

PR #845, opened 2022-07-04
Migrate gotest extension from azure-tools to this repo

tadelesh:
This PR try to migrate all the code of `@autorest/gotest` extension from `azure-tools` repo to this repo. Original code position: https://github.com/Azure/azure-sdk-tools/tree/main/tools/sdk-testgen/packages/autorest.gotest. Reuse `rush` to manage mono package. Two extensions will get separate eng script and build/release process for `@autorest/go` remains unchanged.

Resolve: https://github.com/Azure/azure-sdk-tools/issues/3379

PR #873, opened 2022-08-24
`publish autorest test server coverage report` only triggered in main branch ci

tadelesh:
This PR fixes the failure of autorest/go publish. The root reason is that when I try to extract and reuse build and test step in this [PR](https://github.com/Azure/autorest.go/pull/845), I forgot to exclude the coverage report step from the publish pipeline. Revert #869, #870 and close #871.

### DIFF
--- a/eng/steps/build-test.yaml
+++ b/eng/steps/build-test.yaml
@@ -101,4 +101,3 @@ steps:
           cd src/node_modules/@microsoft.azure/autorest.testserver;
           npm run coverage -- publish --repo=$(Build.Repository.Name) --ref=$(Build.SourceBranch) --githubToken=$(azuresdk-github-pat) --azStorageAccount=$(storage-coverage-user) --azStorageAccessKey=$(storage-coverage-pass) --version=$currentVersion;
         displayName: "Publish AutoRest Test Server Coverage Report"
-        condition: succeededOrFailed()
--- a/eng/steps/build-test.yaml
+++ b/eng/steps/build-test.yaml
@@ -1,5 +1,6 @@
 parameters:
   GoTestPath: "$(system.defaultWorkingDirectory)/test"
+  PublishAutorestCoverage: false
 
 steps:
   - script: |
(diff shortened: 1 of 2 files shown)


eng/steps/build-test.yaml
```diff
@@ -94,11 +95,10 @@ steps:
...
       failTaskOnFailedTests: true
 
-  - ${{if ne(variables['Build.Reason'], 'PullRequest')}}:
+  - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(parameters.PublishAutorestCoverage, true))}}:
       - pwsh: |
           $currentVersion = node -p -e "require('./src/package.json').version";
           $currentVersion="$currentVersion-$(Build.BuildNumber)";
           cd src/node_modules/@microsoft.azure/autorest.testserver;
-          npm run coverage -- publish --repo=$(Build.Repository.Name) --ref=$(Build.SourceBranch) --githubToken=$(azuresdk-github-pat) --azStorageAccount=$(storage-coverage-user) --azStorageAccessKey=$(storage-coverage-pass) --version=$currentVersion;
+          npm run coverage -- publish --repo=$(Build.Repository.Name) --ref=$(Build.SourceBranch) --githubToken=$(github-token) --azStorageAccount=$(storage-coverage-user) --azStorageAccessKey=$(storage-coverage-pass) --version=$currentVersion;
```
jhendrixMSFT, line 103: Ben fixed this yesterday, did you mean to revert it?

tadelesh: Yes. `github-token`, `storage-coverage-user` and `storage-coverage-pass`is configured in the public ci variable. This step should not be executed in private ci (publish npm package).

chlowell (approved):

RickWinter (approved):

benbp (approved):

eng/steps/build-test.yaml
benbp: Shouldn't we skip publish altogether on the public pipeline? We don't want any credentials exposed for that since any pull request can access it automatically.

tadelesh: > Shouldn't we skip publish altogether on the public pipeline? We don't want any credentials exposed for that since any pull request can access it automatically.

The value set to the variable `github-token` in public repo is `skip`. This pipeline was set up long time ago and I just did some refactor. It seems the common pipeline for all languages' codegen.

Issue #867, opened 2022-08-22
Add client ctor with credential support if swagger provide security definitions


tadelesh:
For most auto-generated LLC packages, they have clear security definitions to indicate how to get credential when using the service (some examples: [#1](https://github.com/Azure/azure-rest-api-specs/blob/81cd88a080c4bf4bb251afbe62892a6e220cb2b4/specification/cognitiveservices/data-plane/Language/stable/2022-05-01/analyzeconversations-authoring.json#L8), [#2](https://github.com/Azure/azure-rest-api-specs/blob/81cd88a080c4bf4bb251afbe62892a6e220cb2b4/specification/cognitiveservices/data-plane/Speech/SpeechToText/stable/v3.0/speechtotext.json#L5628), [#3](https://github.com/Azure/azure-rest-api-specs/blob/81cd88a080c4bf4bb251afbe62892a6e220cb2b4/specification/maps/data-plane/Microsoft.Maps/FeatureState/preview/2.0/featurestate.json#L23)).
In order to provide a more friendly client and reduce manual work for DPG, I'd like to auto generate the ctor with credential according to the swagger.
Regarding there will be multiple security definitions for one client, the client ctor will use security definitions name as suffix, such as `NewFeatureStateClientWithAADTokenCredential`, `NewFeatureStateClientWithSharedKeyCredential`, `NewFeatureStateClientWithSasTokenCredential`.
Also, a new `KeyCredential` and `KeyCredentialPolicy` will be added to `azcore`. 
This is somehow a DPG feature and also will resolve this [issue](https://github.com/Azure/autorest.go/issues/822).
Any concern? @jhendrixMSFT  @JeffreyRichter

JeffreyRichter:
I do not think you can ultimately be successful with this, which is why we have not done it already.
First, we are assuming that all Azure services will support TokenCredential now and so we do want to auto-generate ctors for this and because we (Azure) want customers to use this cred primarily, we name this ctor NewFeatureStateClient; we do NOT add the WithAADTokenCredential suffix but this ctor does take a TokenCredentail.

Second, most new Azure services will use TokenCred as the only auth mechanism and so we'll get pretty far with code-gen'ing this ctor.

Third, you can't auto-gen SharedKey or SAS creds because what services do with these varies and it's not in the swagger. For example, Storage requires both of these creds to create strings to sign and how to construct the string is not defined in the swagger so it must be manually written. And because the algorithm is service-specific, it must live in the service's package and not in the core package.

[...]

PR #878, opened 2022-09-15
Generate ctor for data plane if OAuth2 definition exist

tadelesh:
Resolve https://github.com/Azure/autorest.go/issues/867

### DIFF
--- a/src/generator/operations.ts
+++ b/src/generator/operations.ts
@@ -7,7 +7,7 @@ import { Session } from '@autorest/extension-base';
 import { capitalize, comment, KnownMediaType, uncapitalize } from '@azure-tools/codegen';
 import { ApiVersions, ArraySchema, ByteArraySchema, ChoiceSchema, CodeModel, ConstantSchema, DateTimeSchema, DictionarySchema, GroupProperty, ImplementationLocation, NumberSchema, Operation, OperationGroup, Parameter, Property, Protocols, Response, Schema, SchemaResponse, SchemaType } from '@autorest/codemodel';
 import { values } from '@azure-tools/linq';
-import { aggregateParameters, formatConstantValue, getSchemaResponse, isArraySchema, isBinaryResponseOperation, isMultiRespOperation, isPageableOperation, isSchemaResponse, isTypePassedByValue, isLROOperation, commentLength } from '../common/helpers';
+import { aggregateParameters, formatConstantValue, getSchemaResponse, isArraySchema, isBinaryResponseOperation, isMultiRespOperation, isPageableOperation, isSchemaResponse, isTypePassedByValue, isLROOperation, commentLength, hasOAuth2SecurityDefinition, getOAuth2SecuritySchema } from '../common/helpers';
 import { OperationNaming } from '../transform/namer';
 import { contentPreamble, elementByValueForParam, formatParameterTypeName, formatStatusCodes, formatValue, getClientDefaultValue, getResponseEnvelope, getResponseEnvelopeName, getResultFieldName, getStatusCodes, hasDescription, hasResultProperty, hasSchemaResponse, skipURLEncoding, sortAscending, getCreateRequestParameters, getCreateRequestParametersSig, getMethodParameters, getParamName, formatParamValue, dateFormat, datetimeRFC1123Format, datetimeRFC3339Format, sortParametersByRequired, substituteDiscriminator } from './helpers';
 import { ImportManager } from './imports';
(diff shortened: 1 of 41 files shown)


test/keyvault/7.2/azkeyvault/zz_generated_client.go
```diff
@@ -27,8 +28,23 @@ type Client struct {
 }
 
 // NewClient creates a new instance of Client with the specified values.
-// pl - the pipeline used for sending requests and handling responses.
-func NewClient(pl runtime.Pipeline) *Client {
+// credential - used to authorize requests. Usually a credential from azidentity.
+// options - pass nil to accept the default values.
+func NewClient(credential azcore.TokenCredential, options *azcore.ClientOptions) *Client {
```
jhendrixMSFT, line 33: We need to allow for a hand-written options type for per-client extensibility ([example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/keyvault/azcertificates/custom_client.go#L25)).  Probably should just take a `*ClientOptions` type which will need to be user-provided.

tadelesh: Could you help to explain the reason why we need this wrapped client options?

jhendrixMSFT: All data-plane clients must have their own client options type.  This allows future expansion without any breaking changes and is part of the design guidelines.

tadelesh: I've added client option types for each client. Please help to review again.

jhendrixMSFT: To clarify, the `ClientOptions` type should _not_ be generated, else you can't hand-edit them.  It will require that the SDK author provide them.

tadelesh: I see. I've updated the code and resolved the test.

test/synapse/2019-06-01/azspark/zz_generated_batch_client.go
```diff
@@ -19,17 +20,25 @@ import (
...
-// newBatchClient creates a new instance of batchClient with the specified values.
+// NewBatchClient creates a new instance of BatchClient with the specified values.
 // endpoint - The workspace development endpoint, for example https://myworkspace.dev.azuresynapse.net.
 // livyAPIVersion - Valid api-version for the request.
 // sparkPoolName - Name of the spark pool.
-// pl - the pipeline used for sending requests and handling responses.
-func newBatchClient(endpoint string, livyAPIVersion *string, sparkPoolName string, pl runtime.Pipeline) *batchClient {
+// credential - used to authorize requests. Usually a credential from azidentity.
+// options - pass nil to accept the default values.
+func NewBatchClient(endpoint string, livyAPIVersion *string, sparkPoolName string, credential azcore.TokenCredential, options *BatchClientOptions) *BatchClient {
```
jhendrixMSFT, line 36: This poses an interesting problem.  `livyAPIVersion` is optional so it belongs in the options type.  But we're not generating the options type.  We don't have to fix it in this PR, but we can't ship any packages like this.

src/generator/operations.ts
```diff
@@ -198,7 +205,7 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
     } else {
       clientText += `func ${clientCtor}(${methodParams.join(', ')}) *${clientName} {\n`;
     }
-    if (azureARM) {
+    if (azureARM || hasOauth2Security) {
       // data-plane doesn't take client options
```
jhendrixMSFT, line 209: Nit, comment is no longer applicable.

src/generator/operations.ts
```diff
@@ -327,7 +331,7 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
       clientText += '\treturn client\n';
     }
     clientText += '}\n\n';
-
+
```
jhendrixMSFT, line 334: Nit, extra spaces.

jhendrixMSFT (approved):

test/keyvault/7.2/azkeyvault/client_options.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// ClientOptions contains the optional settings for Client.
+type ClientOptions struct {
+	azcore.ClientOptions
+}
+
+// HSMSecurityDomainClientOptions contains the optional settings for Client.
```
chlowell, line 18: I guess this is from operations.ts:121. Looks like replacing `clientCtor` with `clientName` would fix it, but it's odd that `clientCtor === 'Client'`

tadelesh: These client options are manual written. I've refined all the comment for client options.

PR #885, opened 2022-09-29
Suggestion: Improve generated code comments for `godoc`

weikanglim:
Current behavior: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/keyvault/armkeyvault#KeysClient.Get
image

Comparison of before vs. after:
image

`go` has support for some document formatting. See [here](https://tip.golang.org/doc/comment). I'm suggesting two changes to take use of formatting:

- API version generated should be its own paragraph. This allows the first sentence to accurately describe method behavior with API version gen as an addendum.
- Parameter lists are formatted as a list


jhendrixMSFT (approved):

jhendrixMSFT:
When I look at this locally it's slightly different.  Looks like a code block, not a list.  How did you create your snippet?

image

weikanglim:
> When I look at this locally it's slightly different. Looks like a code block, not a list. How did you create your snippet?
> 
> image

That's weird. This exact code snippet looks correct on mine:
```golang
// GetEmpty - Get complex types with array property which is empty
// If the operation fails it returns an *azcore.ResponseError type.
//
// Generated from API version 2016-02-29
//   - options - ArrayClientGetEmptyOptions contains the optional parameters for the ArrayClient.GetEmpty method.
func GetEmpty() {
}
```

image

weikanglim:
@jhendrixMSFT `go` would detect that as a codeblock if the bullet list is somehow unrecognized:

From the [docs](https://tip.golang.org/doc/comment) on Code Blocks:

> A code block is a span of indented or blank lines not starting with a bullet list marker or numbered list marker. It is rendered as preformatted text (a `<pre>` block in HTML).

I could reproduce this behavior but only if I made the bullet list invalid (like not introducing a space after the bullet list marker):

```golang
// GetEmpty - Get complex types with array property which is empty
// If the operation fails it returns an *azcore.ResponseError type.
//
// Generated from API version 2016-02-29
//	-options - ArrayClientGetEmptyOptions contains the optional parameters for the ArrayClient.GetEmpty method.
```

jhendrixMSFT:
@weikanglim did you generate your snippet running a local instance of `pkgsite` or is this on the official site?

jhendrixMSFT:
My bad, I was running an older version of pkgsite. Things look as expected.

PR #886, opened 2022-10-07
Export method options types when client isn't

jhendrixMSFT:


src/transform/transform.ts
```diff
@@ -339,8 +339,10 @@ async function processOperationRequests(session: Session<CodeModel>) {
       // don't do this for paging next link operation as this isn't part of the public API
       if (!op.language.go!.paging || !op.language.go!.paging.isNextOp) {
         // create a type named <OperationGroup><Operation>Options
-        const optionalParamsGroupName = `${group.language.go!.clientName}${opName}Options`;
-        const desc = `${optionalParamsGroupName} contains the optional parameters for the ${group.language.go!.clientName}.${opName} method.`;
+        // the client might not be exported, but the options should be.
+        const optsGroupName = capitalize(group.language.go!.clientName);
+        const optionalParamsGroupName = `${optsGroupName}${opName}Options`;
+        const desc = `${optionalParamsGroupName} contains the optional parameters for the ${optsGroupName}.${opName} method.`;
```
chlowell, line 345:
```suggestion
        const desc = `${optionalParamsGroupName} contains the optional parameters for the ${group.language.go!.clientName}.${opName} method.`;
```
so we don't also capitalize an unexported client's name in its docs

chlowell (approved):

src/transform/transform.ts
jhendrixMSFT: Oh yes good point, fixed.

Issue #887, opened 2022-10-11
Generate more useful response usage for all the examples


tadelesh:
Some feedbacks from customer interviews and surveys said that they want to get more information about how to use the response model in the examples. Current examples only use blank identifier to resolve response. We could generate the response usage according to example response definition.

PR #888, opened 2022-10-12
feat: add response struct demo for all the examples

tadelesh:
Resolve: #887


packages/autorest.gotest/test/integrationtest/output/compute/armcompute/virtualmachines_client_example_test.go
```diff
@@ -40,6 +40,241 @@ func ExampleVirtualMachinesClient_NewListByLocationPager() {
...
+		// 				},
+		// 				"vmId": "{vmId}"
+		// 			},
+		// 			"tags": {
+		// 				"RG": "rg",
+		// 				"testTag": "1"
+		// 			}
+		// 		}
+		// 	]
+		// }
```
tadelesh, line 277: @xboxeer @ArcturusZhang It seems too long if we comment all the response struct and JSON. Any suggestion?

xboxeer: is this the longest or just "middle size"?

tadelesh: I'm not sure. Perhaps compute has the most complicated examples. You could review all the `**/*_example_test.go` files [here](https://github.com/Azure/autorest.go/tree/a51dd41438a735d97b7187b48c5d4daaf88f4692/packages/autorest.gotest/test/integrationtest/output) to see whether they are what you want.

xboxeer: Others like signalR seems to be a reasonable length, let's roll this out first and see what is the feedback there

tadelesh: OK. Any concern? @ArcturusZhang @lirenhe

lirenhe: could you help to explain why we need the raw response? Is it from customer's ask?

xboxeer: > could you help to explain why we need the raw response? Is it from customer's ask?

exactly, customer would like copy json output in their unit test.

lirenhe: For this case, can we suggest them to use swagger example directly?

tadelesh: > For this case, can we suggest them to use swagger example directly?

Each example has a comment point to the swagger example. But it seems not intuitional for the customers.

xboxeer: > For this case, can we suggest them to use swagger example directly?

For customers who are not familiar with REST API, they would be uncomfortable to look for output in REST API Reference site
Another thing is that versioning, REST API reference site currently only support latest version API while SDK is fixed to one version API, we don't want bring extra information to customer about the versioning thing.

packages/autorest.gotest/src/generator/exampleGenerator.ts
```diff
@@ -28,27 +29,48 @@ export class ExampleCodeGenerator extends BaseCodeGenerator {
 
       let fileName = exampleModel.operationGroup.language.go!.clientName.toLowerCase();
       if (fileName !== 'client') {
-        fileName = fileName.substring(0, fileName.length-6) + '_client';
+        fileName = fileName.substring(0, fileName.length - 6) + '_client';
```
ArcturusZhang, line 32: Should we verify that this does end with client for sure?

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
```diff
@@ -32,11 +33,41 @@ func ExampleAppsClient_Get() {
 	if err != nil {
 		log.Fatalf("failed to finish the request: %v", err)
 	}
-	// TODO: use response item
+	// TODO: use response
```
ArcturusZhang, line 36: should we remove the `TODO`?

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
```diff
@@ -32,11 +33,41 @@ func ExampleAppsClient_Get() {
...
 	_ = res
+	// If HTTP returns as the '200' response defined in example definition, response struct should like:
```
ArcturusZhang, line 38: should we also emphasize it here that some values below are fake and may or may not be in its proper format?
For instance, the `principalId`

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
tadelesh: If we remove the TODO part, will it be weird that we use blank identifier to hold the res? @xboxeer What do you think?

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
tadelesh: @xboxeer Could you help to refine the sentence here?

xboxeer: If the HTTP response code is 200 as defined in example definition, your response structure would look like this. Please pay attention that all the values in the output are fake values for just demo purposes,

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
xboxeer: > should we remove the `TODO`?

@ArcturusZhang do you have any reason why we should remove the TODO, for me both having it or remove it looks good

xboxeer (approved):
LGTM

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/apps_client_example_test.go
```diff
@@ -32,11 +33,40 @@ func ExampleAppsClient_Get() {
 	if err != nil {
 		log.Fatalf("failed to finish the request: %v", err)
 	}
-	// TODO: use response item
```
ArcturusZhang, line 35: For instance we could add some comments like
```
// use the response here. For demo we just throw it into black hole
```

ArcturusZhang (approved):

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/configservers_client_example_test.go
```diff
@@ -66,11 +83,27 @@ func ExampleConfigServersClient_BeginUpdatePut() {
...
+	// 					URI: to.Ptr("https://github.com/fake-user/fake-repository.git"),
+	// 				},
+	// 			},
+	// 			ProvisioningState: to.Ptr(armappplatform.ConfigServerStateSucceeded),
+	// 		},
+	// 	}
 }
 
-// Generated from example definition: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/appplatform/resource-manager/Microsoft.AppPlatform/preview/2020-11-01-preview/examples/ConfigServers_UpdatePatch.json
+// Generated from example definition: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appplatform/resource-manager/Microsoft.AppPlatform/preview/2020-11-01-preview/examples/ConfigServers_UpdatePatch.json
```
lirenhe, line 106: I think last time we discussed about referencing the sample from a specific commit.
So here, why are we still using the one in main branch?

tadelesh: Some of the tests are from the local swagger file that we could not find the commit. For those use public swagger repo, it will be referenced to a stable commit like this [test](https://github.com/Azure/autorest.go/blob/b585f2190764256c49c371c0ec90cc6ed2f796d0/packages/autorest.gotest/test/integrationtest/output/agrifood/armagrifood/extensions_client_example_test.go#L20).

lirenhe (approved):

PR #893, opened 2022-10-21
make step variable generation inline and support generation of all variable types

### LINKED ISSUE #890: testgen could not handle non-string variable and step variable in api scenario correctly
Example:
```
scope: ResourceGroup

variables:
  globalLocation: Global
  publicPort:
    type: int
    value: 6910

scenarios:
  - scenario: SignalR
    description: Microsoft.SignalRService/Basic_CRUD
    steps:
      - step: SignalR_CreateOrUpdate
        exampleFile: ../examples/SignalR_CreateOrUpdate.json
        requestUpdate:
          - replace: /parameters/properties/publicPort
            value: $(publicPort)
          - replace: /parameters/location
            value: $(location)-test1
        variables:
          location: $(globalLocation)-test2
        outputVariables:
          signalRId:
            fromResponse: /id
```

tadelesh:
Resolve #890.
For API scenario, step variable is meaningless. So we just generate them inline with the real value. Also, all variable types support ref, so change related logic to support variable replacement.

### DIFF
--- a/packages/autorest.gotest/src/generator/scenarioTestGenerator.ts
+++ b/packages/autorest.gotest/src/generator/scenarioTestGenerator.ts
@@ -15,6 +15,7 @@ import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
 import { MockTestDataRender } from './mockTestGenerator';
 import { OavStepType } from '@autorest/testmodeler/dist/src/common/constant';
 import { Step, Variable } from 'oav/dist/lib/apiScenario/apiScenarioTypes';
+import { assert } from 'console';
 
 export class ScenarioTestDataRender extends MockTestDataRender {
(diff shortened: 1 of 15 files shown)


packages/autorest.gotest/src/generator/scenarioTestGenerator.ts
```diff
@@ -15,14 +15,14 @@ import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
 import { MockTestDataRender } from './mockTestGenerator';
 import { OavStepType } from '@autorest/testmodeler/dist/src/common/constant';
 import { Step, Variable } from 'oav/dist/lib/apiScenario/apiScenarioTypes';
+import { assert } from 'console';
```
changlong-liu, line 18: do we need this import?

changlong-liu (approved):

Issue #892, opened 2022-10-20
module name and module version should not be generated if generating a sub-package


tadelesh:
Currently, all the mgmt. packages will always generate module name and module version.
This behavior will cause generating in-sub-package module name and module version when we generate sub-packages of one profile module, which is not correct.
The module name and module version are used in the pipeline constructor and then will be used in the telemetry data of each HTTP call. For sub-package, we believe module name and module version should be defined in the module root and should be public that each sub-package could use to build pipeline. 
For generator, we need to introduce some config to help to get the root module name and module version.

tadelesh:
@jhendrixMSFT I'm going to add a config of `sub-package`. With the existed `module` config, I could get the fully info of the package path. And I will suppose there is a `ModuleName` and `ModuleVersion` const defined in the module root package. So I could use them with the pipeline creation. What do you think?

tadelesh:
We could use transform to do it. Close for now.

PR #894, opened 2022-10-28
Support generating sub package

tadelesh:
Resolve: #892
When generating sub package, go.mod will not be generated, module name and module version constants will not be generated for management plane and will use user-defined `internal.ModelName` and `internal.ModelVersion` instead.


src/generator/operations.ts
```diff
@@ -207,12 +207,22 @@ export async function generateOperations(session: Session<CodeModel>): Promise<O
...
-      clientText += "\tif err != nil {\n"
+      const subPackage = await session.getValue('sub-package', '');
+      if (subPackage === '') {
+        clientText += "\tpl, err := armruntime.NewPipeline(moduleName, moduleVersion, credential, runtime.PipelineOptions{}, options)\n";
+      } else {
+        const modName = await session.getValue('module', '');
+        if (modName === '') {
+          throw new Error(`cannot generate sub package without module specific`);
+        }
+        imports.add(`${modName}/internal`)
```
jhendrixMSFT, line 222: We shouldn't be generating code that depends on an unspecified requirement.

jhendrixMSFT (changes requested):

jhendrixMSFT:
Generally speaking, I think it would be preferable to support this outside of the code generator.  Can't we fix up the module name/version constants via a transform?

Issue #896, opened 2022-11-01
Cannot generate body setter for consumes: `application/x-www-form-urlencoded`


tadelesh:
When generating ACR package, it happens that some operation's param could not be set to HTTP body. It seems current codegen could not handle form data consturction (consumes: `application/x-www-form-urlencoded`). Example: https://github.com/Azure/azure-rest-api-specs/blob/195cd610db0accd0422c3e00a72df739ab4de677/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json#L1066.

PR #898, opened 2022-11-02
Add support for form data request

tadelesh:
Resolve: #896. It seems only ACR data plane swagger has such pattern.


src/generator/operations.ts
```diff
@@ -947,6 +947,30 @@ function createProtocolRequest(group: OperationGroup, op: Operation, imports: Im
     text += '\t\treturn nil, err\n';
     text += '\t}\n';
     text += '\treturn req, nil\n';
+  } else if (mediaType === 'form') {
```
jhendrixMSFT, line 950: Can this be merged with the `multipart` section or does it have to be separate?

tadelesh: It has to be separate. According to the w3 [spec](https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4), the body encoding method for `application/x-www-form-urlencoded` and `multipart/form-data are different.

jhendrixMSFT (approved):

PR #902, opened 2023-01-04
Use pointer-to-type for metadata collections

jhendrixMSFT:
This is cruft from an earlier design that was missed when we changed map's value types to be pointer-to-type.
Makes metadata collections consistent across headers and bodies.


jhendrixMSFT:
Related to https://github.com/Azure/azure-sdk-for-go/issues/19688

chlowell (approved):

PR #903, opened 2023-01-04
Update CI to latest Ubuntu image

jhendrixMSFT:
18.04 is being deprecated.

### DIFF
--- a/eng/pipelines/ci-gotest.yml
+++ b/eng/pipelines/ci-gotest.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         Linux_Go119:
-          vm.image: "ubuntu-18.04"
+          vm.image: "ubuntu-20.04"
           go.version: "1.19.1"
 
     pool:
--- a/eng/pipelines/ci-gotest.yml
+++ b/eng/pipelines/ci-gotest.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         Linux_Go119:
-          vm.image: "ubuntu-20.04"
+          vm.image: "ubuntu-22.04"
           go.version: "1.19.1"
 
     pool:
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         Linux_Go119:
-          vm.image: "ubuntu-18.04"
+          vm.image: "ubuntu-20.04"
           go.version: "1.19.1"
 
     pool:
--- a/eng/pipelines/ci.yml
+++ b/eng/pipelines/ci.yml
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         Linux_Go119:
-          vm.image: "ubuntu-20.04"
+          vm.image: "ubuntu-22.04"
           go.version: "1.19.1"
 
     pool:
--- a/eng/pipelines/publish-dev-release.yml
+++ b/eng/pipelines/publish-dev-release.yml
@@ -2,7 +2,7 @@ trigger: none
 pr: none
 
 pool:
-  vmImage: "ubuntu-20.04"
+  vmImage: "ubuntu-22.04"
 
 steps:
   - template: ../steps/set-env.yaml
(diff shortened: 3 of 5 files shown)


eng/pipelines/ci-gotest.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         Linux_Go119:
-          vm.image: "ubuntu-18.04"
+          vm.image: "ubuntu-20.04"
```
chlowell, line 26: Is 22.04 available? It's the latest LTS

jhendrixMSFT: Good call, done.

chlowell (approved):

jhendrixMSFT:
@benbp this has been consistently failing reading various content from `https://raw.githubusercontent.com/Azure/azure-rest-api-specs`.  Is this due to throttling?  Is there some way we can cache this stuff?

jhendrixMSFT:
@weshaggard pinging you since Ben is out.  I'm at a loss as to why this constantly fails.  Could it be throttling?

weshaggard (approved):

weshaggard:
> @weshaggard Wes Haggard FTE pinging you since Ben is out. I'm at a loss as to why this constantly fails. Could it be throttling?

It doesn't look to be a throttling issue to me. I know with the move to Ubuntu22 they have updated and removed some toolsets so I wonder if there is something different with the networking stack in the new run.

weshaggard:
Well I'm guessing it isn't a agent version issue because there appears to be similar errors in an earlier run with 18, https://dev.azure.com/azure-sdk/public/_build/results?buildId=2088776&view=results. 

Do you have any idea of how many times that URL is being pulled? It could be a rate limiting issue but the error messages don't seem to be very helpful so it is hard to say for sure.

weshaggard:
If they are using the same token then it could contribute to the limit. If they aren't using a token at all and are anonymous I believe github uses IP or IP-ranges to determine limit so it could potentially impact depending on the agent machines used.

jhendrixMSFT:
Here's a fresh failure.

https://dev.azure.com/azure-sdk/public/_build/results?buildId=2104770&view=results

We aren't using a token.  Do you think it would help?

Is there any way we can cache the downloaded content?  While it does change on occasion, it's pretty infrequent.

weshaggard:
If the issue is throttling then yes using a token will help up the limit. 

If you wish to cache the content that would have to be handled in the code that is pulling from the URL. There isn't any magic proxy we can put in place that will just cache that content.

jhendrixMSFT:
Thanks, I'll try a token then.  Is there a certain way I should be creating one?

jhendrixMSFT:
We could clone but there are a few problems with that.
- The specs repo is huge, and we only need a small portion of it.
- We reference content from a fixed commit so the tests are repeatable, and you can't clone from a specific commit.
- Related, while we could clone and then check out the desired commit, we can no longer shallow clone.

Seems like a token might be the better choice (please correct me if I got the above wrong).

weshaggard:
Those are fair issues. A token might help but it you are still doing a lot of calls it only helps somewhat and it might still be worth caching the files in the code somehow.

jhendrixMSFT:
I'm following up with the Autorest folks to see if there's a way we can cache the temporary content it creates when referencing specs via HTTPS.

What would be considered "a lot of calls"?  In CI we have 10 tests that each pull down one or more files from the specs repo.  Let's be really generous and say it's a total of 100 files (I'm assuming each file fetched counts towards rate limiting).  Is this a lot?

weshaggard:
I wouldn't consider 10 or even 100 to be an issue. Perhaps it is something other than this as I cannot tell from the errors that this is the real issue. Perhaps some more error handling should be added first to try and be sure we are fixing the correct problem.

weshaggard:
Quoting from github docs:
> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting

So if the tests are doing pull that go over 60 per hour then it could be throttling. Either way we should still add some extra error handling and confirm and then consider options.

PR #907, opened 2023-01-11
Add --rawjson-as-bytes option

jhendrixMSFT:
When enabled (off by default), schema types Any and AnyObject are emitted as []byte indicating they're raw JSON.

### DIFF
--- a/.scripts/regeneration.js
+++ b/.scripts/regeneration.js
@@ -17,7 +17,7 @@ const goMappings = {
     'binarygroup': ['body-binary.json', '--remove-unreferenced-types'],
     'booleangroup': ['body-boolean.json', '--remove-unreferenced-types'],
     'bytegroup': ['body-byte.json', '--remove-unreferenced-types'],
-    'complexgroup': ['body-complex.json', '--remove-unreferenced-types'],
+    'complexgroup': ['body-complex.json', '--remove-unreferenced-types', '--rawjson-as-bytes'],
     'complexmodelgroup': ['complex-model.json', '--remove-unreferenced-types'],
     'custombaseurlgroup': ['custom-baseUrl.json', '--remove-unreferenced-types'],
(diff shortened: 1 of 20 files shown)


.scripts/regeneration.js
```diff
@@ -17,7 +17,7 @@ const goMappings = {
     'binarygroup': ['body-binary.json', '--remove-unreferenced-types'],
     'booleangroup': ['body-boolean.json', '--remove-unreferenced-types'],
     'bytegroup': ['body-byte.json', '--remove-unreferenced-types'],
-    'complexgroup': ['body-complex.json', '--remove-unreferenced-types'],
+    'complexgroup': ['body-complex.json', '--remove-unreferenced-types', '--rawjson-as-bytes'],
```
jhendrixMSFT, line 20: This was enabled here, and a few others, where no change was expected in the generated code.

src/transform/transform.ts
```diff
@@ -219,16 +222,32 @@ function substitueDiscriminator(item: Property | Parameter | SchemaResponse): Sc
...
+        schema.language.go!.rawJSONAsBytes = rawJSONAsBytes;
+        return '[]byte';
+      }
       return 'map[string]any';
     case SchemaType.Array:
       const arraySchema = <ArraySchema>schema;
-      arraySchema.language.go!.elementIsPtr = !isTypePassedByValue(arraySchema.elementType);
       const arrayElem = <Schema>arraySchema.elementType;
+      if (rawJSONAsBytes && (arrayElem.type === SchemaType.Any || arrayElem.type === SchemaType.AnyObject)) {
+        return 'any';
```
jhendrixMSFT, line 248: Hmm this looks wrong.

jhendrixMSFT: Fixed.

chlowell (approved):

PR #910, opened 2023-01-18
Use clients from azcore

jhendrixMSFT:
Generated clients now contain an internal client from azcore. Removed client constructors for non-ARM targets.
Error on optional client parameters for ARM.
Exporting clients is now on by default.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/867

Issue #823, opened 2022-05-11
Provide one client for all operations in one package for control plane


tadelesh:
Some possible problems:
1. lazy load or create all sub clients when initializing
2. client param need include all sub clients param
3. non-breaking change

tadelesh:
Some prototype [here](https://github.com/tadelesh/azure-sdk-for-go/blob/prototype/sdk/resourcemanager/resources/armresources/resource_manager_client.go).

Two possible ways to resolve client param besides subscriptionID, cred and options (for now, only three mgmt. RPs have such issue): one is to add to the get method of `ResourceManagerClient`, another is moving extra client params into operation (breaking change).

```go
type ResourceManagerClient struct {
	subscriptionID string
	host           string
	pl             runtime.Pipeline

	resourcesClient             *Client
	deploymentOperationClient   *DeploymentOperationsClient
	...
}

func NewResourceManagerClient(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*ResourceManagerClient, error) {
	if options == nil {
		options = &arm.ClientOptions{}
	}
	ep := cloud.AzurePublicCloud.Services[cloud.ResourceManager].Endpoint
	if c, ok := options.Cloud.Services[cloud.ResourceManager]; ok {
		ep = c.Endpoint
	}
	pl, err := armruntime.NewPipeline(moduleName, moduleVersion, credential, runtime.PipelineOptions{}, options)
	if err != nil {
		return nil, err
	}
	return &ResourceManagerClient{
		subscriptionID: subscriptionID,
		host:           ep,
		pl:             pl,
	}, nil
}

func (mgmtClient *ResourceManagerClient) Resources() *Client {
	if mgmtClient.resourcesClient == nil {
		mgmtClient.resourcesClient = &Client{
			subscriptionID: mgmtClient.subscriptionID,
			host:           mgmtClient.host,
			pl:             mgmtClient.pl,
		}
	}
	return mgmtClient.resourcesClient
}

func (mgmtClient *ResourceManagerClient) DeploymentOperation() *DeploymentOperationsClient {
	if mgmtClient.deploymentOperationClient == nil {
		mgmtClient.deploymentOperationClient = &DeploymentOperationsClient{
			subscriptionID: mgmtClient.subscriptionID,
			host:           mgmtClient.host,
			pl:             mgmtClient.pl,
		}
	}
	return mgmtClient.deploymentOperationClient
}

...
```

[...]

PR #912, opened 2023-02-13
Generate client factory for arm

tadelesh:
Resolve: https://github.com/Azure/autorest.go/issues/823.
For all 2364 mgmt. clients, there are only five modules with 42 clients that have additional client parameters besides subscription ID. After reviewed all of them, I believe most of them are from wrong swagger definition, and I will double check with service team and use directive to fix them in their next version.


tadelesh:
There is one thing I did not achieve for now: deep copy options before client creation check. My idea is to add the function to `azcore/arm` package.

test/compute/2019-12-01/armcompute/zz_client_factory.go
```diff
@@ -0,0 +1,188 @@
...
+// DO NOT EDIT.
+
+package armcompute
+
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+)
+
+// ClientFactory is a factory help to create all clients in this modules.
```
jhendrixMSFT, line 17:
```go
// ClientFactory is a client factory used to create any client in this module.
// Don't use this type directly, use NewClientFactory instead.
```

test/compute/2019-12-01/armcompute/zz_client_factory.go
```diff
@@ -0,0 +1,188 @@
...
+type ClientFactory struct {
+	subscriptionID string
+	credential     azcore.TokenCredential
+	options        *arm.ClientOptions
+}
+
+// NewClientFactory creates a new instance of ClientFactor with the specified values.
```
jhendrixMSFT, line 24:
```go
// NewClientFactory creates a new instance of ClientFactory with the specified values.
// The parameter values will be propagated to any client created from this factory.
```

jhendrixMSFT (changes requested):
Some minor refinements on the doc comments.

jhendrixMSFT:
@JeffreyRichter here is the client factory codegen.

JeffreyRichter:
> There is one thing I did not achieve for now: deep copy options before client creation check because it is not guaranteed that nothing change for options after check. My idea is to add a deep copy function for options to `azcore/arm` package.

I don't think we should need this. Just make a copy with *options because it shouldn't contain any pointers. Plus, client ctors should all be making copies to ensure that any customer code modifications to an options structure does NOT impact any clients already constructed.

tadelesh:
> > There is one thing I did not achieve for now: deep copy options before client creation check because it is not guaranteed that nothing change for options after check. My idea is to add a deep copy function for options to `azcore/arm` package.
> 
> I don't think we should need this. Just make a copy with *options because it shouldn't contain any pointers. Plus, client ctors should all be making copies to ensure that any customer code modifications to an options structure does NOT impact any clients already constructed.

There are some slices inside the options (e.g., `PerCallPolicies`, `PerRertryPolicies`). Though changing the element in a slice should be a rare usage, it could be possible that options will be changed after factory created.

test/databoxedge/2021-02-01/armdataboxedge/zz_client_factory.go
```diff
@@ -32,106 +34,106 @@ func NewClientFactory(subscriptionID string, credential azcore.TokenCredential,
 	}
 	return &ClientFactory{
 		subscriptionID: subscriptionID, credential: credential,
-		options: options,
+		options: *options,
```
jhendrixMSFT, line 37: This will panic if `options` is `nil`.

JeffreyRichter:
> > > There is one thing I did not achieve for now: deep copy options before client creation check because it is not guaranteed that nothing change for options after check. My idea is to add a deep copy function for options to `azcore/arm` package.
> > 
> > 
> > I don't think we should need this. Just make a copy with *options because it shouldn't contain any pointers. Plus, client ctors should all be making copies to ensure that any customer code modifications to an options structure does NOT impact any clients already constructed.
> 
> There are some slices and map inside the options (e.g., `PerCallPolicies`, `PerRertryPolicies`, `Services` in `Cloud.Configuration`). Though changing the element in a slice or delete a map element should be a rare usage for options, it could be possible after factory created. And the change of cloud configuration bypass the check in factory CTOR. Shall we handle it?

Yes, you're right. make a deep copy.

test/databoxedge/2021-02-01/armdataboxedge/zz_client_factory.go
jhendrixMSFT: You can add the `nil` handling in the `Clone()` impl.

xboxeer (approved):
Had an offline discussion with Chenjie, the new API LGTM

test/compute/2019-12-01/armcompute/zz_client_factory.go
```diff
@@ -0,0 +1,190 @@
...
+//   - credential - used to authorize requests. Usually a credential from azidentity.
+//   - options - pass nil to accept the default values.
+func NewClientFactory(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
+	_, err := arm.NewClient("armcompute.ClientFactory", moduleVersion, credential, options)
+	if err != nil {
+		return nil, err
+	}
+	return &ClientFactory{
+		subscriptionID: subscriptionID, credential: credential,
+		options: options.Clone(),
```
tadelesh, line 38: Add `Clone` logic. @jhendrixMSFT please help to review again. Thanks.

jhendrixMSFT (approved):

PR #941, opened 2023-03-27
Omit tags for JSON marshalling/unmarshalling

jhendrixMSFT:
They aren't used and just cause confusion.


chlowell (approved):

ExerciseBook:
Hello, I have some problem about this pr.

Thus we delete json tag from json struct.

imagine there is a scene, the original struct is
```
{
    "a_field_name_starts_with_lowercase_letter": 1
}
```
we can not let golang process it well.

maybe we can add a configuration? let us choose omit it or not?

Issue #946, opened 2023-04-10
nil of any type serialize to null


tadelesh:
For example, this [model](https://github.com/Azure/azure-sdk-for-go/blob/610f7bd3342f04618e722ce44ca4c99702f2117c/sdk/resourcemanager/compute/armcompute/models.go#L7145) contains any type properties, even no value set to those properties, it will be serialized to `"protectedSettings": null`. It is because the reflect null value judgement [here](https://github.com/Azure/azure-sdk-for-go/blob/610f7bd3342f04618e722ce44ca4c99702f2117c/sdk/resourcemanager/compute/armcompute/models_serde.go#L18852) will not work for any/interface{} type. We may need special logic to handle it.

jhendrixMSFT:
Introduced in https://github.com/Azure/autorest.go/pull/734

PR #947, opened 2023-04-10
Fix populating any type

jhendrixMSFT:
Strcuts with fields of any type were always sending a JSON null as the address of the field was never nil.
Reverted rawjson-as-bytes for some test cases to see and test changes.

Fixes https://github.com/Azure/autorest.go/issues/946

### DIFF
--- a/test/maps/azalias/polymorphic_helpers_test.go
+++ b/test/maps/azalias/polymorphic_helpers_test.go
@@ -166,5 +166,5 @@ func TestInterfaceJSONNull(t *testing.T) {
 	b, err := json.Marshal(props1)
 	require.NoError(t, err)
 	require.Contains(t, string(b), `"anyObject":null`)
-	require.NotContains(t, string(b), `"anything":null`)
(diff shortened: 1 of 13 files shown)


test/maps/azalias/polymorphic_helpers_test.go
```diff
@@ -154,3 +158,13 @@ func TestInterfaceNil(t *testing.T) {
...
 }
+
+func TestInterfaceJSONNull(t *testing.T) {
+	props1 := TypeWithRawJSON{
+		AnyObject: azcore.NullValue[*any](),
+	}
+	b, err := json.Marshal(props1)
+	require.NoError(t, err)
+	require.Contains(t, string(b), `"anyObject":null`)
+	require.NotContains(t, string(b), `"anything":null`)
```
chlowell, line 169: Couldn't this be stricter because having `nil` value, `Anything` shouldn't be serialized?
```suggestion
	require.NotContains(t, string(b), "anything")
```

chlowell (approved):

PR #948, opened 2023-04-21
Move autorest.go code generator to packages subdirectory

jhendrixMSFT:
This makes the layout consistent with other packages and sets us up for future addition of a TypeSpec package.


jhendrixMSFT:
@benbp I renamed ci.yml to ci-go.yml to better distinguish the CI files which of course hosed part of CI.  Is there a way I can modify the pipeline so this new one is picked up?

packages/autorest.go/tsconfig.json
RickWinter, line 1: Are these changes b/c we are using a more recent node?

packages/autorest.go/test/synapse/2019-06-01/azspark/zz_models.go
RickWinter, line 1: Did you intend for this cleanup?

jhendrixMSFT: There was a recent change in the autorest front-end that introduced this delta.  Not sure why and I will look into that, but for now, the change is expected.

RickWinter (approved):

packages/autorest.go/tsconfig.json
jhendrixMSFT: No, I'm just making this current.  I "borrowed" it from the autorest.typescript tsconfig.

jhendrixMSFT: In addition, the majority of these settings were in the root tsconfig.json which has been removed.  I did find two settings I missed though and have added them back.

PR #949, opened 2023-04-21
Stop using remote github assets for regeneration

jhendrixMSFT:
Transient network errors or throttling can cause CI to become unstable. Added necessary specs and config files so assets can be locally built. Switched some packages to a more recent API version. Removed API version in output paths since we won't be building API versions side by side.


jhendrixMSFT:
@tadelesh can you help me understand the gotest failure here?

chlowell (approved):

jhendrixMSFT:
Thanks.  In the future, if a swagger is updated, how do we regenerate the content for `gotest`?

tadelesh:
> Thanks. In the future, if a swagger is updated, how do we regenerate the content for `gotest`?

Currently, I do it manually with copy and paste, I'll try to add a rush script to do so.

jhendrixMSFT:
Sounds good.  Maybe something similar to `rush regenerate` for rebuilding the generated code.

Issue #501, opened 2020-10-13
Include tracing in track 2 codegen


jhendrixMSFT:
Add support for tracing (e.g. opencensus) in track 2 codegen.

jhendrixMSFT:
Include request/response detail in span annotation, something like request url, request body, response body, etc.

jhendrixMSFT:
SDK APIs should ideally not be top-level entries, i.e. child if a parent span is available.  This way you can see how the SDK APIs are called from apps.

jhendrixMSFT:
Codegen changes will be in `preview.50` of the code generator.

PR #954, opened 2023-05-09
Insert calls to runtime.StartSpan in APIs

jhendrixMSFT:
This requries using named return values in APIs so that the deferred call to endSpan() can capture the error.


jhendrixMSFT:
Fixes https://github.com/Azure/autorest.go/issues/501

packages/autorest.go/test/autorest/go.mod
```diff
@@ -3,17 +3,26 @@ module generatortests
...
 require (
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.3.0
-	github.com/google/go-cmp v0.5.4
-	github.com/stretchr/testify v1.7.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.6.1-0.20230509175118-ce926c4cadfc
+	github.com/Azure/azure-sdk-for-go/sdk/tracing/azotel v0.0.0-00010101000000-000000000000
+	github.com/google/go-cmp v0.5.9
+	github.com/stretchr/testify v1.8.1
+	go.opentelemetry.io/otel v1.11.2
+	go.opentelemetry.io/otel/exporters/jaeger v1.11.0
```
jhendrixMSFT, line 11: Forgot to pull this out, fixing...

packages/autorest.go/src/generator/operations.ts
```diff
@@ -414,31 +416,35 @@ function generateOperation(op: Operation, imports: ImportManager): string {
...
   }
-  const zeroResp = getZeroReturnValue(op, 'op');
+  if (!isLROOperation(op) && injectSpans) {
+    text += `\tctx, endSpan := runtime.StartSpan(ctx, "${clientName}.${opName}", client.internal.Tracer(), nil)\n`;
+    text += '\tdefer func() { endSpan(err) }()\n';
+  }
   text += `\treq, err := client.${info.protocolNaming.requestMethod}(${reqParams})\n`;
   text += `\tif err != nil {\n`;
-  text += `\t\treturn ${zeroResp}, err\n`;
+  text += `\t\treturn\n`;
```
chlowell, line 425: I've seen enough concern about bare returns to think we should probably avoid them even in generated code, to forestall questions/complaints about them.

chlowell (commented):
Is this repo the best place for a simple end-to-end test verifying a generated client started and ended a span as expected? (I guess that could go in a management plane test suite instead)

packages/autorest.go/src/generator/operations.ts
jhendrixMSFT: Generally speaking, I agree with you (I'm not a fan of named return values).  We could retool this to emit the calls to `endSpan(err)` on all of the error paths, it might get a bit ugly though (although perhaps not as ugly as it is now).  Using named return values does simplify this though.  I will take a closer look.

chlowell: Couldn't we simply `return err` in these cases?

jhendrixMSFT: Unfortunately no, as `endSpan(err)` terminates the span (success or failure).

jhendrixMSFT: Here are the ways we could do it.  Not that bad really.  I think I prefer option 2 as we never have to remember to inject calls to `endSpan(err)` if/when we add any additional returns (unlikely but possible).  However, we have to be careful that we don't shadow `err`.  I don't have a strong preference though so could be convinced otherwise.
```go
func (client *AvailabilitySetsClient) CreateOrUpdate(ctx context.Context, resourceGroupName string, availabilitySetName string, parameters AvailabilitySet, options *AvailabilitySetsClientCreateOrUpdateOptions) (AvailabilitySetsClientCreateOrUpdateResponse, error) {
	ctx, endSpan := runtime.StartSpan(ctx, "AvailabilitySetsClient.CreateOrUpdate", client.internal.Tracer(), nil)
	req, err := client.createOrUpdateCreateRequest(ctx, resourceGroupName, availabilitySetName, parameters, options)
	if err != nil {
		endSpan(err)
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	resp, err := client.internal.Pipeline().Do(req)
	if err != nil {
		endSpan(err)
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	if !runtime.HasStatusCode(resp, http.StatusOK) {
		err := runtime.NewResponseError(resp)
		endSpan(err)
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	res, err := client.createOrUpdateHandleResponse(resp)
	endSpan(err)
	return res, err
}

func (client *AvailabilitySetsClient) CreateOrUpdate2(ctx context.Context, resourceGroupName string, availabilitySetName string, parameters AvailabilitySet, options *AvailabilitySetsClientCreateOrUpdateOptions) (AvailabilitySetsClientCreateOrUpdateResponse, error) {
	var err error
	ctx, endSpan := runtime.StartSpan(ctx, "AvailabilitySetsClient.CreateOrUpdate", client.internal.Tracer(), nil)
	defer func() { endSpan(err) }()
	var req *policy.Request
	req, err = client.createOrUpdateCreateRequest(ctx, resourceGroupName, availabilitySetName, parameters, options)
	if err != nil {
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	var resp *http.Response
	resp, err = client.internal.Pipeline().Do(req)
	if err != nil {
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	if !runtime.HasStatusCode(resp, http.StatusOK) {
		err = runtime.NewResponseError(resp)
		return AvailabilitySetsClientCreateOrUpdateResponse{}, err
	}
	var res AvailabilitySetsClientCreateOrUpdateResponse
	res, err = client.createOrUpdateHandleResponse(resp)
	return res, err
}
```

jhendrixMSFT: I overthought option 2 a bit.  We don't need to declare all vars up front, just `err`.  So I'll go with that.

jhendrixMSFT:
Will add tests to verify spans are properly ended.

packages/autorest.go/test/acr/azacr/zz_authentication_client.go
```diff
@@ -103,18 +108,23 @@ func (client *AuthenticationClient) exchangeAADAccessTokenForAcrRefreshTokenHand
...
 func (client *AuthenticationClient) ExchangeAcrRefreshTokenForAcrAccessToken(ctx context.Context, service string, scope string, refreshToken string, options *AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenOptions) (AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenResponse, error) {
+	var err error
+	ctx, endSpan := runtime.StartSpan(ctx, "AuthenticationClient.ExchangeAcrRefreshTokenForAcrAccessToken", client.internal.Tracer(), nil)
+	defer func() { endSpan(err) }()
 	req, err := client.exchangeAcrRefreshTokenForAcrAccessTokenCreateRequest(ctx, service, scope, refreshToken, options)
 	if err != nil {
 		return AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenResponse{}, err
 	}
-	resp, err := client.internal.Pipeline().Do(req)
+	httpResp, err := client.internal.Pipeline().Do(req)
```
chlowell, line 118:
```suggestion
	var httpResp *http.Response
	httpResp, err = client.internal.Pipeline().Do(req)
```

packages/autorest.go/test/acr/azacr/zz_authentication_client.go
```diff
@@ -103,18 +108,23 @@ func (client *AuthenticationClient) exchangeAADAccessTokenForAcrRefreshTokenHand
...
 		return AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenResponse{}, err
 	}
-	if !runtime.HasStatusCode(resp, http.StatusOK) {
-		return AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenResponse{}, runtime.NewResponseError(resp)
+	if !runtime.HasStatusCode(httpResp, http.StatusOK) {
+		err = runtime.NewResponseError(httpResp)
+		return AuthenticationClientExchangeAcrRefreshTokenForAcrAccessTokenResponse{}, err
 	}
-	return client.exchangeAcrRefreshTokenForAcrAccessTokenHandleResponse(resp)
+	resp, err := client.exchangeAcrRefreshTokenForAcrAccessTokenHandleResponse(httpResp)
```
chlowell, line 126: Shadowing here as well

packages/autorest.go/test/acr/azacr/zz_authentication_client.go
jhendrixMSFT: This is what tripped me up earlier with my option 1/2 thing.  Since `err` is in the same scope it isn't shadowed.  To keep us honest, I've updated CI to check for this.

chlowell: 🤦 right, my bad, it's the case in which `:=` combines short variable declaration with reassignment. Good idea, asking a computer to keep an eye on this

chlowell (approved):

PR #955, opened 2023-05-09
Add fakes subpackage

jhendrixMSFT:
Disabled by default, add --generate-fakes to enable. Consolidated LRO/pageable naming.


packages/autorest.go/src/fake/servers.ts
```diff
@@ -7,11 +7,11 @@ import { Session } from '@autorest/extension-base';
 import { ArraySchema, ByteArraySchema, ChoiceSchema, CodeModel, ConstantSchema, DateTimeSchema, GroupProperty, ImplementationLocation, NumberSchema, OperationGroup, Operation, Parameter, SchemaType, SerializationStyle } from '@autorest/codemodel';
 import { capitalize, uncapitalize } from '@azure-tools/codegen';
 import { values } from '@azure-tools/linq';
-import { aggregateParameters, formatConstantValue, getSchemaResponse, isBinaryResponseOperation, isLROOperation, isPageableOperation, isTypePassedByValue } from '../common/helpers';
+import { aggregateParameters, formatConstantValue, getSchemaResponse, isBinaryResponseOperation, isLROOperation, isPageableOperation } from '../common/helpers';
```
RickWinter, line 10: Why two entries for the imports from ../common/helpers?  We should import all the types we use from the same package in one import statement.

RickWinter:
Why did all the resp rspError change names to respr and resprError?

packages/autorest.go/src/fake/servers.ts
jhendrixMSFT: Fixed.

chlowell (approved):

Issue #951, opened 2023-05-03
Add custom extension to exclude generated marshallers/unmarshallers


jhendrixMSFT:
This is required to support a new feature in azblob where blob names are base-64 encoded.

jhendrixMSFT:
Also fix XML handling of `,chardata`

jhendrixMSFT:
Fixed in `preview.49` of the code generator.

PR #957, opened 2023-05-15
Add feature to omit generated SerDe methods

jhendrixMSFT:
Adding extension x-ms-go-omit-serde-methods to a type will cause the code generator to skip emitting SerDe methods.
Fixed handling of XML text data (x-ms-text extension).

Fixes https://github.com/Azure/autorest.go/issues/951


packages/autorest.go/src/generator/structs.ts
```diff
@@ -182,6 +182,9 @@ export function getXMLSerialization(prop: Property, lang: Language): string {
   if (prop.schema.serialization?.xml?.name) {
     // xml can specifiy its own name, prefer that if available
     serialization = prop.schema.serialization.xml.name;
+  } else if (prop.schema.serialization?.xml?.text) {
+    // see https://github.com/Azure/autorest/tree/main/docs/extensions#x-ms-text
+    serialization = ',chardata';
```
chlowell, line 187: Is the link supposed to explain this value? I'm still wondering about the leading comma, and what "chardata" means

jhendrixMSFT: It explains where the value of `text` comes from (the `x-ms-text` extension).  The value `,chardata` is what the in-box [XML parser](https://pkg.go.dev/encoding/xml#Unmarshal) uses to indicate if a field is raw character data.  I can update the comment to clarify this.

chlowell (approved):

PR #959, opened 2023-06-02
Generated file headers trigger edit warning

chlowell:
Go tools use a conventional regex to identify generated code. This ensures our generated files match, so for example the Go extension for VS Code will warn when editing them.


packages/autorest.go/test/synapse/azspark/zz_models.go
```diff
@@ -207,7 +206,9 @@ type Session struct {
 	// Dictionary of
 	AppInfo    map[string]*string
 	ArtifactID *string
-	Errors     []*ServiceError
+
+	// The error information.
+	Errors []*ServiceError
```
chlowell, line 211: I don't know why we didn't have this comment before, but I get the same diff on main.

jhendrixMSFT: Hmm could this come from the version of autorest being used in CI?  We don't pin to a specific version.  If you update your autorest installation to latest and regenerate `azspark` does the comment show up?

chlowell: Hey, that did the trick 🏆

packages/autorest.go/src/generator/helpers.ts
```diff
@@ -20,9 +20,16 @@ export async function contentPreamble(session: Session<CodeModel>, packageName?:
...
-  text += `// +build go1.18\n\n${headerText}\n// DO NOT EDIT.\n\n`;
-  text += `package ${packageName}\n\n`;
+  let text = `//go:build go1.18\n// +build go1.18\n\n`
+  // ensure tools recognize the file as generated according to
+  // https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source
+  text += headerText.replace(/^\/\/ Code generated .*\.$/m, '$& DO NOT EDIT.');
+  if (text.match(/^\/\/ Code generated .* DO NOT EDIT\.$/m)) {
+    text += "\n"
+  } else {
+    text += "// Code generated by @autorest/go. DO NOT EDIT.\n\n"
```
jhendrixMSFT, line 30: Is there a case where this branch is executed or is this just defensive?

jhendrixMSFT (approved):

packages/autorest.go/src/generator/helpers.ts
chlowell: We land here when there's no license header or the header doesn't have a "Code generated ..." line. I see the same MIT header everywhere so practically speaking it's probably just future proofing.

Looking at this again, I saw it could be simplified by moving a newline.

PR #964, opened 2023-06-08
Move to latest LTS of node

jhendrixMSFT:


jhendrixMSFT:
@tadelesh can you please take a look at the autorest.gotest failures?

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/__debug/go-tester-pre.yaml
```diff
@@ -7721,7 +7721,7 @@ operationGroups:
                 resourceGroupName: myResourceGroup
                 serviceName: myservice
                 subscriptionId: 00000000-0000-0000-0000-000000000000
-              x-ms-original-file: file:///D:/Workspace/tadelesh/autorest.go/swagger/specification/appplatform/resource-manager/Microsoft.AppPlatform/preview/2020-11-01-preview/examples/Services_Get.json
+              x-ms-original-file: file:///C:/git/Azure/autorest.go/swagger/specification/appplatform/resource-manager/Microsoft.AppPlatform/preview/2020-11-01-preview/examples/Services_Get.json
```
chlowell, line 7724: Does anything use this path?

jhendrixMSFT: The test ignores these lines when it [diffs](https://github.com/Azure/autorest.go/blob/main/packages/autorest.gotest/test/integrationtest/testAutorest.ts#LL8C91-L8C91) the assets.

chlowell: Then why do they need to be apparently real paths? I guess these values serve some purpose or you wouldn't have changed them. I'm just looking at this thinking there must be a better way to serve that purpose because this way is so fragile.

jhendrixMSFT: I think it's just part of the yaml spew created by the testgen tool.  Presumably this maps back to the full path of the input file when the tool is executed.  From the path, I'm guessing that the testgen tool uses the JSON examples for an OpenAPI spec as part of its input.

chlowell: Oh, I think I get it now--all this yaml is generated by a tool you ran after upgrading Node and the consumer of the yaml doesn't require these paths to exist. Maybe there's a better way for the generator to handle paths but it's orthogonal to upgrading Node and nothing is broken by the paths changing every time someone runs the generator.

chlowell (approved):

PR #967, opened 2023-06-12
Update to latest core/m4

jhendrixMSFT:
This requires updating gotest to the latest testmodeler which uses a compatible version of core.
Added support for new SchemaType ArmId.
Added missing import for io in some cases.


jhendrixMSFT:
@tadelesh can you please update this PR with a fix for the jest build failures?  I believe it has something to do with an improperly set `moduleDirectories` but I haven't been able to resolve it (I'm not familiar with jest).

tadelesh:
@jhendrixMSFT I've temporally resolved the problem that could unblock your release. The latest version of testmodeler has some pack issue. I will work with @raych1 to resolve it in another PR.

jhendrixMSFT:
@tadelesh thanks for the fast turnaround!

packages/autorest.go/test/autorest/mediatypesgroupwithnormailzedoperationname/zz_constants.go
```diff
@@ -52,41 +52,41 @@ func PossibleContentType1Values() []ContentType1 {
 	}
 }
 
-// ContentType1AutoGenerated - Content type for upload
-type ContentType1AutoGenerated string
+// ContentType2 - Content type for upload
+type ContentType2 string
```
chlowell, line 56: I'm curious whether we could see a change like this in a module we've shipped but I suppose if we did, we'd simply add a transform to keep the released name.

chlowell (approved):

packages/autorest.go/test/autorest/mediatypesgroupwithnormailzedoperationname/zz_constants.go
jhendrixMSFT: Fortunately, APIs that specify multiple content types are very rare so I don't expect this to be a problem in practice.  You're correct though that if this were to become a problem we can rename via a transform (which is unfortunate as the new names a better IMO).

jhendrixMSFT:
@tadelesh terribly sorry but could you please update this again?  I lost the changes when I rebased and rerunning the tests deletes all of the content for armcompute.

tadelesh:
> @tadelesh Chenjie Shi FTE terribly sorry but could you please update this again? I lost the changes when I rebased and rerunning the tests deletes all of the content for armcompute.

fixed.

tadelesh (approved):

PR #968, opened 2023-06-13
Move fakes generator into the go-protocol pipeline

jhendrixMSFT:
Since it's a subset of code generation it doesn't need to be its own pipeline.  In addition, this will simplify upcoming changes for the input agnostic adapter.


packages/autorest.go/test/acr/azacr/fake/zz_containerregistry_server.go
```diff
@@ -213,15 +213,15 @@ func (c *ContainerRegistryServerTransport) dispatchCreateManifest(req *http.Requ
 	if err != nil {
 		return nil, err
 	}
+	if val := server.GetResponse(respr).ContentLength; val != nil {
```
jhendrixMSFT, line 216: In the constructor for `StructDef` we sort the properties.  This sorting was not preserved in the previous implementation.

PR #972, opened 2023-06-16
Split options and response type emitter

jhendrixMSFT:
Models, options, and response types all share the same emitter.  Nice in theory, but there are enough subtle differences that it's become a bit of spaghetti code.
Options and response types now have their own emitters.  The emitter for models will be cleaned up in a separate change.
Added doc comment for the embedded result type in a response envelope.


packages/autorest.go/src/generator/models.ts
```diff
@@ -35,19 +34,19 @@ export async function generateModels(session: Session<CodeModel>): Promise<model
   let needsJSONPopulateByteArray = false;
   let needsJSONPopulateAny = false;
   let serdeTextBody = '';
-  structs.sort((a: StructDef, b: StructDef) => { return sortAscending(a.Language.name, b.Language.name) });
+  structs.sort((a: ModelDef, b: ModelDef) => { return sortAscending(a.Language.name, b.Language.name) });
```
RickWinter, line 37: Bit of a nit, should we call this local models and have this instead:
```
const models = generateModels(modelImports, serdeImports, session.model.schemas.objects);
```

If you agree please make the change in a follow up PR to avoid more noise in this one.

jhendrixMSFT: Good catch, will address separately.

packages/autorest.go/src/generator/models.ts
```diff
@@ -485,3 +497,153 @@ function generateAliasType(structDef: StructDef, receiver: string, forMarshal: b
   text += `\t}\n`;
   return text;
 }
+
+// represents a method on a model
```
RickWinter, line 501: Nit, This could be in its own file which is imported.  Would help simplify the amount of code in this file.

jhendrixMSFT: Given that this type is now exclusively used for models.go codegen, I wanted to keep it local to this file (it was in structs.ts before I renamed it).

packages/autorest.go/src/generator/options.ts
```diff
@@ -3,12 +3,13 @@
...
  *--------------------------------------------------------------------------------------------*/
 
+import { capitalize, comment } from '@azure-tools/codegen';
 import { Session } from '@autorest/extension-base';
-import { CodeModel, GroupProperty } from '@autorest/codemodel';
+import { CodeModel, ConstantSchema, GroupProperty, ImplementationLocation, Parameter, SchemaType } from '@autorest/codemodel';
 import { values } from '@azure-tools/linq';
-import { contentPreamble, sortAscending } from './helpers';
+import { commentLength } from '../common/helpers';
+import { contentPreamble, hasDescription, sortAscending } from './helpers';
```
RickWinter, line 11: Could helpers in generator be part of common/helpers?

jhendrixMSFT: The way this is currently architected, `./helpers.ts` contains shared code across contents in the `generator` directory, and `common/helpers` contains code shared across code in the `generator` and `transform` directories.  Note that `common/helpers` will be going away with the adapter refactor as there will no longer be common code across those directories.

RickWinter (approved):

PR #977, opened 2023-06-26
Always include time type metadata on the code model

jhendrixMSFT:
Unix time always requires the helper so don't make it conditional.

### DIFF
--- a/packages/autorest.go/src/transform/transform.ts
+++ b/packages/autorest.go/src/transform/transform.ts
@@ -275,26 +275,30 @@ function schemaTypeToGoType(codeModel: CodeModel, schema: Schema, type: 'Propert
       constSchema.valueType.language.go!.name = schemaTypeToGoType(codeModel, constSchema.valueType, type);
       return constSchema.valueType.language.go!.name;
     }
-    case SchemaType.DateTime:
-      // header/query param values are parsed separately so they don't need custom types
+    case SchemaType.DateTime: {
+      const dateTime = <DateTimeSchema>schema;
+      if (dateTime.format === 'date-time-rfc1123') {
+        schema.language.go!.internalTimeType = 'timeRFC1123';
+      } else {
+        schema.language.go!.internalTimeType = 'timeRFC3339';
+      }
       if (type === 'InBody') {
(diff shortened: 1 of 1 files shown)


packages/autorest.go/src/transform/transform.ts
```diff
@@ -275,26 +275,30 @@ function schemaTypeToGoType(codeModel: CodeModel, schema: Schema, type: 'Propert
       constSchema.valueType.language.go!.name = schemaTypeToGoType(codeModel, constSchema.valueType, type);
       return constSchema.valueType.language.go!.name;
     }
-    case SchemaType.DateTime:
-      // header/query param values are parsed separately so they don't need custom types
+    case SchemaType.DateTime: {
+      const dateTime = <DateTimeSchema>schema;
+      if (dateTime.format === 'date-time-rfc1123') {
```
chlowell, line 280: At first glance it looks like the two conditional blocks should be combined but I take it they're separate because setting `hasTimeRFC*` would make incorrect code for header and query values? I suppose I'm uncomfortable because e.g. `hasTimeRFC1123` doesn't suggest its effect, so it isn't clear to me why it shouldn't be set for header or query values.

chlowell (approved):

packages/autorest.go/src/transform/transform.ts
jhendrixMSFT: The `hasTimeRFC*` flags below indicate if we need to generate the time helpers in `time.go`.  We only need the helpers for `time.Time` types within structs.  For `time.Time` values in headers/query parameters we use `time.Time.Format()` and `time.Parse()` so the helpers aren't required.  We could always generate them (we used to do this) but I got tired of the yellow squiggles for unreferenced code.

jhendrixMSFT: In hindsight, I probably should have named the flags `generateTimeRFC*` for clarity.

jhendrixMSFT: The flags have been renamed.  I also applied the same fix for `dateType`.

PR #978, opened 2023-06-27
Make the ResumeToken a real parameter

jhendrixMSFT:
Add it to the list of parameters for the operation so it's treated like any other optional parameter.
Use conditional operator for checking a param's protocol implementation. Added a flag to the resume token instead of relying on a heuristic.

### DIFF
--- a/packages/autorest.go/src/generator/operations.ts
+++ b/packages/autorest.go/src/generator/operations.ts
@@ -510,7 +510,7 @@ function createProtocolRequest(group: OperationGroup, op: Operation, imports: Im
   } else {
     throw new Error(`no host or endpoint defined for method ${group.language.go!.clientName}.${op.language.go!.name}`);
   }
-  const hasPathParams = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http?.in === 'path'; }).any();
+  const hasPathParams = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http?.in === 'path'; }).any();
   // storage needs the client.u to be the source-of-truth for the full path.
   // however, swagger requires that all operations specify a path, which is at odds with storage.
   // to work around this, storage specifies x-ms-path paths with path params but doesn't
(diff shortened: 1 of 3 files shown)


packages/autorest.go/src/generator/operations.ts
```diff
@@ -576,7 +576,7 @@ function createProtocolRequest(group: OperationGroup, op: Operation, imports: Im
     // add query parameters
     const encodedParams = new Array<Parameter>();
     const unencodedParams = new Array<Parameter>();
-    for (const qp of values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http!.in === 'query'; })) {
+    for (const qp of values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined && each.protocol.http?.in === 'query'; })) {
```
RickWinter, line 579: We used to explicitly disallow null, now we will allow it.  If that was a bug why didn't we ever hit it?  Is this a gap in our test suite?

RickWinter (commented):
I'm not clear on why we want to accept `null` now.  Previously that was going to fail.

packages/autorest.go/src/generator/operations.ts
jhendrixMSFT: It's because all parameters from REST participate in some portion of the HTTP request, so `http` was never `null`.  Now that `ResumeToken` is a real parameter, and it doesn't go in any portion of the HTTP request, we must use the conditional operator.

packages/autorest.go/src/generator/operations.ts
```diff
@@ -671,7 +671,7 @@ function createProtocolRequest(group: OperationGroup, op: Operation, imports: Im
     text += '\truntime.SkipBodyDownload(req)\n';
   }
   // add specific request headers
-  const headerParam = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined; }).where((each: Parameter) => { return each.protocol.http!.in === 'header'; });
+  const headerParam = values(aggregateParameters(op)).where((each: Parameter) => { return each.protocol.http !== undefined; }).where((each: Parameter) => { return each.protocol.http?.in === 'header'; });
```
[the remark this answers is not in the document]
jhendrixMSFT: It can be simplified even further, and since we're already touching these lines, I've fixed all of them up.

RickWinter (approved):

PR #985, opened 2023-07-05
autorest.gotest issue

Alancere:
- subscriptionID type uuid

- mock test: module version


packages/autorest.go/src/generator/helpers.ts
```diff
@@ -575,6 +575,13 @@ export async function getParentImport(session: Session<CodeModel>): Promise<stri
   const modName = await session.getValue('module', 'none');
   const containingMod = await session.getValue('containing-module', 'none');
   if (modName !== 'none') {
+    const majorVersion = (await session.getValue('module-version', '1.0.0')).split('.')[0];
+    if (Number(majorVersion) > 1) {
+      const version = modName.substring(`github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/${clientPkg.substring(3)}/${clientPkg}`.length).match(/v\d/)?.[0];
```
jhendrixMSFT, line 580: Is this just something for testing out a change?  We shouldn't have hard-coded module identities.

tadelesh: @Alancere you should not hard code the modName, you just need to judge whether the last part of modName is vX.

tadelesh: @jhendrixMSFT for control plane package, we set module-name in swagger repo's config. we do not want to change it for every major version release. @Alancere is helping to add a logic to add the suffix `vX` to the module name.

jhendrixMSFT: So in the config, we'll have the following bits of data.
```
module: github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/foo/armfoo
module-version: v1.2.3
```
Are you saying that when `armfoo` goes to `v2.0.0` that only the `module-version` info is updated?

tadelesh: Yes. Will it have any side effects for data plane?

tadelesh (approved):

packages/autorest.go/src/generator/helpers.ts
tadelesh: @jhendrixMSFT could you help to review again?

packages/autorest.go/src/generator/helpers.ts
```diff
@@ -575,6 +575,12 @@ export async function getParentImport(session: Session<CodeModel>): Promise<stri
   const modName = await session.getValue('module', 'none');
   const containingMod = await session.getValue('containing-module', 'none');
   if (modName !== 'none') {
+    const majorVersion = (await session.getValue('module-version', '1.0.0')).split('.')[0];
+    if (Number(majorVersion) > 1) {
+      if (modName.split('/')[-1] !== 'v' + majorVersion) {
+        return modName + '/v' + majorVersion;
```
jhendrixMSFT, line 581: This is going to double up the major version number if one already exists in the module name ([example](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/resourcemanager/compute/armcompute/autorest.md)).

If we want the contract to be that `module` must _not_ contain a major version suffix then we need to verify that when reading its value and throw and error if it does.

tadelesh: Line 580 has handled such situation: if the module name has vX suffix, then use it directly.

Alancere: Hi @jhendrixMSFT, modified according to above discussion, please review.

packages/autorest.go/src/generator/helpers.ts
```diff
@@ -575,11 +575,12 @@ export async function getParentImport(session: Session<CodeModel>): Promise<stri
   const modName = await session.getValue('module', 'none');
   const containingMod = await session.getValue('containing-module', 'none');
   if (modName !== 'none') {
+    if (modName.split('/')[-1].match(/v\d/) !== null) {
```
tadelesh, line 578: The regex will miss `v12`.

jhendrixMSFT:
@chlowell can you please take a look at the changes to `autorest.go`?

.scripts/regeneration.js
```diff
@@ -170,7 +170,7 @@ function generateFromReadme(name, readme, tag, outputDir, additionalArgs) {
         console.log('generating ' + readme);
         outputDir = fullPath(outputDir);
         cleanGeneratedFiles(outputDir);
-        exec('autorest --use=./packages/autorest.go ' + readme + ' --go --tag=' + tag + ' --file-prefix="zz_" --modelerfour.lenient-model-deduplication --license-header=MICROSOFT_MIT_NO_VERSION --module-version=0.1.0 --output-folder=' + outputDir + ' ' + additionalArgs + ' ' + switches.join(' '), autorestCallback(outputDir, readme));
+        exec('autorest --use=./packages/autorest.go ' + readme + ' --go --tag=' + tag + ' --file-prefix="zz_" --modelerfour.lenient-model-deduplication --license-header=MICROSOFT_MIT_NO_VERSION --output-folder=' + outputDir + ' ' + additionalArgs + ' ' + switches.join(' '), autorestCallback(outputDir, readme));
```
chlowell, line 173: Is this necessary? It appears every module still uses version 0.1.0.

chlowell (approved):
Looks good to me

jhendrixMSFT (approved):

packages/autorest.go/src/generator/constants.ts
```diff
@@ -33,10 +23,12 @@ export async function generateConstants(session: Session<CodeModel>): Promise<st
   }
   // data-plane clients must manage their own constants for these values
   if (<boolean>session.model.language.go!.azureARM) {
-    const version = await getModuleVersion(session);
+    if (session.model.language.go!.moduleVersion === '') {
+      throw new Error('--module-version is a required parameter');
+    }
```
tadelesh, line 28: You don't need this bc it has been handled in `namer.ts`.

packages/autorest.go/src/generator/gomod.ts
```diff
@@ -44,5 +44,9 @@ export async function generateGoModFile(session: Session<CodeModel>, existingGoM
     // the existing version of azcore is less than the specified version so update it
     existingGoMod = existingGoMod.replace(/github\.com\/Azure\/azure-sdk-for-go\/sdk\/azcore\s+v\d+\.\d+\.\d+(?:-[a-zA-Z0-9_.-]+)?/, azcore);
   }
+  // now check if the module name needs to be replaced due to a major version increase
+  if (!existingGoMod.match(`module ${session.model.language.go!.module}$`)) {
+    existingGoMod = existingGoMod.replace(/module \S+/, `module ${session.model.language.go!.module}`);
+  }
```
tadelesh, line 50: If `module` is empty, you should not do this.

tadelesh:
You should also update related logic in gotest.

tadelesh (approved):

tadelesh:
@jhendrixMSFT if you are ok, we are going to release a new version of autorest/go as well as autorest/gotest.

packages/autorest.go/src/generator/constants.ts
jhendrixMSFT: That was a bug.  `module-version` is not required by default but is required when `--azure-arm` is set.  I will fix the bug and add this back with a better error message.

packages/autorest.go/src/generator/gomod.ts
jhendrixMSFT: Agreed.  I have slightly refined this.

PR #986, opened 2023-07-06
Add option slice-elements-byval

jhendrixMSFT:
When enabled, slices won't be generated as pointer-to-type.


jhendrixMSFT:
@JeffreyRichter here's the change we were discussing to generate slices with their elements by value.

richardpark-msft (approved):

JeffreyRichter (approved):

PR #989, opened 2023-07-07
Sort field when populating the discriminator

jhendrixMSFT:
Properly set elementIsPtr so it matches what's actually generated.


packages/autorest.go/src/transform/transform.ts
```diff
@@ -199,6 +199,7 @@ function substitueDiscriminator(item: Property | Parameter | SchemaResponse): Sc
       if (!discriminatorSchemas.has(sliceDiscriminators)) {
         discriminatorSchema = new ArraySchema(sliceDiscriminators, 'slice of discriminators', elementSchema);
         discriminatorSchema.language.go! = discriminatorSchema.language.default;
+        discriminatorSchema.language.go!.elementIsPtr = false;
```
jhendrixMSFT, line 202: We are correctly generating this already, it just makes the metadata match the reality.

chlowell (approved):

PR #995, opened 2023-07-18
Support concurrent usage of fake pager and poller APIs

jhendrixMSFT:
Create a pager/poller state machine per combination of HTTP verb and URL path so that API calls for different resources don't overlap. Bump version number in preparation for release.

Fixes the issue reported in https://github.com/Azure/azure-sdk-for-go/issues/16613#issuecomment-1640192507

### DIFF
--- a/packages/autorest.go/src/generator/fake/internal.ts
+++ b/packages/autorest.go/src/generator/fake/internal.ts
@@ -5,23 +5,44 @@
 
 import { Session } from '@autorest/extension-base';
 import { CodeModel } from '@autorest/codemodel';
+import { values } from '@azure-tools/linq';
 import { contentPreamble } from '../helpers';
+import { ImportManager } from '../imports';
+import { isLROOperation, isPageableOperation } from '../../common/helpers';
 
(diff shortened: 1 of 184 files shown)


packages/autorest.go/src/generator/fake/internal.ts
```diff
@@ -85,3 +106,47 @@ func contains[T comparable](s []T, v T) bool {
...
+	}
+}
+
+type tracker[T any] struct {
+	items map[string]*T
+	mu sync.Mutex
+}
+
+func (p *tracker[T]) key(req *http.Request) string {
+	path := req.URL.Path
```
jhendrixMSFT, line 123: This assumes that path will be unique per invocation.  Probably always true for ARM, but is it universally true?  I'm thinking perhaps a `POST` for the same path but with differing query parameters.

jhendrixMSFT: Unfortunately we don't propagate query parameters when creating the fake poller or injecting next links for paged operations.  Will fix this in the next `azcore` beta.

chlowell (approved):

PR #998, opened 2023-07-21
Skip test instead of commenting it out

jhendrixMSFT:
Makes it easier to find tests that have been disabled and why.

### DIFF
--- a/packages/autorest.go/test/autorest/headergroup/headergroup_test.go
+++ b/packages/autorest.go/test/autorest/headergroup/headergroup_test.go
@@ -109,16 +109,15 @@ func TestHeaderParamEnum(t *testing.T) {
 	require.Zero(t, result)
 }
 
-// func TestHeaderParamExistingKey(t *testing.T) {
-// 	client := newHeaderClient(t)
-// 	result, err := client.ParamExistingKey(context.Background(), "overwrite")
-// 	if err != nil {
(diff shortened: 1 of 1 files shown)


packages/autorest.go/test/autorest/headergroup/headergroup_test.go
```diff
@@ -109,16 +109,15 @@ func TestHeaderParamEnum(t *testing.T) {
...
-// 	result, err := client.ParamExistingKey(context.Background(), "overwrite")
-// 	if err != nil {
-// 		t.Fatalf("ParamExistingKey: %v", err)
-// 	}
-// 	if !reflect.ValueOf(result).IsZero() {
-//		t.Fatal("expected zero-value result")
-//	}
-// }
+func TestHeaderParamExistingKey(t *testing.T) {
+	t.Skip("won't fix: https://github.com/Azure/azure-sdk-for-go/issues/19890")
```
RickWinter, line 113: Does this show up in the report as skipped?

RickWinter (approved):

packages/autorest.go/test/autorest/headergroup/headergroup_test.go
jhendrixMSFT: Yes.  If you look at the test report and select "Other" in the filter you can see all of the skipped tests.

chlowell (approved):

PR #999, opened 2023-07-21
Skip literal path parameters in the fake servers

jhendrixMSFT:
Literal path parameters are embedded constant values so they're never passed as an explicit parameter and can be omitted.

### DIFF
--- a/packages/autorest.go/test/autorest/azurespecialsgroup/fake/zz_skipurlencoding_server.go
+++ b/packages/autorest.go/test/autorest/azurespecialsgroup/fake/zz_skipurlencoding_server.go
@@ -243,12 +243,6 @@ func (s *SkipURLEncodingServerTransport) dispatchGetSwaggerPathValid(req *http.R
 	if s.srv.GetSwaggerPathValid == nil {
 		return nil, &nonRetriableError{errors.New("fake for method GetSwaggerPathValid not implemented")}
 	}
-	const regexStr = `/azurespecials/skipUrlEncoding/swagger/path/valid/(?P<unencodedPathParam>[!#&$-;=?-\[\]_a-zA-Z0-9~%@]+)`
-	regex := regexp.MustCompile(regexStr)
-	matches := regex.FindStringSubmatch(req.URL.EscapedPath())
-	if matches == nil || len(matches) < 1 {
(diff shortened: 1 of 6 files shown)


packages/autorest.go/test/autorest/azurespecialsgroup/fake/zz_skipurlencoding_server.go
```diff
@@ -243,12 +243,6 @@ func (s *SkipURLEncodingServerTransport) dispatchGetSwaggerPathValid(req *http.R
 	if s.srv.GetSwaggerPathValid == nil {
 		return nil, &nonRetriableError{errors.New("fake for method GetSwaggerPathValid not implemented")}
 	}
-	const regexStr = `/azurespecials/skipUrlEncoding/swagger/path/valid/(?P<unencodedPathParam>[!#&$-;=?-\[\]_a-zA-Z0-9~%@]+)`
-	regex := regexp.MustCompile(regexStr)
-	matches := regex.FindStringSubmatch(req.URL.EscapedPath())
```
jhendrixMSFT, line 248: This is effectively dead code as the matched value isn't passed to the fake.

packages/autorest.go/test/autorest/complexmodelgroup/fake/zz_complexmodel_server.go
```diff
@@ -122,7 +122,7 @@ func (c *ComplexModelServerTransport) dispatchList(req *http.Request) (*http.Res
 	const regexStr = `/subscriptions/(?P<subscriptionId>[!#&$-;=?-\[\]_a-zA-Z0-9~%@]+)/resourcegroups/(?P<resourceGroupName>[!#&$-;=?-\[\]_a-zA-Z0-9~%@]+)/Microsoft.Cache/Redis`
 	regex := regexp.MustCompile(regexStr)
 	matches := regex.FindStringSubmatch(req.URL.EscapedPath())
-	if matches == nil || len(matches) < 2 {
+	if matches == nil || len(matches) < 1 {
```
jhendrixMSFT, line 125: In this case, `subscriptionId` is a literal (`resourceGroupName` isn't).  That's why the count was decreased.

chlowell (approved):

Issue #895, opened 2022-10-28
Omit response handler and envelope for paging "next handler"


jhendrixMSFT:
E.g. in `paginggroup` we generate `nextFragmentHandleResponse()` and `PagingClientNextFragmentResponse` which are unused.  We should be able to omit those.

Issue #935, opened 2023-03-06
[Go] Support Typespec as the input for Go CodeGen


josefree:
related to https://github.com/Azure/azure-sdk-for-go-pr/issues/18

jhendrixMSFT:
We now have the `typespec-go` emitter which consumes tsp.

PR #1002, opened 2023-07-28
Decouple code generation stage from M4

jhendrixMSFT:
Removed code generator's dependency on M4, replacing it with an input agnostic abstraction layer (gocodemodel.ts).
The naming and transformation phase still interface with M4. At the end of transformation, the transformed M4 code model is converted into a GoCodeModel which is then sent to the code generator. Updated applicable references from gotest and in one case copied the old code over which still talks to M4.

Fixes https://github.com/Azure/autorest.go/issues/895
First phase of fix for https://github.com/Azure/autorest.go/issues/935


packages/autorest.go/test/autorest/paginggroup/zz_paging_client.go
```diff
@@ -1114,63 +1152,7 @@ func (client *PagingClient) getWithQueryParamsHandleResponse(resp *http.Response
...
-	}
-	reqQP := req.Raw().URL.Query()
-	reqQP.Set("api_version", apiVersion)
-	req.Raw().URL.RawQuery = reqQP.Encode()
-	req.Raw().Header["Accept"] = []string{"application/json"}
-	return req, nil
-}
-
-// nextFragmentHandleResponse handles the NextFragment response.
-func (client *PagingClient) nextFragmentHandleResponse(resp *http.Response) (PagingClientNextFragmentResponse, error) {
```
jhendrixMSFT, line 1137: These response handlers weren't used and are now omitted.

packages/autorest.go/test/autorest/paginggroup/zz_paging_client.go
```diff
@@ -381,6 +381,25 @@ func (client *PagingClient) getMultiplePagesFragmentNextLinkHandleResponse(resp
 	return result, nil
 }
 
+// nextFragmentCreateRequest creates the nextFragmentCreateRequest request.
+func (client *PagingClient) nextFragmentCreateRequest(ctx context.Context, apiVersion string, tenant string, nextLink string) (*policy.Request, error) {
```
jhendrixMSFT, line 385: These pager helpers were moved next to their callers.

packages/autorest.go/test/autorest/paginggroup/zz_response_types.go
```diff
@@ -97,18 +97,3 @@ type PagingClientGetSinglePagesResponse struct {
 type PagingClientGetWithQueryParamsResponse struct {
 	ProductResult
 }
-
-// PagingClientNextFragmentResponse contains the response from method PagingClient.NewNextFragmentPager.
-type PagingClientNextFragmentResponse struct {
```
jhendrixMSFT, line 102: Useless types from the internal paging next methods.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT, line 1: This is the abstraction layer.

packages/autorest.go/test/autorest/azurespecialsgroup/zz_header_client.go
```diff
@@ -114,11 +114,10 @@ func (client *HeaderClient) customNamedRequestIDHeadCreateRequest(ctx context.Co
 
 // customNamedRequestIDHeadHandleResponse handles the CustomNamedRequestIDHead response.
 func (client *HeaderClient) customNamedRequestIDHeadHandleResponse(resp *http.Response) (HeaderClientCustomNamedRequestIDHeadResponse, error) {
-	result := HeaderClientCustomNamedRequestIDHeadResponse{}
+	result := HeaderClientCustomNamedRequestIDHeadResponse{Success: resp.StatusCode >= 200 && resp.StatusCode < 300}
```
jhendrixMSFT, line 117: Switching to type predicates made it cleaner to inline population of `Success`.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -0,0 +1,1257 @@
...
+
+  attribute: boolean;
+
+  text: boolean;
+}
+
+export type ParameterLocation = 'client' | 'method';
+
+export interface ParameterAnnotations {
+  httpStyle?: ',' | '|' | ' ' | '\\t'
```
jhendrixMSFT, line 364: This should be refactored into its own parameter type.

jhendrixMSFT: Improved, but could use further refinement.

jhendrixMSFT:
@tadelesh can you please take a look at the gotest failures?  I don't think this is a simply the case of the baselines needing to be updated.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -0,0 +1,1252 @@
...
+    this.interface = iface;
+    this.annotations = annotations;
+  }
+}
+
+export class InterfaceType implements InterfaceType {
+  constructor(name: string, discriminatorField: string) {
+    this.name = name;
+    this.discriminatorField = discriminatorField;
+    // possibleTypes and rootType are required. however, we have a chicken-and-egg
```
chlowell, line 1011: Should this comment be on the class, so it's more visible?

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -0,0 +1,1252 @@
...
+
+// Struct describes a vanilla struct definition (pretty much exclusively used for parameter groups/options bag types)
+// UDTs that are sent/received are modeled as ModelType.
+export interface StructType {
+  name: string;
+
+  description?: string;
+
+  // there are only a few corner-cases where a struct has no fields
+  fields?: Array<StructField>;
```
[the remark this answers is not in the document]
jhendrixMSFT: The only thing I was trying to indicate are the cases where it's legal for an array to contain no items, forcing you to programmatically handle that case.  That said, we aren't at present throwing any errors when an array that _shouldn't_ be empty is (maybe we should be).

packages/autorest.gotest/src/util/codegenBridge.ts
```diff
@@ -115,3 +115,64 @@ export function getSchemaResponse(op: Operation): SchemaResponse | undefined {
...
+      // NOTE: this check must come last as non-required optional constants
+      // in header/query params get dumped into the optional params group
+      continue;
+    }
+    params.push(param);
+  }
+  // move global optional params to the end of the slice
+  params.sort(sortParametersByRequired);
+  // add any parameter groups.  optional groups go last
+  paramGroups.sort((a: GroupProperty, b: GroupProperty) => {
```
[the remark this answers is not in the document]
jhendrixMSFT: Good catch, fixed.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT: I moved it to the class ctor.

tadelesh:
> @tadelesh can you please take a look at the gotest failures? I don't think this is simply the case of the baselines needing to be updated.

Since the stored code model has been changed, gotest could not get the original model info. Is that possible to keep the original code model and put the new go code model as an extra info? I don't think I could change to the new go code model for test generation very quickly. Or you may just ignore the failure temporally.

lirenhe:
@jhendrixMSFT,  please hold the merge of this PR as this would impact the test generation for all mgmt. SDK libraries.
We need to figure the action needed for this effort.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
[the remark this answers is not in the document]
jhendrixMSFT: @richardpark-msft given your JS background, do you have any thoughts on this?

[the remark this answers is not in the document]
richardpark-msft: An empty array sounds better - otherwise you have to check both it being null/undefined _and_ if it's empty. Might as well save the check.

From what you described it does sound like you're worried some places might not be checking, so it blowing up because of a null/undefined is better. Is that right?

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -0,0 +1,1252 @@
...
+  } else {
+    throw new Error(`unhandled type ${typeof(type)}`);
+  }
+}
+
+///////////////////////////////////////////////////////////////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////////////////
+
+export class Info implements Info {
+  constructor(title: string) {
```
richardpark-msft, line 825: JS has such nice object-literal initialization. Do you even need constructors?

packages/autorest.go/src/generator/gomod.ts
```diff
@@ -20,9 +19,12 @@ export async function generateGoModFile(session: Session<CodeModel>, existingGoM
 
   // here we specify the minimum version of azcore as required by the code generator.
   // the version can be overwritten by passing the --azcore-version switch during generation.
-  const version = await session.getValue('azcore-version', '1.6.1');
-  if (!version.match(/^\d+\.\d+\.\d+(?:-[a-zA-Z0-9_.-]+)?$/)) {
-    throw new Error(`azcore version ${version} must in the format major.minor.patch[-beta.N]`);
+  let version = '1.6.1';
+  if (codeModel.options.azcoreVersion) {
+    if (!codeModel.options.azcoreVersion.match(/^\d+\.\d+\.\d+(?:-[a-zA-Z0-9_.-]+)?$/)) {
```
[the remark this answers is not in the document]
richardpark-msft: Just for future maintenance it'd be nice to include a few examples of what it's intended to match (just in a comment above)

packages/autorest.go/src/gocodemodel/gocodemodel.ts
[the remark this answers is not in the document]
jhendrixMSFT: Not that it blows up, but whether or not it's expected/allowed to be empty.  The majority of arrays are always expected to contain at least one item, but we don't actually verify this at present.

packages/autorest.go/src/generator/constants.ts
```diff
@@ -3,55 +3,53 @@
...
 
-import { Session } from '@autorest/extension-base';
 import { comment } from '@azure-tools/codegen';
-import { CodeModel, ChoiceValue, Schemas } from '@autorest/codemodel';
-import { length, values } from '@azure-tools/linq';
-import { contentPreamble, hasDescription, sortAscending } from './helpers';
-import { commentLength } from '../common/helpers';
+import { values } from '@azure-tools/linq';
+import { contentPreamble } from './helpers';
+import { commentLength } from './helpers';
```
RickWinter, line 9: Nit:
```suggestion
import { contentPreamble, commentLength } from './helpers';
```

packages/autorest.go/src/generator/models.ts
```diff
@@ -3,28 +3,28 @@
...
-import { Session } from '@autorest/extension-base';
+import { GoCodeModel, ModelType, MapType, ModelField, ModelFormat, PolymorphicType, InterfaceType, PossibleType } from '../gocodemodel/gocodemodel';
+import { getLiteralValueTypeName, getTypeDeclaration, isMapType, isBytesType, isInterfaceType, isModelType, isPrimitiveType, isPolymorphicType, isSliceType, isTimeType, isLiteralValue } from '../gocodemodel/gocodemodel';
 import { capitalize, comment } from '@azure-tools/codegen';
-import { ByteArraySchema, CodeModel, ConstantSchema, DictionarySchema, Language, ObjectSchema, Property, Schema, SchemaType } from '@autorest/codemodel';
 import { values } from '@azure-tools/linq';
-import { formatConstantValue, isArraySchema, isDictionarySchema, isObjectSchema, commentLength, aggregateProperties } from '../common/helpers';
-import { contentPreamble, hasDescription, getClientDefaultValue, sortAscending } from './helpers';
+import { commentLength } from './helpers';
+import { contentPreamble, formatLiteralValue, sortAscending } from './helpers';
```
RickWinter, line 11:
```suggestion
import { commentLength, contentPreamble, formatLiteralValue, sortAscending } from './helpers';
```

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT: fixed

packages/autorest.go/src/generator/constants.ts
jhendrixMSFT: fixed

packages/autorest.go/src/generator/models.ts
jhendrixMSFT: fixed

packages/autorest.go/src/generator/gomod.ts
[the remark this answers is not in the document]
jhendrixMSFT: added a comment

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT: This dovetails with the whole "null array vs empty array" conversation (that was my thinking anyways).  They might be able to be removed.  What's the idiom in TS when one should/shouldn't write a constructor for a type?

jhendrixMSFT:
@chlowell @richardpark-msft I will handle the nullable arrays and ctors in a separate PR so we can focus on those changes separately.  Let me know if you have any concerns with that.

packages/autorest.go/package.json
```diff
@@ -20,7 +20,7 @@
     "stop-testserver": "./node_modules/.bin/autorest-testserver stop",
     "watch": "tsc -p . --watch",
     "prepare": "npm run build",
-    "test": "jest --ci",
+    "test": "jest --ci --passWithNoTests",
```
RickWinter, line 23: Why do we want to pass if no tests run?

jhendrixMSFT: Because we don't, at present, have any TS unit tests.

packages/autorest.go/src/generator/constants.ts
```diff
@@ -3,55 +3,52 @@
...
-  if (session.model.language.go!.host) {
-    text += `const host = "${session.model.language.go!.host}"\n\n`;
+  let text = contentPreamble(codeModel);
+  if (codeModel.host) {
+    text += `const host = "${codeModel.host}"\n\n`;
   }
   // data-plane clients must manage their own constants for these values
-  if (<boolean>session.model.language.go!.azureARM) {
-    if (session.model.language.go!.moduleVersion === '') {
+  if (codeModel.type === 'azure-arm') {
```
RickWinter, line 22: We do this check is a few places, Should we make this a helper function?  Or at a minimum make `azure-arm` a constant.

jhendrixMSFT: The value is a discriminated union, see [here](https://github.com/Azure/autorest.go/blob/6b12a984bfccc7093f342d7757889ec12d09a75c/packages/autorest.go/src/gocodemodel/gocodemodel.ts#L40).

packages/autorest.go/src/generator/polymorphics.ts
```diff
@@ -3,137 +3,135 @@
...
-      const discriminatorInterface = recursiveUnwrapArrayDictionary(prop.schema).language.go!.discriminatorInterface;
-      if (discriminatorInterface) {
-        scalars.add(discriminatorInterface);
-        maps.add(discriminatorInterface);
+    } else if (isMapType(type)) {
+      const leafType = recursiveUnwrapMapSlice(type);
+      if (isInterfaceType(leafType)) {
+        scalars.add(leafType.name);
+        maps.add(leafType.name);
       }
```
RickWinter, line 49: What if its not one of the three types?  Is that ok?

This could be a case statement that throws if we the value is something other than an interface, slice or map.  Also are those mutually exclusive properties?

RickWinter (approved):

packages/autorest.go/src/generator/constants.ts
RickWinter: Makes sense, Why not wrap that into a function to avoid having `azure-arm` written throughout code?
if(codeModel.isAzureARM())

jhendrixMSFT: Here, `codeModel` is an interface, not a class, so it can't have any methods.  There might be some magic to make this work.  Will revisit in a separate PR.

PR #1003, opened 2023-08-01
Switch to using SanitizePagerPollerPath

jhendrixMSFT:
The formal way to sanitize these URLs instead of relying on an implementation detail.

### DIFF
--- a/.scripts/regeneration.js
+++ b/.scripts/regeneration.js
@@ -99,13 +99,13 @@ for (namespace in goMappings) {
 }
 
 const blobStorage = './swagger/specification/storage/data-plane/Microsoft.BlobStorage/readme.md';
-generateFromReadme("azblob", blobStorage, 'package-2021-12', 'test/storage/azblob', '--module-version=0.1.0 --inject-spans --azcore-version=1.7.0-beta.2');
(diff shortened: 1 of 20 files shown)


.scripts/regeneration.js
```diff
@@ -99,13 +99,13 @@ for (namespace in goMappings) {
 }
 
 const blobStorage = './swagger/specification/storage/data-plane/Microsoft.BlobStorage/readme.md';
-generateFromReadme("azblob", blobStorage, 'package-2021-12', 'test/storage/azblob', '--module-version=0.1.0 --inject-spans --azcore-version=1.7.0-beta.2');
+generateFromReadme("azblob", blobStorage, 'package-2021-12', 'test/storage/azblob', '--module-version=0.1.0 --inject-spans --azcore-version=1.8.0-beta.1.0.20230719175911-2cd542cfdd5c');
```
jhendrixMSFT, line 102: These will need to be updated with the actual release once it's happened.

jhendrixMSFT: @chlowell ready now.

chlowell (approved):

PR #1016, opened 2023-09-05
Fix generating duplicate interfaces

Alancere:
Remove duplicates from allDiscriminators


tadelesh (approved):

packages/autorest.go/src/generator/interfaces.ts
```diff
@@ -33,6 +33,9 @@ export async function generateInterfaces(session: Session<CodeModel>): Promise<s
     }
   }
 
+  // remove duplicate
+  allDiscriminators = Array.from(new Set(allDiscriminators));
```
jhendrixMSFT, line 37: Maybe we should just update the condition on line 30 to skip adding an entry if it already exists.

Also, do you have an example of where this happens?

Alancere: Encountered when generating services: 
- sdk/resourcemanager/security/armsecurity
- sdk/resourcemanager/mediaservices/armmediaservices

jhendrixMSFT (approved):

PR #1017, opened 2023-09-05
Fix various codegen bugs

jhendrixMSFT:
Ensure unique var names when splitting arrays.
Fix fake type name generation for slices when slice-elements-byval is enabled.
Fix time serde codegen when slice-elements-byval is enabled. Include method parameter types when generating fake polymorphic helpers. Fix fake codegen for types that coalesce to string.


jhendrixMSFT:
Thanks @tadelesh for the bug report.

chlowell (approved):
Is it feasible to add test coverage for these bugs?

PR #1023, opened 2023-09-11
Fix fake generation for slice of primitive type params

jhendrixMSFT:
The inner condition was too restrictive compared to the outer condition, leading to an error being thrown.
Fixed regression generating polymorphic helpers that was introduced when feature/gocodemodel merged to main.

The regression was part of the change made in https://github.com/Azure/autorest.go/pull/1017


jhendrixMSFT:
I need to look why the test added for the polymorphic params didn't catch this.

jhendrixMSFT:
We were over-generating polymorphic helpers for fakes.  This has been corrected.

chlowell (approved):

Issue #1027, opened 2023-09-15
Time array type byval serialization problem


tadelesh:
Introduced in this PR: https://github.com/Azure/autorest.go/pull/986.
In ser/deser mehtod, `time.Time` type could not convert to `timeRFC3339` type directly (pointer in previous logic is ok).
![image](https://github.com/Azure/autorest.go/assets/1726438/de9bd5fd-fdbb-4df5-b4b8-e3aa58ca890a)

cc: @Alancere

jhendrixMSFT:
Which code generation version are you using?  I noticed this recently and fixed it in https://github.com/Azure/autorest.go/pull/1017

tadelesh:
It seems without `slice-elements-byval` flag, the marshal file is still trying to deal with val not pointer.

Issue #961, opened 2023-06-07
Fakes path regex is broken for URL-encoded values


jhendrixMSFT:
https://github.com/Azure/autorest.go/blob/c0de3acc643df1f3180d5d0723c7514692f946eb/packages/autorest.go/src/fake/servers.ts#L475

tadelesh:
@jhendrixMSFT I don't know this problem has been resolved thoroughly. We happened into runtime error dealing with the following url: https://github.com/Azure/azure-rest-api-specs/blob/80c21c17b4a7aa57f637ee594f7cfd653255a7e0/specification/iothub/resource-manager/Microsoft.Devices/preview/2023-06-30-preview/iothub.json#L1084

PR #1029, opened 2023-09-18
Fakes and slice-elements-byval fixes

jhendrixMSFT:
Escapse path chars that could be interpreted as regex tokens. Fix codegen for header params that are arrays (it used to throw). Don't use package name prefix for slices of non-constant types. Fixed regression in slice-elements-byval from GoCodeModel merge.

Addendum to https://github.com/Azure/autorest.go/issues/961
Fixes https://github.com/Azure/autorest.go/issues/1027

### DIFF
--- a/packages/autorest.go/src/generator/fake/servers.ts
+++ b/packages/autorest.go/src/generator/fake/servers.ts
@@ -467,7 +467,7 @@ function createPathParamsRegex(method: Method): string {
   // note that some path params are optional.
   let urlPath = method.httpPath;
   // escape any characters in the path that could be interpreted as regex tokens
-  // per the RFC, these are the pchars that also double as regex tokens
+  // per RFC3986, these are the pchars that also double as regex tokens
   // . $ * + ()
   urlPath = urlPath.replace(/([.$*+()])/g, '\\$1');
   for (const param of values(method.parameters)) {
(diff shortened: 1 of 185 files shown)


packages/autorest.go/src/generator/fake/servers.ts
```diff
@@ -466,6 +466,10 @@ function createPathParamsRegex(method: Method): string {
   // each path param will replaced with a regex capture.
   // note that some path params are optional.
   let urlPath = method.httpPath;
+  // escape any characters in the path that could be interpreted as regex tokens
+  // per the RFC, these are the pchars that also double as regex tokens
```
chlowell, line 470: **p**chars?

chlowell (approved):

packages/autorest.go/src/generator/fake/servers.ts
jhendrixMSFT: I was looking at [RFC3986](https://www.rfc-editor.org/rfc/rfc3986), I thought it meant "path character".

chlowell: 👍 thanks, that's the grammar I was missing. I'd probably put the RFC number here for anyone who wants to go down that rabbit hole

jhendrixMSFT: Good call, comment updated.

PR #1033, opened 2023-09-19
Fix multi-collection format in fakes

jhendrixMSFT:
Multi-collection format was being treated as a CSV which is incorrect. Add first-class support for parameter collection formats instead of relying on type inspection and optional field heuristics.

### DIFF
--- a/packages/autorest.go/src/gocodemodel/gocodemodel.ts
+++ b/packages/autorest.go/src/gocodemodel/gocodemodel.ts
@@ -379,21 +379,35 @@ export interface ParameterGroup {
   params: Array<Parameter>;
 }
 
-export type HeaderType = BytesType | ConstantType | MapType | PrimitiveType | SliceType | TimeType | LiteralValue;
+export type HeaderType = BytesType | ConstantType | PrimitiveType | TimeType | LiteralValue;
 
 // parameter is sent via an HTTP header
 export interface HeaderParameter extends Parameter {
   headerName: string;
 
   type: HeaderType;
-
-  collectionPrefix?: string;
-
-  // for slice types, this is the delimiter to use instead of the default ','
-  delimiter?: '|' | ' ' | '\\t'
 }
 
-export type PathParameterType = BytesType | ConstantType | PrimitiveType | SliceType | TimeType | LiteralValue;
+export type CollectionFormat = 'csv' | 'ssv' | 'tsv' | 'pipes';
+
+export interface HeaderCollectionParameter extends Parameter {
+  headerName: string;
+
+  type: SliceType;
+
+  collectionFormat: CollectionFormat;
+}
+
+// this is a special type to support x-ms-header-collection-prefix (i.e. storage)
+export interface HeaderMapParameter extends Parameter {
+  headerName: string;
+
+  type: MapType;
+
+  collectionPrefix: string;
+}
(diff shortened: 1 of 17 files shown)


tadelesh:
One thing I wonder if you have take into account: if query param is empty, empty judge is needed before split.
image

jhendrixMSFT:
Opened https://github.com/Azure/autorest.go/pull/1035 to fix splitting on an empty value.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -742,6 +810,10 @@ export function isConstantType(type: PossibleType): type is ConstantType {
   return (<ConstantType>type).values !== undefined;
 }
 
+export function isHeaderMapResponse(resp: HeaderResponse | HeaderMapResponse): resp is HeaderMapResponse {
```
chlowell, line 813: I haven't seen this syntax before. I guess the function accepts a parameter of either type but returns undefined when the parameter is a HeaderResponse?

packages/autorest.go/src/gocodemodel/gocodemodel.ts
```diff
@@ -465,6 +498,10 @@ export function isFormBodyParameter(param: Parameter): param is FormBodyParamete
   return (<FormBodyParameter>param).formDataName !== undefined;
 }
 
+export function isFormBodyCollectionParameter(param: Parameter): param is FormBodyCollectionParameter {
+  return (<FormBodyCollectionParameter>param).formDataName !== undefined && (<FormBodyCollectionParameter>param).collectionFormat !== undefined;
```
chlowell, line 502: TypeScript question: can we define the interface such that callers can cast instead of calling this helper?

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT: [This](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) will explain it better than I can.

packages/autorest.go/src/gocodemodel/gocodemodel.ts
jhendrixMSFT: We could.  I imagine it would look something like how M4 types have a [`type`](https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/src/transform/transform.ts#L264) field and we switch/case on them.  Do you think that would be easier to use?  Personally, I've really been happy with using type guards as the compiler automatically converts `param` to `FormBodyCollectionParameter` (or whatever type predicate is defined in the type guard) for me if the shape of the data matches.

chlowell: No, I don't suppose that would be easier. This pattern just strikes me as odd because I expect the type system to have a simpler built-in way to tell whether an object satisfies some interface, and this function makes the criteria the object's data, not its shape. But this perception is probably due to my unfamiliarity with TypeScript and its idioms.

chlowell (approved):

tadelesh:
After this fix, we still met the following problem with armcdn:
21d2541b-bf4a-4067-9990-283084fc0b2e

PR #1037, opened 2023-09-21
Add skeleton project for typespec-go

jhendrixMSFT:
The plugin has the tsp emitter entry point and that's it.
Set up basic CI pipeline to compile and lint typespec-go.
Removed exclulde on swagger directory for autorest.go as it uses it.


jhendrixMSFT:
Hmm, perhaps this should target `main` instead.  Then I can rebase the `feature/typespec-go` branch on this change and set up the pipeline once.

RickWinter (approved):

PR #1045, opened 2023-10-10
Add missing helpers in a few cases

jhendrixMSFT:
Whenever we use a parse helper, set the flag to generate them. Bump package version in prep for release.

### DIFF
--- a/packages/autorest.go/package.json
+++ b/packages/autorest.go/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@autorest/go",
-  "version": "4.0.0-preview.56",
+  "version": "4.0.0-preview.57",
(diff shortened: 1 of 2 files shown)


chlowell (approved):
Should the test suite include one of the cases formerly missing helpers?

jhendrixMSFT:
We can, but it would require adding two new E2E swaggers to generate which seems a bit expensive (we didn't catch it in `azalias` due to too much overlap within the same module).  Also, for this particular case, I felt it was ok to cheat a bit as the parse helpers and setting the flag must always show up in pairs; i.e. if you emit `parseWithCast` then there must also be a `requiredHelpers.parseWithCast = true` right next to it.

PR #1067, opened 2023-11-07
unmarshal: disallow unknown fields

serbrech:
attempting to resolve https://github.com/Azure/azure-sdk-for-go/issues/21892

- [x] disallow unknown fields
- [x] add a generator option to enable this behavior
- [x] add tests

### DIFF
--- a/packages/autorest.go/.scripts/regeneration.js
+++ b/packages/autorest.go/.scripts/regeneration.js
@@ -10,7 +10,7 @@ const sem = require('./semaphore')(8);
 const swaggerDir = 'packages/autorest.go/node_modules/@microsoft.azure/autorest.testserver/swagger/';
 
 const goMappings = {
-    'additionalpropsgroup': ['additionalProperties.json', '--remove-unreferenced-types'],
+    'additionalpropsgroup': ['additionalProperties.json', '--remove-unreferenced-types', '--disallow-unknown-fields'],
     'arraygroup': ['body-array.json', '--remove-unreferenced-types'],
     'azurereportgroup': ['azure-report.json', '--remove-unreferenced-types'],
     'azurespecialsgroup': ['azure-special-properties.json', '--head-as-boolean', '--remove-unreferenced-types'],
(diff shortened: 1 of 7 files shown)


packages/autorest.go/.scripts/regeneration.js
```diff
@@ -128,6 +128,9 @@ generateFromReadme("armdataboxedge", databoxedge, 'package-2021-02-01', 'test/da
 const acr = './swagger/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json';
 generate("azacr", acr, 'test/acr/azacr', '--module="azacr" --module-version=0.1.0 --openapi-type="data-plane" --rawjson-as-bytes --generate-fakes --azcore-version=1.9.0-beta.1');
 
+const acrDisallowUnknownFields = './swagger/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json';
+generate("azacrdisallowunknown", acrDisallowUnknownFields, 'test/acr/azacr-disallow-unknown', '--module="azacr" --module-version=0.1.0 --openapi-type="data-plane" --rawjson-as-bytes --generate-fakes --disallow-unknown-fields --azcore-version=1.9.0-beta.1');
```
serbrech, line 132: tried to use this to generate models with --disallow-unknown-fields, but could not get this script to work

packages/autorest.go/.scripts/regeneration.js
```diff
@@ -128,6 +128,9 @@ generateFromReadme("armdataboxedge", databoxedge, 'package-2021-02-01', 'test/da
 const acr = './swagger/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json';
 generate("azacr", acr, 'test/acr/azacr', '--module="azacr" --module-version=0.1.0 --openapi-type="data-plane" --rawjson-as-bytes --generate-fakes --azcore-version=1.9.0-beta.1');
 
+const acrDisallowUnknownFields = './swagger/specification/containerregistry/data-plane/Azure.ContainerRegistry/stable/2021-07-01/containerregistry.json';
```
jhendrixMSFT, line 131: I think this can be removed and just use `acr` from above?

packages/autorest.go/.scripts/regeneration.js
jhendrixMSFT: Few things.
`--module="azacrdisallowunknown"`
The `--azcore-version` can be removed as `v1.9.0` is now the default version.
If you `rush regenerate` it should work. Or did you try that and it failed?

jhendrixMSFT:
I took the liberty of rebasing on `main` and made a few small changes.

packages/autorest.go/test/autorest/additionalpropsgroup/zz_models_serde.go
```diff
@@ -100,6 +100,8 @@ func (p *PetAPInProperties) UnmarshalJSON(data []byte) error {
 		case "status":
 			err = unpopulate(val, "Status", &p.Status)
 			delete(rawMsg, key)
+		default:
```
jhendrixMSFT, line 103: I was surprised by this. I think we have a codegen bug (unrelated to this change).

jhendrixMSFT: Opened https://github.com/Azure/autorest.go/issues/1069 to track this.

jhendrixMSFT (approved):

jhendrixMSFT:
@serbrech if you're good with the changes I made I'll merge this and release an updated code generator today.

serbrech:
give us a little bit of time to think of it, and how exactly the service will handle it before merging/releasing

jhendrixMSFT:
Sounds good. I have another pending change that I need to release today, so we can release this change at a later time when ready.

packages/autorest.go/src/generator/models.ts
```diff
@@ -379,6 +380,9 @@ function generateJSONUnmarshallerBody(modelType: ModelType | PolymorphicType, re
   if (addlProps) {
     unmarshalBody += '\t\tdefault:\n';
     unmarshalBody += emitAddlProps('\t', addlProps);
+  } else if (options.disallowUnknownFields) {
+    unmarshalBody += '\t\tdefault:\n';
+    unmarshalBody += `\t\t\terr = fmt.Errorf("unmarshalling type %T, unknown field %q", ${receiver}, key)\n`;
```
jim-minter, line 385: I'm +1 on this PR as is, but wanted to ask a question about this line to check we've thought it through enough.

At the moment our service doesn't return unmarshalling errors to the user verbatim, we just return a fairly general error.

If we enabled this toggle, to make full use of it, I think it'd imply starting passing through the error string directly from json.Unmarshal to the end user.  Is that what we want?
 
Leaking %T isn't wonderful, but being able to construct an error message saying "foo.properties[123].bar.badField is not allowed" would be a pain, so it seems pragmatic to me.

We wouldn't want a defined error type like UnrecognisedFieldError{ TypeName string, FieldName string } to be able to pick out this error?

jim-minter (approved):

packages/autorest.go/src/generator/models.ts
jim-minter: `"unmarshalling type %T, unknown field %q"` is a copy from the Go standard library JSON implementation.  This works for me, let's not over complicate.

serbrech: I thought about this, but opted to keep it simple and match what the json package does.
even the error string is the same in fact.

serbrech:
@jhendrixMSFT we're good with this as it is. you can rebase/merge whenever it works for you (no rush either :) )

PR #1072, opened 2023-11-13
Support XML additional properties with empty values

jhendrixMSFT:
Use the empty string (not nil) if the key has no value.

Codegen fix for https://github.com/Azure/azure-sdk-for-go/issues/21887


richardpark-msft (approved):
Looks good for Storage. Any concerns about this "activating" for other libraries or are XML responses rare?

jhendrixMSFT:
This will apply to all SDKs that use XML.

PR #1077, opened 2023-11-27
Move codegen and code model out of autorest.go

jhendrixMSFT:
This will allow it to be shared with the TypeSpec emitter. The code will be bundled with the respective packages.
Migrated from CommonJS to EcmaScript modules as the TypeSpec emitter is EcmaScript based.
Autorest.gotest has removed its dependency on autorest.go, copying the handful of shared code.


chlowell (approved):

packages/codegen.go/clientFactory.ts
```diff
@@ -3,10 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as go from '../gocodemodel/gocodemodel';
+import * as go from '../codemodel.go/gocodemodel.js';
```
tadelesh, line 6: Is it possible to import with `@azure-tools/codemodel.go`? Maybe we need to release a first version for it.

jhendrixMSFT: I explicitly chose this design as I want to avoid having another dependency. Releasing `codemodel.go` as a node package would mean that updating the code generator (autorest and/or typespec) means updating/releasing `codemodel.go` then the respective code generators.

PR #1082, opened 2023-11-29
Improvements to the ARM client factory

jhendrixMSFT:
Share the *arm.Client across SDK clients instead of always creating a new one (and not throwing away the "test" one).


jhendrixMSFT:
@tadelesh I'll wait for your ack before merging. Also will clean up CODEOWNERS.

tadelesh (approved):

PR #1085, opened 2023-11-30
Get basic codegen for TypeSpec working

jhendrixMSFT:
This should cover basic CRUD operations and supporting types. It does not include support for LROs, pagination, and polymorphism. The layout and flow is similar to the autorest version; the input is adapted to the Go code model, fed to the code generator, and the results written to disk.
There are chunks of commented out code and TODO comments where shortcuts were taken (e.g. hard-coded values) to get things standing up. Added "rush tspcompile" which is analogous to "rush regenerate" to rebuild the world. ~The only difference is that the script will also mod tidy after formatting.~
The short names for rush script parameters have been removed as they kept colliding with in-box machinery.
The TS CI has been updated to build/lint/run/publish results. Centralized initialization of field arrays in the code model. Added code model support for raw JSON as bytes and slice elements by val configuration options.
Removed hard-coded pointer to time types in operation request/response handlers; it now honors the user-specified setting.


packages/typespec-go/tsconfig.json
```diff
@@ -16,10 +16,9 @@
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
```
jhendrixMSFT, line 19: Temporary, to ease development.

packages/typespec-go/src/tcgcadapter/mappings.ts
jhendrixMSFT, line 1: I believe that this, and `namer.ts`, can be shared across the emitters (will look into it).

jhendrixMSFT: Fixed.

jhendrixMSFT:
One thing I've noticed with the current relative module approach is that making a change to say `codegen.go` followed by a `rush build` will _not_ build `autorest.go` and `typespec-go` which is annoying (`rush rebuild` will). I _think_ we can solve this via [project references](https://www.typescriptlang.org/docs/handbook/project-references.html), will investigate further. No need to hold up this PR for that though.

chlowell (approved):

packages/typespec-go/test/cadlranch/arraygroup/arraygroup_test.go
```diff
@@ -0,0 +1,188 @@
+//go:build go1.18
```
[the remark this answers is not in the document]
jhendrixMSFT: Added all the missing license headers.

PR #1089, opened 2023-12-04
Add more cadl ranch coverage

jhendrixMSFT:
Fixed an issue that caused an unmarshaller to be generated when the
type contains no fields, leading to a compilation error.
Renamed a few test modules to clarify their cadl ranch source.


packages/codegen.go/src/models.ts
```diff
@@ -618,13 +618,13 @@ class ModelDef {
   readonly Name: string;
   readonly Format: go.ModelFormat;
   readonly Description?: string;
-  readonly Fields?: Array<go.ModelField>;
+  readonly Fields: Array<go.ModelField>;
```
jhendrixMSFT, line 621: This was a missed case when we refactored the Go code model to remove optional arrays.

chlowell (approved):

PR #1093, opened 2023-12-05
Make discriminatorValue a LiteralValue

jhendrixMSFT:
Previously, the code model was leaking an implementation detail that was specific to how M4 models the discriminator value. That's now been moved to the M4 adapter.

### DIFF
--- a/packages/autorest.go/src/m4togocodemodel/types.ts
+++ b/packages/autorest.go/src/m4togocodemodel/types.ts
@@ -109,7 +109,7 @@ export function adaptModel(obj: m4.ObjectSchema): go.ModelType | go.PolymorphicT
     modelType = new go.PolymorphicType(obj.language.go!.name, <go.InterfaceType>iface, annotations);
     // only non-root and sub-root discriminators will have a discriminatorValue
     if (obj.discriminatorValue) {
-      (<go.PolymorphicType>modelType).discriminatorValue = obj.discriminatorValue;
+      (<go.PolymorphicType>modelType).discriminatorValue = getDiscriminatorLiteral(obj.discriminatorValue);
     }
   } else {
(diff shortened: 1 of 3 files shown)


packages/autorest.go/src/m4togocodemodel/types.ts
```diff
@@ -544,20 +537,35 @@ function adaptLiteralValue(constSchema: m4.ConstantSchema): go.LiteralValue {
...
+  return literalConst;
+}
+
+function adaptStringLiteral(literal: string): go.LiteralValue {
+  const keyName = `literal-string-${literal}`;
+  let literalString = types.get(keyName);
+  if (literalString) {
+    return <go.LiteralValue>literalString;
+  }
+  literalString = new go.LiteralValue(new go.PrimitiveType('string'), literal);
```
chlowell, line 564: Is it this need for a `PrimitiveType` that prevents writing a function that handles both string and const literals?

jhendrixMSFT: That and the type cache key names are slightly different.

chlowell: I'm just looking at this thinking, if there were a type for string literals having `name` "string" and some way to produce a `PrimtiveType`, then `adaptStringLiteral` could collapse into `adaptConstLiteral`. Would that be painfully complex or semantically weird?

jhendrixMSFT: I've refactored this so it's consolidated within `getDiscriminatorLiteral`. I reverted the other part of the changes which is adapting M4 to Go code model. Part of what was making things complicated earlier was trying to mix "adapting code" with "adapted code". We should keep those kinds of things separate.

chlowell (approved):

PR #1095, opened 2023-12-11
Add support for discriminated type

jhendrixMSFT:
Moving naming of the discriminator interface to the naming package. Fixed another case of "no undefined arrays", make it empty. Added source map support to typespec-go for better stacks on error. Substitution of the interface type for discriminators has been merged into getPossibleType.
Extended getLiteralValue to handle enum value literals and other types. Removed validation of overlapping fields with different types as the TypeSpec compiler does this for us.


packages/typespec-go/.scripts/tspcompile.js
```diff
@@ -21,11 +21,11 @@ const cadlRanch = {
   'extensiblegroup': ['type/enum/extensible'],
   'fixedgroup': ['type/enum/fixed'],
   'emptygroup': ['type/model/empty'],
-  //'enumdiscriminatorgroup': ['type/model/inheritance/enum-discriminator'],
-  //'nesteddiscriminatorgroup': ['type/model/inheritance/nested-discriminator'],
-  //'nodiscriminatedgroup': ['type/model/inheritance/not-discriminated'],
-  //'recursivegroup': ['type/model/inheritance/recursive'],
-  //'singlediscriminatorgroup': ['type/model/inheritance/single-discriminator'],
+  'enumdiscgroup': ['type/model/inheritance/enum-discriminator'],
+  //'nesteddiscgroup': ['type/model/inheritance/nested-discriminator'],
```
jhendrixMSFT, line 25: This scenario is multiple discriminator properties on a type. The TS team is debating on the validity/usefulness of this and might remove it. So we're skipping this for the time being.

chlowell (approved):

PR #1096, opened 2023-12-11
Move auto-generated field comments to codegen.go

jhendrixMSFT:
This required adding usage flags to the Go code model to describe if a model is input/output/both. This also fixed a corner-case where a field was incorrectly flagged as REQUIRED when it's really CONSTANT. Replaced __filename with the ES module equivalent.

### DIFF
--- a/packages/codegen.go/src/helpers.ts
+++ b/packages/codegen.go/src/helpers.ts
@@ -309,10 +309,15 @@ export function formatValue(paramName: string, type: go.PossibleType, imports: I
 // returns the clientDefaultValue of the specified param.
 // this is usually the value in quotes (i.e. a string) however
 // it could also be a constant.
-export function formatLiteralValue(value: go.LiteralValue): string {
+// specify withCast: false for doc comments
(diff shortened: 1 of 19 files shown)


packages/codegen.go/src/helpers.ts
```diff
@@ -309,10 +309,15 @@ export function formatValue(paramName: string, type: go.PossibleType, imports: I
 // returns the clientDefaultValue of the specified param.
 // this is usually the value in quotes (i.e. a string) however
 // it could also be a constant.
-export function formatLiteralValue(value: go.LiteralValue): string {
+// specify withCast: false for doc comments
```
chlowell, line 312: If `withCast` is relevant only to doc comments, maybe it should be e.g. `docComment` instead

chlowell (approved):

packages/codegen.go/src/helpers.ts
jhendrixMSFT: It's true at present, but I suppose it might not be in the future. Maybe I should just remove the comment.

PR #1102, opened 2023-12-15
Add coverage for tsp durations

jhendrixMSFT:
tsp provisions for different encodings besides ISO8601 including numeric values. This required splitting getPossibleTypes into two functions as the built-in types are a subset of SdkType.
For header/query collection parameters, set the element type by value to true as that's what we do in the autorest version. Add ISO\d+ to the list of acronyms to upper-case.


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -212,25 +167,75 @@ export class typeAdapter {
         this.types.set(keyName, mapType);
         return mapType;
       }
-      case 'int32': {
-        const int32Key = 'int32';
-        let int32 = this.types.get(int32Key);
-        if (int32) {
-          return int32;
+      case 'duration': {
```
jhendrixMSFT, line 170: New case for duration.

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -212,25 +167,75 @@ export class typeAdapter {
...
-        int64 = new go.PrimitiveType(int64Key);
-        this.types.set(int64Key, int64);
-        return int64;
+        return this.getModel(type);
+      default:
+        throw new Error(`unhandled property kind ${type.kind}`);
+    }
+  }
+
+  private getBuiltInType(type: tcgc.SdkBuiltInType): go.PossibleType {
```
jhendrixMSFT, line 192: The contents of this function were moved from `getPossibleType`.

packages/typespec-go/test/cadlranch/encode/durationgroup/custom_client.go
```diff
@@ -0,0 +1,46 @@
+//go:build go1.18
```
chlowell, line 1: Do we need these in this repo?

chlowell (approved):

PR #1104, opened 2024-01-03
Rename StandardType to QualifiedType

jhendrixMSFT:
Expand usage beyond the standard library, e.g. ETag from azcore. Remove the need to duplicate the package name in the type name. It will be extracted in calls to getTypeDeclaration().

### DIFF
--- a/packages/codemodel.go/src/gocodemodel.ts
+++ b/packages/codemodel.go/src/gocodemodel.ts
@@ -155,7 +155,7 @@ export interface PolymorphicType extends StructType {
 }
 
 // PossibleType describes what can be modeled e.g. in an OpenAPI specification
-export type PossibleType = BytesType | ConstantType | InterfaceType | LiteralValue | MapType | ModelType | PolymorphicType | PrimitiveType | SliceType | StandardType | TimeType;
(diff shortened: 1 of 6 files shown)


packages/codemodel.go/src/gocodemodel.ts
```diff
@@ -769,9 +769,9 @@ export interface LiteralValue {
   literal: any;
 }
 
-// StandardType is a type from the Go standard library (excluding time.Time)
-export interface StandardType {
-  // this is the fully-qualified name (e.g. io.Reader)
+// PackageType is a type from some package, e.g. the Go standard library (excluding time.Time)
+export interface PackageType {
```
chlowell, line 773: Would QualifiedType be more to the point i.e., the crucial detail is that we must qualify references to this type?

chlowell (approved):

packages/codemodel.go/src/gocodemodel.ts
jhendrixMSFT: I like it and it aligns with the terminology used in the [language spec](https://go.dev/ref/spec#Qualified_identifiers).

PR #1105, opened 2024-01-03
Consolidate custom command scripts

jhendrixMSFT:
The modtidy and buildvet scripts are now project agnostic. Consolidated copies of semaphore.js, reverting it to its CommonJS format from commit history.


.scripts/buildvet.js
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
-import { execSync } from 'child_process';
-import * as fs from 'fs';
+const execSync = require('child_process').execSync;
+const fs = require('fs');
+const path = require('path');
 
-recursiveFindGoMod('./packages/autorest.go/test');
+recursiveFindGoMod(process.env.RUSH_INVOKED_FOLDER);
```
jhendrixMSFT, line 7: We can't use `process.cwd()` as rush always sets it to the location of the `rush.json` file which is usually in the root of the repo.

chlowell (approved):

PR #1113, opened 2024-01-22
Add custom test templates for gotests

jhendrixMSFT:
Used to generate test stubs.
NOTE: requires HEAD of https://github.com/cweill/gotests at present.

### DIFF
--- a/common/templates/header.tmpl
+++ b/common/templates/header.tmpl
@@ -9,8 +9,5 @@ package {{.Package}}_test
 
 import (
 	"testing"
-	"{{.Package}}"
-
-	"github.com/stretchr/testify/require"
 )
 {{end}}
--- a/common/templates/header.tmpl
+++ b/common/templates/header.tmpl
@@ -1,0 +1,16 @@
+{{define "header"}}
+//go:build go1.18
+// +build go1.18
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
To use these templates, edit settings for `Go: Generate Tests Flags` and add the following lines.

`-exported`
`-template_dir`
`<repo_root>/common/templates`

Note that there are a few rough edges at present.
- The generated file names will have the same `zz_` prefix which will need to be removed.
- Since this generates test stubs for all exported methods, you will see a `zz_models_serde_test.go` which needs to be deleted.

common/templates/header.tmpl
```diff
@@ -0,0 +1,16 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+package {{.Package}}_test
+
+import (
+	"testing"
+	"{{.Package}}"
+
```
chlowell, line 13: This is super nitpicky, but should the package under test be grouped with standard library imports?

```suggestion
	"testing"

	"{{.Package}}"
```

chlowell (approved):

common/templates/header.tmpl
jhendrixMSFT: I meant to remove these. I had them in place as I was trying to emit calling client constructors but couldn't get it to work.

Issue #1125, opened 2024-01-30
Omit `Client` prefix on generated types


jhendrixMSFT:
For packages with one client, prefixing the options and response types with `Client` is redundant and makes the code ugly. Couple of options.

- for packages with a single client, maybe we can disable this by default. this might be problematic though if more clients are added
- make the above an opt-in switch, and if multiple clients are in the package produce an error

PR #1126, opened 2024-01-30
Add single-client option for code generation

jhendrixMSFT:
When enabled, the client name prefix will be omitted from options and response types. If there are multiple clients, an error is returned. Added missing doc comments for options types.
Fixed doc comments for methods that return pagers.

Fixes https://github.com/Azure/autorest.go/issues/1125


richardpark-msft (approved):
Thanks for doing this so fast. Literally just talked THIS morning!

PR #1127, opened 2024-01-30
Add pipeline for publishing cadl-ranch test coverage

jhendrixMSFT:
Replaced some relative paths with fully qualified ones. Updated to latest version of Node.

### DIFF
--- a/eng/pipelines/cadl-ranch-typespec.yml
+++ b/eng/pipelines/cadl-ranch-typespec.yml
@@ -9,9 +9,6 @@ pr: none
 
 steps:
   - template: ../steps/set-env.yaml
-    parameters:
-      NodeVersion: "18.x"
-      GoVersion: "$(go.version)"
 
   - template: ../steps/build-test-typespec.yaml
 
--- a/eng/pipelines/cadl-ranch-typespec.yml
+++ b/eng/pipelines/cadl-ranch-typespec.yml
@@ -1,0 +1,25 @@
+trigger:
+  branches:
+    include:
+      - main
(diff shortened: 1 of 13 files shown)


eng/pipelines/cadl-ranch-typespec.yml
```diff
@@ -0,0 +1,25 @@
...
+      - main
+  paths:
+    include:
+      - packages/typespec-go/test/cadlranch
+pr: none
+
+steps:
+  - template: ../steps/set-env.yaml
+    parameters:
+      NodeVersion: "18.x"
```
jhendrixMSFT, line 13: @benbp do we need to set these params since they're already defined in `set-env.yaml`?

jhendrixMSFT:
@benbp could you please help set up the internal pipeline for this? It's based on the [Python version](https://github.com/Azure/autorest.python/blob/getAllOperations/eng/pipelines/internal-ci.yml). The gist is that whenever we update the cadl-ranch tests in `main` we want this to trigger. It will also need some secrets wired in.

RickWinter (approved):

eng/pipelines/cadl-ranch-typespec.yml
benbp: If you're not trying to override the default then no.

benbp:
@jhendrixMSFT https://dev.azure.com/azure-sdk/internal/_build/index?definitionId=6688&_a=completed

jhendrixMSFT:
Thanks @benbp !

PR #1129, opened 2024-01-31
Add lenient handling of RFC3339 as ISO8601

jhendrixMSFT:
While RFC3339 requires a 'T' or 't' character separating the date and time, ISO8601 allows a single space character.
Changed the time-zone offset regexp to be non-capturing.

Codegen change for https://github.com/Azure/azure-sdk-for-go/issues/22272

### DIFF
--- a/packages/codegen.go/src/time.ts
+++ b/packages/codegen.go/src/time.ts
@@ -156,16 +156,20 @@ function generateRFC3339Helper(preamble: string, dateTime: boolean, time: boolea
 ${imports.text()}
 
 // Azure reports time in UTC but it doesn't include the 'Z' time zone suffix in some cases.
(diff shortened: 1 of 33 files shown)


packages/codegen.go/src/time.ts
```diff
@@ -181,17 +185,33 @@ func (t dateTimeRFC3339) MarshalText() ([]byte, error) {
 }
 
 func (t *dateTimeRFC3339) UnmarshalJSON(data []byte) error {
-	layout := utcDateTimeJSON
-	if tzOffsetRegex.Match(data) {
+	tzOffset := tzOffsetRegex.Match(data)
```
RickWinter, line 188: How much of perf hit do we take b/c of this change?

jhendrixMSFT: No clue, but I doubt it's significant given we're already O(N^2) due to nested called to `json.Unmarshal` :(

RickWinter (approved):

PR #1135, opened 2024-02-06
Remove workspaces from dependencies

jhendrixMSFT:
Replaced with workspaces definitions.

### DIFF
--- a/eng/pipelines/publish-dev-release-go.yml
+++ b/eng/pipelines/publish-dev-release-go.yml
@@ -15,5 +15,6 @@ steps:
       cd packages/autorest.go
       npm version --no-git-tag-version $currentVersion
       npm pack;
+      npm install -g autorest-go-$currentVersion.tgz
       npx publish-release --token $(package-write-token) --repo autorest.go --owner azure --name "Autorest for Go v$currentVersion" --tag v$currentVersion --notes='Preview version of Autorest for Go track 2' --prerelease --editRelease false --assets autorest-go-$currentVersion.tgz --target_commitish $(Build.SourceBranchName);
     displayName: "Publish Dev Release"
--- a/eng/pipelines/publish-dev-release-go.yml
+++ b/eng/pipelines/publish-dev-release-go.yml
@@ -16,8 +16,8 @@ steps:
       npm version --no-git-tag-version $currentVersion
       npm pack;
       npm install -g autorest-go-$currentVersion.tgz
-      if (!$?) {
-        exit 1
+      if ($LASTEXITCODE) {
+        exit $LASTEXITCODE
       }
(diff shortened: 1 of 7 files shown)


packages/autorest.go/package.json
```diff
@@ -63,8 +68,6 @@
     "@autorest/extension-base": "~3.5.2",
     "@azure-tools/codegen": "~2.9.2",
     "@autorest/codemodel": "~4.19.3",
-    "@azure-tools/codegen.go": "workspace:*",
```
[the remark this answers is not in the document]
timotheeguerin: have you moved to pure `npm` worksapce? the `worksapce:` prefix is I thikn something only `pnpm` support(and rush if you configure to use pnpm workspace instead of the default system)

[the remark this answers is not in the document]
timotheeguerin: if you moved to `pnpm` workspace then you should use `pnpm pack` instead of `npm pack` 
if you moved to `npm` workspace then it should be good to change to the correct version I think.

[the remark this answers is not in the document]
jhendrixMSFT: Is the pure npm workspace what's described [here](https://docs.npmjs.com/cli/v10/using-npm/workspaces)?

eng/pipelines/publish-dev-release-go.yml
```diff
@@ -15,5 +15,9 @@ steps:
       cd packages/autorest.go
       npm version --no-git-tag-version $currentVersion
       npm pack;
+      npm install -g autorest-go-$currentVersion.tgz
+      if (!$?) {
```
jhendrixMSFT, line 19: @benpb does this ealy exit look correct to you?

chlowell (approved):

packages/autorest.go/package.json
```diff
@@ -43,6 +43,11 @@
   },
   "homepage": "https://github.com/Azure/autorest.go#readme",
   "readme": "https://github.com/Azure/autorest.go/blob/main/readme.md",
+  "workspaces": [
```
timotheeguerin, line 46: do you actually need this, this feels super weird to me. As this is meant to define this package as a workspace(naming is terrible) and those as project in the workspace

jhendrixMSFT: Ah I see. Removed.

eng/pipelines/publish-dev-release-go.yml
benbp: @jhendrixMSFT this works but I prefer:

```suggestion
      if ($LASTEXITCODE) {
        exit $LASTEXITCODE
      }
```

benbp: I also like to add this to the top of scripts sometimes:

```
$ErrorActionPreference = 'Stop'  # on error halt execution
$PSNativeCommandUseErrorActionPreference = $true  # on non-powershell binary failure inherit error action preference
```

However this doesn't work well with some cli tools that like to add to the stderr stream, as powershell treats those as failure. So then you have to either do `2>&1` or just handle LASTEXITCODE.

PR #1137, opened 2024-02-07
Update to latest tcgc

jhendrixMSFT:
This includes support for int8, int16, and float primitive types.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -121,17 +121,18 @@ export class typeAdapter {
       case 'armId':
       case 'boolean':
       case 'bytes':
-      case 'date':
       case 'decimal':
       case 'decimal128':
       case 'etag':
+      case 'float':
       case 'float32':
       case 'float64':
       case 'guid':
+      case 'int8':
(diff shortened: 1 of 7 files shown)


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -304,6 +306,7 @@ export class typeAdapter {
         this.types.set(etagKey, etag);
         return etag;
       }
+      case 'float': // C# and Java define float as 32 bits so we're following suit
```
RickWinter, line 309: This should be in the TypeSpec specification.  Is there a bug to spec out the size of a float?

RickWinter (approved):

packages/typespec-go/src/tcgcadapter/types.ts
jhendrixMSFT: Unfortunately, the size is absent, see the [docs](https://typespec.io/docs/language-basics/built-in-types#numeric-types). It could be the case that this is language dependent (if it is, the TS docs should call this out).

RickWinter: FYI @bterlson

Issue #1098, opened 2023-12-13
tsp: add support for hierarchical clients


jhendrixMSFT:
It will follow the same pattern as the client factory that we generate for ARM but will simply be called `Client`. We'll retain the name `ClientFactory` for ARM for back-compat.

PR #1141, opened 2024-02-13
Add support for hierarchical clients and client params

jhendrixMSFT:
Client accessor methods are now modeled in the Go code model. These are methods that return a child client instance that's been populated with values from the parent.
Client params are now properly adapted. This required an updated tcgc which properly models constant endpoint params as such. Fixed an issue where client host params were missing a cast or other type conversion to string.
Fixed ordering of method params so that body params are sequenced after any other required params.
Added some more test coverage and updated existing tests to use the new hierarchical clients as applicable.

Fixes https://github.com/Azure/autorest.go/issues/1098


packages/codegen.go/src/operations.ts
```diff
@@ -166,8 +173,24 @@ export async function generateOperations(codeModel: go.CodeModel): Promise<Array
       clientText += '}\n\n';
     }
 
-    // generate operations
+    // generate client accessors and operations
     let opText = '';
+    for (const clientAccessor of client.clientAccessors) {
+      const methodName = `New${clientAccessor.subClient.clientName}`;
```
chlowell, line 179: Would this work and if so, are there dragons?
```suggestion
      const methodName = `${clientAccessor.subClient.ctorName}`;
```

packages/typespec-go/test/cadlranch/azure/client-generator-core/accessgroup/zz_access_client.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) Go Code Generator. DO NOT EDIT.
+
+package accessgroup
+
+import "github.com/Azure/azure-sdk-for-go/sdk/azcore"
+
+// AccessClient contains the methods for the _Specs_.Azure.ClientGenerator.Core.Access group.
```
chlowell, line 12: Are real TSPs ugly like this? Wondering whether we should try to polish this string

packages/codegen.go/src/operations.ts
```diff
@@ -53,6 +56,8 @@ export async function generateOperations(codeModel: go.CodeModel): Promise<Array
     clientText += '// Don\'t use this type directly, use ';
     if (azureARM) {
       clientText += `${client.ctorName}() instead.\n`;
+    } else if (client.parent) {
+      clientText += `[${client.parent.clientName}.${client.clientName}] instead.\n`;
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed.

packages/codegen.go/src/operations.ts
jhendrixMSFT: Good catch, I missed `ctorName`. Fixed.

packages/typespec-go/test/cadlranch/azure/client-generator-core/accessgroup/zz_access_client.go
jhendrixMSFT: At present, this comes from the namespaces defined in the tsp. [This one](https://github.com/Azure/cadl-ranch/blob/main/packages/cadl-ranch-specs/http/azure/client-generator-core/access/main.tsp#L11) is indeed ugly, but I would expect that we'd never actually ship something with a name like this as some languages like C# etc directly use it in the generated namespace (i.e. not just a doc comment). [Example](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/ai/OpenAI.Assistants/assistants/routes.tsp#L10) from OpenAI.

chlowell (approved):

packages/typespec-go/test/cadlranch/azure/client-generator-core/accessgroup/zz_access_client.go
jhendrixMSFT: Hmm and maybe we should end the sentence with `namespace` instead of `group`? It kind of is leaking a source implementation detail though. I could go either way. Open to suggestions too.

jhendrixMSFT: This made me realize that `sdkClient` has a `description` field and that we should use that when available, falling back to a synthesized description (I changed it to `namespace` instead of `group`).

chlowell: Well, I dunno. I like "namespace" better than "group" because it's technically correct and that's the best kind of correctness, however I expect that will be irrelevant to virtually all readers of our docs because the comment speaks to someone trying to map a tsp to generated code--that's not a 3p user. Off the top of my head I don't know how we better serve those readers with this comment. Maybe that's not so important though, because we write better comments on constructors?

PR #1150, opened 2024-02-22
Add tsp support for LROs

jhendrixMSFT:
Hard-coded the name for the ResumeToken parameter type. Fixed a bug with tracking headers across operation responses. Emit azcore.ResponseError for Foundations.ErrorResponse type.


packages/typespec-go/test/cadlranch/azure/core/lro/lrorpcgroup/custom_client.go
```diff
@@ -0,0 +1,40 @@
+//go:build go1.18
+// +build go1.18
```
RickWinter, line 2: Do we need to keep generating this?

RickWinter (approved):

packages/typespec-go/test/cadlranch/azure/core/lro/lrorpcgroup/custom_client.go
jhendrixMSFT: This was copy/paste that predated its removal. Fixed.

PR #1154, opened 2024-03-04
Move response envelope SerDe to its own file

jhendrixMSFT:


packages/autorest.go/src/generator/generator.ts
```diff
@@ -108,10 +108,17 @@ export async function generateCode(host: AutorestExtensionHost) {
...
       host.writeFile({
         filename: `${filePrefix}responses.go`,
-        content: responses,
+        content: responses.responses,
+        artifactType: 'source-file-go'
+      });
+    }
+    if (responses.serDe.length > 0) {
+      host.writeFile({
+        filename: `${filePrefix}responses_serde.go`,
```
jhendrixMSFT, line 120: Technically, there's no "ser" for response envelopes, but `responses_de.go` seems weird.

chlowell (approved):
What's the motivation for this?

packages/autorest.go/src/generator/generator.ts
chlowell: Right, at first glance I'd guess that's German responses

jhendrixMSFT:
We're adding another case where we'll need to generate an unmarshaler on the response envelope. e.g. union types in TSP will likely require an unmarshaler. Cleans up the clutter, especially if we add more cases over time.

PR #1157, opened 2024-03-05
Canonicalize Content-Type header

jhendrixMSFT:
Request.SetBody will preserve an existing Content-Type header value but the check is done on the canonicalized header name.

### DIFF
--- a/packages/codegen.go/src/operations.ts
+++ b/packages/codegen.go/src/operations.ts
@@ -703,6 +703,10 @@ function createProtocolRequest(client: go.Client, method: go.Method | go.NextPag
     }
   }
   for (const param of headerParams.sort((a: go.HeaderParameter, b: go.HeaderParameter) => { return helpers.sortAscending(a.headerName, b.headerName);})) {
+    if (param.headerName.match(/^content-type$/)) {
+      // canonicalize content-type as req.SetBody checks for it via its canonicalized name :(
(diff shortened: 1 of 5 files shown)


packages/codegen.go/src/operations.ts
```diff
@@ -703,6 +703,10 @@ function createProtocolRequest(client: go.Client, method: go.Method | go.NextPag
     }
   }
   for (const param of headerParams.sort((a: go.HeaderParameter, b: go.HeaderParameter) => { return helpers.sortAscending(a.headerName, b.headerName);})) {
+    if (param.headerName.match(/^content-type$/)) {
+      // canonicalize content-type as req.SetBody checks for it via its canonicalized name :(
```
jhendrixMSFT, line 707: For reference: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/exported/request.go#L219

chlowell: My first thought was "we could change that comparison", but every change I came up with was worse than this

chlowell (approved):

packages/codegen.go/src/operations.ts
jhendrixMSFT: That was my initial thought as well, and I came to the same conclusion.

PR #1159, opened 2024-03-08
Update to the latest autorest.testserver

jhendrixMSFT:
Previous version no longer builds with recent versions of node. Updated some tests with expected changes, and added a few new ones. Include SDK logging when running tests.


packages/autorest.go/test/autorest/paginggroup/paginggroup_test.go
```diff
@@ -38,6 +39,23 @@ func httpClientWithCookieJar() policy.Transporter {
 	return http.DefaultClient
 }
 
+func TestAppendAPIVersion(t *testing.T) {
+	t.Skip("Go doesn't append api-version")
```
chlowell, line 43: Should we have this test then?

chlowell (approved):

packages/autorest.go/test/autorest/paginggroup/paginggroup_test.go
jhendrixMSFT: I wanted to be explicit that we're skipping this test (and others) and didn't just overlook them.

PR #1160, opened 2024-03-15
Consolidate module name and version into Module type

jhendrixMSFT:
Since the two must be set together or not at all, this reflects that requirement in the code model.
Moved module name and version validation to Module's constructor so it can be used by both code generators.
Move ARM constructor generation to its own function.

### DIFF
--- a/packages/codemodel.go/src/gocodemodel.ts
+++ b/packages/codemodel.go/src/gocodemodel.ts
@@ -62,16 +62,22 @@ export interface Options {
   // the module into which the package is being generated
   containingModule?: string;
 
-  // the module being generated
-  module?: string;
-
-  moduleVersion?: string;
+  // module and containingModule are mutually exclusive
+  module?: Module;
 
   azcoreVersion?: string;
 
   rawJSONAsBytes: boolean;
 
   sliceElementsByval: boolean;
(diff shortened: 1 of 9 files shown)


packages/codemodel.go/src/gocodemodel.ts
```diff
@@ -956,6 +962,24 @@ export class Options implements Options {
   }
 }
 
+export class Module implements Module {
+  constructor(name: string, version: string) {
+    if (name.match(/\/v\d+$/)) {
+      throw new Error('module name must not contain major version suffix');
+    }
+    if (!version.match(/^(\d+\.\d+\.\d+(?:-beta\.\d+)?)?$/)) {
+      throw new Error(`module version ${version} must in the format major.minor.patch[-beta.N]`);
```
chlowell, line 971:
```suggestion
      throw new Error(`module version ${version} must be in the format major.minor.patch[-beta.N]`);
```

chlowell (approved):

packages/codemodel.go/src/gocodemodel.ts
jhendrixMSFT: Fixed.

PR #1164, opened 2024-03-20
Fix tsp dev publish pipeline

jhendrixMSFT:
Use correct PAT var name.

### DIFF
--- a/packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/__debug/go-tester-pre.yaml
+++ b/packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/__debug/go-tester-pre.yaml
@@ -40238,11 +40238,11 @@ language:
     name: AppPlatformManagementClient
     description: ''
     azureARM: true
-    containingModule: none
+    containingModule: ''
     generateDateTimeRFC3339Helper: true
     groupParameters: true
     headAsBoolean: true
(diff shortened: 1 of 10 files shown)


packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/__debug/go-tester-pre.yaml
```diff
@@ -40238,11 +40238,11 @@ language:
     name: AppPlatformManagementClient
     description: ''
     azureARM: true
-    containingModule: none
+    containingModule: ''
     generateDateTimeRFC3339Helper: true
     groupParameters: true
     headAsBoolean: true
-    module: github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform/v2
+    module: github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform
```
RickWinter, line 40245: Was moving back from /v2 intentional?

RickWinter (approved):

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/__debug/go-tester-pre.yaml
jhendrixMSFT: This is just the intermediate metadata which has slightly changed. The end result is still the same, i.e. we generate the code with the `/v2` suffix as expected.

PR #1168, opened 2024-03-25
Add fakes for client accessors

jhendrixMSFT:
Fake method dispatch was split into its own, internal, method to handle clients that contain both client accessors and methods. Added helper initServer for populating client accessor transports. Added doc comments for fields in server factories.


packages/typespec-go/test/cadlranch/client/naminggroup/fake/zz_naming_server.go
```diff
@@ -14,10 +14,15 @@ import (
...
 	"net/http"
 	"net/url"
+	"strings"
+	"sync"
 )
 
 // NamingServer is a fake server for instances of the naminggroup.NamingClient type.
 type NamingServer struct {
+	// ModelServer contains the fakes for client ModelClient
+	ModelServer ModelServer
```
jhendrixMSFT, line 24: Here's an example where we mix client accessors with client methods.

packages/typespec-go/test/cadlranch/client/naminggroup/fakes_test.go
jhendrixMSFT, line 1: Added test covering mixed server and method fake dispatching.

packages/codegen.go/src/fake/servers.ts
```diff
@@ -193,13 +218,63 @@ export async function generateServers(codeModel: go.CodeModel): Promise<ServerCo
...
+  } else if (finalSubClients.length > 0) {
+    content += `\treturn ${receiverName}.${dispatchToClientFake}(req, method[:strings.Index(method, ".")])\n`;
+  } else {
+    content += `\treturn ${receiverName}.${dispatchMethodFake}(req, method)\n`;
+  }
+  content += '}\n\n'; // end Do
+  return content;
+}
+
+function generateServerTransportClientDispatch(serverTransport: string, subClients: Array<go.Client>, imports: ImportManager): string {
```
jhendrixMSFT, line 249: This is pretty much the same thing we do in `src/fake/factory.ts`. It will likely get consolidated at some point.

chlowell (approved):

PR #1171, opened 2024-03-27
Fix handling of content types other than JSON

jhendrixMSFT:
This removes some hard-coded JSON content types for parameters and responses and includes support for streaming bodies/responses. The changes to multipart/form will be cleaned up separately.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/clients.ts
+++ b/packages/typespec-go/src/tcgcadapter/clients.ts
@@ -223,6 +223,20 @@ export class clientAdapter {
     }
   }
 
+  private adaptContentType(contentTypeStr: string): 'binary' | 'JSON' | 'Text' | 'XML' {
+    // we only recognize/support JSON, text, and XML content types, so assume anything else is binary
+    // TODO: this seems goofy but maybe it's ok?
+    let contentType: go.BodyFormat = 'binary';
+    if (contentTypeStr.match(/json/i)) {
+      contentType = 'JSON';
+    } else if (contentTypeStr.match(/text/i)) {
+      contentType = 'Text';
+    } else if (contentTypeStr.match(/xml/i)) {
+      contentType = 'XML';
+    }
+    return contentType;
+  }
+
   private adaptMethodParameter(param: tcgc.SdkBodyParameter | tcgc.SdkHeaderParameter | tcgc.SdkPathParameter | tcgc.SdkQueryParameter, optionalGroup?: go.ParameterGroup): go.Parameter {
     if (param.isApiVersionParam && param.clientDefaultValue) {
       // we emit the api version param inline as a literal, never as a param.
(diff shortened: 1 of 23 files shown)


packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -223,6 +223,20 @@ export class clientAdapter {
     }
   }
 
+  private adaptContentType(contentTypeStr: string): 'binary' | 'JSON' | 'Text' | 'XML' {
+    // we only recognize/support JSON, text, and XML content types, so assume anything else is binary
+    // TODO: this seems goofy but maybe it's ok?
```
RickWinter, line 228: Is the goofy part that we assume binary?

It does seem odd that contentType isn't required. Is this by chance because the http spec says by default its binary?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -354,48 +366,78 @@ export class clientAdapter {
...
-      if (sdkResponse.name === 'OperationStatus' || sdkResponse.name === 'ResourceOperationStatus') {
-        return respEnv;
+    // we have a response type, determine the content type
+    let contentType: go.BodyFormat = 'binary';
+    if (sdkMethod.kind === 'lro') {
+      // we can't grovel through the operation responses for LROs as some of them
+      // return only headers, thus have no content type. while it's highly likely
+      // to only ever be JSON, this will be broken for LROs that return text/plain
+      // or a binary response. the former seems unlikely, the latter though...??
+      // TODO: follow up with tcgc team about this
```
RickWinter, line 396: Can you make this an issue so it doesn't get forgotten. Add issue ID here so we can remove when closed on.

packages/typespec-go/test/cadlranch/payload/multipartgroup/zz_formdata_client.go
```diff
@@ -20,7 +21,7 @@ type FormDataClient struct {
 
 // Basic - Test content-type: multipart/form-data
 //   - options - FormDataClientBasicOptions contains the optional parameters for the FormDataClient.Basic method.
-func (client *FormDataClient) Basic(ctx context.Context, body MultiPartRequest, options *FormDataClientBasicOptions) (FormDataClientBasicResponse, error) {
+func (client *FormDataClient) Basic(ctx context.Context, body io.ReadSeekCloser, options *FormDataClientBasicOptions) (FormDataClientBasicResponse, error) {
```
jhendrixMSFT, line 24: This is incorrect and will be fixed in a separate PR for multipart/form support.

RickWinter (approved):

packages/typespec-go/src/tcgcadapter/clients.ts
jhendrixMSFT: I just don't like having a heuristic like this. It's probably fine in practice for Azure though. I'll remove the `TODO` and if it comes back to bite us later we can improve it then.

RFC 822 says the default is `text/plain`. I don't know if this has been superseded though.

packages/typespec-go/src/tcgcadapter/clients.ts
jhendrixMSFT: Done.

PR #1181, opened 2024-03-29
Update to latest tcgc

jhendrixMSFT:
The arm flag is now on the sdkContext.
Client initialization is now mandatory and contains an access field that, when public, means the client is instantiable. The naming of child clients in hierarchical clients has changed. This is described in the code comments.
The modeling of host parameters has changed and includes host template parameters as part of the templateArguments property. HTTP response status codes can now include a status code range. In addition, the responses are now in a map.
Various nameInClient properties have been renamed to name. There is now only one body param for an operation. The possible api-versions are now modeled as an enum which we skip. The generatedName field is now a boolean indicating if the value of name was generated.


packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -33,25 +34,35 @@ export class clientAdapter {
...
+      //     ...
+      //   }
+      // }
+      // interface Completions { ... }
+      //
+      // we want to generate two clients from this,
+      // one name ChatCompletions and the other Completions
+
+      // strip off the Client suffix from the parent client name
+      clientName = parent.name.substring(0, parent.name.length - 6) + clientName;
```
jhendrixMSFT, line 61: This is responsible for the noise in the diff as it caused a lot of client and client accessor method names to change.

jhendrixMSFT: I should also add that SDKs without hierarchical clients weren't affected. i.e. the updated tcgc didn't change the generated code in any way.

packages/typespec-go/test/armlargeinstance/zz_azurelargeinstance_client.go
```diff
@@ -29,23 +29,23 @@ func NewAzureLargeInstanceClient(credential azcore.TokenCredential, options *arm
 	return client, nil
 }
 
-// NewAzureLargeInstancesClient creates a new instance of [AzureLargeInstancesClient].
-func (client *AzureLargeInstanceClient) NewAzureLargeInstancesClient() *AzureLargeInstancesClient {
-	return &AzureLargeInstancesClient{
+// NewAzureLargeInstanceAzureLargeInstancesClient creates a new instance of [AzureLargeInstanceAzureLargeInstancesClient].
```
chlowell, line 32: I think I understand how we end up with this and why the old naming algorithm was incorrect but yikes, these names are bad.

jhendrixMSFT: Agreed this one is particularly bad. The client for `armcodesigning` is a better showcase.

jhendrixMSFT: This example would have better names if `Azure` wasn't the prefix. The stuttering cleanup would then remove the `LargeInstance` prefix and things would make more sense.

packages/typespec-go/src/tcgcadapter/clients.ts
chlowell: Just spitballing--could we get better names by placing subclients in packages e.g. `foo/CompletionsClient` and `foo/chat/CompletionsClient`?

jhendrixMSFT: We could. Unfortunately, it starts to break down when you have structs etc shared across packages. We had considered such an approach early on in track 2 design. i.e. having each client in its own package, and structs etc in a `types` subpackage. The problem there is that you then start to deal with an "import list explosion"; something that used to take one import statement now takes at least two (more if you're using several clients).

Also, this would get extra complicated for hierarchical clients. With the current design, when a parent client creates a child client, it can access its internal fields to populate it. Having them in separate packages would break that. We could solve that with a bunch of type aliasing from an `internal` subpackage but I think that might get overly complicated.

jhendrixMSFT: In addition, if a model type has an unexported helper method, we can no longer access it from outside its defining package. It would mean that we'd need to duplicate the helper as an unexported func across packages that require it.

packages/typespec-go/test/armlargeinstance/zz_azurelargeinstance_client.go
jhendrixMSFT: Ok this has been fixed. I remembered that for Autorest we added the `stutter` command line option to specify a custom stuttering prefix. With that switch set here, the generated code looks much better.

chlowell (approved):

Issue #1193, opened 2024-04-08
feat: @autorest/gotest - explicitly opt-out go-linter


DariuszPorowski:
### Proposal

I's a proposal for go-linter to explicitly disable it by setting the flag in cli like `--go-linter="false"` or in autorest.md

```yaml
go-linter: false
```

### Context

Two use-cases:
- Generation for multiple packages we do in batch and after, we handle linting in one shot for everything. Having linting in the middle from go-linter does not produce any value in this point, just increase processing time.
- 1st stage of generation is on the agent without Go and then produced artifacts are transferred to another agent where Go exists for final linting, etc. Because of missing Go on the 1st agent, the autorest fails - it cannot execute any Go commands include goimports.

### Additional context

The default behaviors should remain the same, go-linter run without any end-user config changes.

I can work on the PR if you agree on this proposal.

CC: @tadelesh

PR #1197, opened 2024-04-09
gotest feature and fix from service team

tadelesh:
resolve: https://github.com/Azure/autorest.go/issues/1193
fix: https://github.com/Azure/autorest.go/issues/1203

### DIFF
--- a/packages/autorest.gotest/src/generator/goLinter.ts
+++ b/packages/autorest.gotest/src/generator/goLinter.ts
@@ -3,20 +3,22 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as path from 'path';
 import { AutorestExtensionHost } from '@autorest/extension-base';
-import { Config, configDefaults } from '../common/constant';
-import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
 import { TestConfig } from '@autorest/testmodeler/dist/src/common/testConfig';
+import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
+import * as path from 'path';
+import { Config, configDefaults } from '../common/constant';
 
 export async function processRequest(host: AutorestExtensionHost): Promise<void> {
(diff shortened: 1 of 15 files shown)


packages/autorest.gotest/src/template/exampleTest.go.njk
```diff
@@ -17,8 +17,8 @@ import (
 {%- set allExamples = [] %}
 {%- for exampleGroup in exampleGroups %}
 {%- for example in exampleGroup.examples %}
-{%- if allExamples.indexOf(example.originalFile)<0 %}
```
raych1, line 20: What's the difference between `originalFile` and `name`?

packages/autorest.gotest/src/generator/goLinter.ts
```diff
@@ -3,20 +3,22 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as path from 'path';
 import { AutorestExtensionHost } from '@autorest/extension-base';
-import { Config, configDefaults } from '../common/constant';
-import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
 import { TestConfig } from '@autorest/testmodeler/dist/src/common/testConfig';
+import { Helper } from '@autorest/testmodeler/dist/src/util/helper';
```
raych1, line 8: Seems we don't have lint rule enabled in this project?

raych1 (commented):
Does the CI failure matter?

packages/autorest.gotest/src/template/exampleTest.go.njk
tadelesh: `name` will always be there, but `originalFile` will be empty if `include-x-ms-examples-original-file` set to false.

packages/autorest.gotest/src/generator/goLinter.ts
tadelesh: what do you mean lint rule?

raych1 (approved):

PR #1199, opened 2024-04-09
Refactor complextHostParams

jhendrixMSFT:
Renamed the field to templatedHost which accurately describes how the value is interpreted. The M4 code model was left as-is to avoid breaking anything in gotest.
Fixed an issue for templated hosts with a single param value.

### DIFF
--- a/packages/autorest.go/src/transform/transform.ts
+++ b/packages/autorest.go/src/transform/transform.ts
@@ -564,6 +564,12 @@ async function processOperationRequests(session: Session<m4.CodeModel>) {
             if (!values(hostParams).where(p => p.language.go!.name === param.language.go!.name).any()) {
               hostParams.push(param);
             }
+            // we special-case fully templated host param, e.g. {endpoint}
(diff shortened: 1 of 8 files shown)


packages/autorest.go/src/transform/transform.ts
```diff
@@ -564,6 +564,12 @@ async function processOperationRequests(session: Session<m4.CodeModel>) {
             if (!values(hostParams).where(p => p.language.go!.name === param.language.go!.name).any()) {
               hostParams.push(param);
             }
+            // we special-case fully templated host param, e.g. {endpoint}
+            // as there's no need to do a find/replace in this case, we'd
+            // just directly use the endpoint param value.
+            if (!(<string>group.language.go!.host).match(/^\{\w+\}$/)) {
+              group.language.go!.complexHostParams = true;
```
jhendrixMSFT, line 571: This is the fix that accounts for the changes in `zz_client.go`.

chlowell (approved):

PR #1202, opened 2024-04-10
Model client constructors

jhendrixMSFT:
Added type Constructor to the Go code model for modeling constructors. Client types now have an array of Constructors, and the ctorName field has been removed.
Consolidated Client.hostParams into Client.parameters. ARM client constructors are now created via the code model. Handling of the host param has been simplified.

### DIFF
--- a/packages/codegen.go/src/clientFactory.ts
+++ b/packages/codegen.go/src/clientFactory.ts
@@ -9,79 +9,82 @@ import { contentPreamble, formatCommentAsBulletItem, formatParameterTypeName, so
 import { ImportManager } from './imports.js';
 
 
-// Creates the content for all <operation>.go files
+// Creates the content for client_factory.go (ARM only)
 export async function generateClientFactory(codeModel: go.CodeModel): Promise<string> {
-  let result = '';
   // generate client factory only for ARM
-  if (codeModel.type === 'azure-arm' && codeModel.clients) {
-    // the list of packages to import
-    const imports = new ImportManager();
-    
-    // there should be at most one client level param: subscriptionID for ARM, any exception is always a wrong swagger definition that we should fix
-    const allClientParams = new Array<go.Parameter>();
-    for (const clients of codeModel.clients) {
-      for (const clientParam of values(clients.parameters)) {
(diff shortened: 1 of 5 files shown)


packages/codegen.go/src/operations.ts
```diff
@@ -943,8 +951,10 @@ function isArrayOfDateTimeForMarshalling(paramType: go.PossibleType): { format:
   }
 }
 
+// returns true if the method requires a response handler.
+// this is used to unmarshal the response body, parse response headers, or both.
 function needsResponseHandler(method: go.Method): boolean {
-  return helpers.hasSchemaResponse(method) || method.responseEnvelope.headers.length > 0 || (go.isLROMethod(method) && !!method.responseEnvelope.result) || go.isPageableMethod(method);
+  return helpers.hasSchemaResponse(method) || method.responseEnvelope.headers.length > 0;
```
jhendrixMSFT, line 957: Unrelated, but was causing some headaches in other, pending work.

packages/codegen.go/src/clientFactory.ts
```diff
@@ -9,79 +9,82 @@ import { contentPreamble, formatCommentAsBulletItem, formatParameterTypeName, so
...
-        allClientParams.push(clientParam);
+  if (codeModel.type !== 'azure-arm' || codeModel.clients.length === 0) {
+    return '';
+  }
+
+  let result = '';
+  // the list of packages to import
+  const imports = new ImportManager();
+  
+  // there should be at most one client level param: subscriptionID for ARM, any exception is always a wrong swagger definition that we should fix
```
chlowell, line 23: Should we make this clear by throwing in that case?

chlowell (approved):

packages/codegen.go/src/clientFactory.ts
jhendrixMSFT: Thinking about this more, I wonder if this will always be true. It's pretty typical that there's only `subscriptionID` (or no client params), but I wonder if this is an actual invariant we need to enforce. I'm thinking that, for now, I will just remove the comment.

PR #1204, opened 2024-04-10
Move client options definition to Client

jhendrixMSFT:
Since there's one client options type per client, it makes more sense to define it on the Client and not per constructor.
Consolidate creation of client options.


RickWinter (approved):
Did we regenerate data plane and control plane to verify this has no public facing effects?

jhendrixMSFT:
Yes, that happens as part of CI.

RickWinter:
[like]  Rick Winter reacted to your message:
________________________________

Issue #1190, opened 2024-04-08
Consolidate example/test generation to use client factory


tadelesh:
Keep consistent with current example generation to use client factory to init client.

PR #1208, opened 2024-04-11
[autorest.gotest] use client factory to init client

Alancere:
https://github.com/Azure/autorest.go/issues/1190

Keep consistent with current example generation to use client factory to init client.
- fake test
- scenarios test
- sample


tadelesh (commented):
@Alancere could you add fake test to integration test. And could you help to validate the change for some of our current service module. Thanks.

packages/autorest.gotest/test/integrationtest/output/signalr/armsignalr/fake_test.go
```diff
@@ -0,0 +1,1718 @@
...
+}
+
+func TestFakeTest(t *testing.T) {
+	suite.Run(t, new(FakeTestSuite))
+}
+
+func (testsuite *FakeTestSuite) TestOperations_List() {
+	ctx := context.Background()
+	fakeServer := fake.OperationsServer{}
+	clientFactory, err := armsignalr.NewClientFactory(testsuite.subscriptionId, testsuite.cred, &arm.ClientOptions{
```
tadelesh, line 49: one service's client factory does not need to init multiple times

packages/autorest.gotest/test/integrationtest/output/signalr/armsignalr/fake_test.go
```diff
@@ -0,0 +1,1718 @@
...
+func (testsuite *FakeTestSuite) TestSignalR_ListBySubscription() {
+	ctx := context.Background()
+	fakeServer := fake.Server{}
+	clientFactory, err := armsignalr.NewClientFactory(testsuite.subscriptionId, testsuite.cred, &arm.ClientOptions{
+		ClientOptions: azcore.ClientOptions{
+			Transport: fake.NewServerTransport(&fakeServer),
+		},
+	})
+	testsuite.Require().NoError(err, "Failed to create client factory")
+	client := clientFactory.NewClient()
```
tadelesh, line 143: is that possible to init one client from one service only once?

packages/autorest.gotest/test/integrationtest/output/signalr/armsignalr/fake_test.go
Alancere: Transport needs to be set up separately for each operation
image

packages/autorest.gotest/test/integrationtest/output/signalr/armsignalr/fake_test.go
Alancere: same above

packages/autorest.gotest/test/integrationtest/output/signalr/armsignalr/fake_test.go
DariuszPorowski: @tadelesh @Alancere I know I'm not directly involved in the dev work here (so treat it as a suggestion), but isn't that what `NewServerFactoryTransport` is for? to match `ClientFactory` with `ServerFactory`? like:

```go
Transport: fake.NewServerFactoryTransport(&fakeServerFactory),
```

Example in your repo:
https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/test/compute/armcompute/fakes_test.go#L52

tadelesh (approved):

packages/autorest.gotest/CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
 
-This log was last generated on Mon, 08 Apr 2024 06:34:25 GMT and should not be manually modified.
+This log was last generated on Mon, 22 Apr 2024 09:31:17 GMT and should not be manually modified.
+
+## 4.7.3
+Mon, 22 Apr 2024 09:31:17 GMT
+
+### Patches
+
+- Test template initializes the client using the client factory
```
tadelesh, line 10:
```suggestion
- Consolidate to use client factory to initialize clients
```

PR #1211, opened 2024-04-11
Add autorest option to fix const stuttering

jhendrixMSFT:
To avoid breaking changes, the option is off by default.

### DIFF
--- a/packages/autorest.go/src/transform/namer.ts
+++ b/packages/autorest.go/src/transform/namer.ts
@@ -4,7 +4,7 @@
  *  --------------------------------------------------------------------------------------------  */
 
 import { Session } from '@autorest/extension-base';
-import { CodeModel, HttpHeader, HttpMethod, Language } from '@autorest/codemodel';
+import { ChoiceSchema, CodeModel, HttpHeader, HttpMethod, Language, SealedChoiceSchema } from '@autorest/codemodel';
 import { visitor, clone, values } from '@azure-tools/linq';
 import { createPolymorphicInterfaceName, ensureNameCase, getEscapedReservedName, packageNameFromOutputFolder, trimPackagePrefix, uncapitalize } from '../../../naming.go/src/naming.js';
 import { aggregateParameters, hasAdditionalProperties } from './helpers.js';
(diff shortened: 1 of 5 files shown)


jhendrixMSFT:
@gracewilcox FYI

packages/autorest.go/test/databoxedge/armdataboxedge/zz_constants.go
```diff
@@ -303,6 +257,52 @@ func PossibleDayOfWeekValues() []DayOfWeek {
 	}
 }
 
+// DeviceKind - The kind of the device.
+type DeviceKind string
+
+const (
+	DataBoxEdgeDeviceKindAzureDataBoxGateway    DeviceKind = "AzureDataBoxGateway"
```
RickWinter, line 264: Why isn't this `DeviceKindAzureDataBoxGateway`? The `DataBoxEdge` portion seems unnecessary.

jhendrixMSFT: Yes good catch, will fix.

jhendrixMSFT: Fixed

packages/autorest.go/src/transform/namer.ts
```diff
@@ -100,23 +100,69 @@ export async function namer(session: Session<CodeModel>) {
   model.language.go!.singleClient = singleClient;
 
   // fix up type names
-  const structNames = new Set<string>();
+  // final names are added to typeNames which is used to detect collisions
+  const typeNames = new Set<string>();
```
chlowell, line 104: Is this just `names` now that there are consts in it too?

chlowell (approved):

packages/autorest.go/src/transform/namer.ts
jhendrixMSFT: Pretty much, but I felt names was a little too generic. It's really `structAndConstNames` but I went with the shorter `typeNames`.

Issue #1214, opened 2024-04-16
tsp compile error: `unhandled property kind offsetDateTime`


Alancere:
typespec: https://github.com/Azure/azure-rest-api-specs-pr/tree/RPSaaSMaster/specification/computeschedule/ComputeSchedule.Management

![image](https://github.com/Azure/autorest.go/assets/46921893/d7ba81fc-5490-41ab-9c76-b5f61d07fd55)

Issue #942, opened 2023-03-29
When can codegen assume GMT/UTC/etc time


jhendrixMSFT:
Investigate if there's a way to know when generated code should set the time zone.

jhendrixMSFT:
For tsp, we know if it's UTC/GMT or not.

PR #1227, opened 2024-04-16
Handle tsp offsetDateTime type

jhendrixMSFT:
Distinguish between UTC and local time types in the code model. Fixed a regex that was too lenient.
Removed some commented out placeholder code.

Fixes https://github.com/Azure/autorest.go/issues/1214

### DIFF
--- a/packages/codemodel.go/src/gocodemodel.ts
+++ b/packages/codemodel.go/src/gocodemodel.ts
@@ -837,6 +837,8 @@ export interface TimeType extends QualifiedType {
   packageName: 'time';
 
   dateTimeFormat: DateTimeFormat;
+
+  utc: boolean;
 }
 
 export type MapValueType = PossibleType;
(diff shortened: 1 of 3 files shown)


packages/codemodel.go/src/gocodemodel.ts
```diff
@@ -837,6 +837,8 @@ export interface TimeType extends QualifiedType {
   packageName: 'time';
 
   dateTimeFormat: DateTimeFormat;
+
+  utc: boolean;
```
jhendrixMSFT, line 841: This just wires in the data. Complete UTC time handling is tracked in https://github.com/Azure/autorest.go/issues/942.

RickWinter (approved):
This had no effect on what is generated? Do we have a test gap?

jhendrixMSFT:
Yes. The cited spec is private so I can't port it to our CI.

PR #1322, opened 2024-05-06
Don't generate empty clients

jhendrixMSFT:
If a client has no methods, skip it. This is a workaround until the underlying issue in TCGC is fixed.
Don't create the fake directory if there are no fakes to generate. Bump emitter version for pending release.

### DIFF
--- a/packages/typespec-go/src/emitter.ts
+++ b/packages/typespec-go/src/emitter.ts
@@ -99,36 +99,38 @@ export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
   }
 
   if (context.options['generate-fakes'] === true) {
-    const fakesDir = context.emitterOutputDir + '/fake';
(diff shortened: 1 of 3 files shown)


packages/typespec-go/src/emitter.ts
```diff
@@ -99,36 +99,38 @@ export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
   }
 
   if (context.options['generate-fakes'] === true) {
-    const fakesDir = context.emitterOutputDir + '/fake';
-    await mkdir(fakesDir, {recursive: true});
     const serverContent = await generateServers(codeModel);
```
jhendrixMSFT, line 102: If you hide whitespace changes this change is way easier to understand.

chlowell (approved):

Issue #1323, opened 2024-05-08
Unreferenced Types


Alancere:
Generate unreferenced types such as `TrackedResourceBase`, `ArmResource`, `ArmResourceBase`

![image](https://github.com/Azure/autorest.go/assets/46921893/c3600c68-af6a-4549-a2f1-cfc815179e9e)

![image](https://github.com/Azure/autorest.go/assets/46921893/060ac186-8f2c-4cae-8cbf-25ecad30de50)

jhendrixMSFT:
This will be fixed in `0.1.0-dev.3` of the emitter. Note that it will require setting option `remove-unreferenced-types` to `true`.

Issue #1305, opened 2024-04-23
Error: didn't find model type name OperationStatusResult for response envelope OperationStatusClientGetResponse


Alancere:
typespec: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/deviceregistry/DeviceRegistry.Management/main.tsp#L76

![image](https://github.com/Azure/autorest.go/assets/46921893/f58de583-28d0-4d18-b4e5-7222e994599e)

jhendrixMSFT:
This is due to over-aggressive removal of LRO status monitor types. For the tests currently checked in, they aren't used, but in this case they are. Note that in both cases the `usage` has them marked as output which seems like a bug. @tadelesh can this be fixed on the tcgc side? If not, we'll need to improve the detection logic.

tadelesh:
I only saw it is used in response. Why `output` usage is wrong?

jhendrixMSFT:
When building https://github.com/Azure/autorest.go/tree/main/packages/typespec-go/test/tsp/CodeSigning.Management

This includes a model `ArmOperationStatusResourceProvisioningState` with usage `Output`. However, none of the operations actually return an instance of this type. In this case, it would be ideal for `usage` to be `undefined` so it's clear that the type isn't actually used.

jhendrixMSFT:
This will be fixed in `0.1.0-dev.3` of the emitter.

tadelesh:
The tcgc usage issues should be fixed in this pr: https://github.com/Azure/typespec-azure/pull/826

PR #1327, opened 2024-05-08
Add remove-unreferenced-types option to tsp emitter

jhendrixMSFT:
This mimics the behavior in the autorest code generator.

Fixes:
- https://github.com/Azure/autorest.go/issues/1305
- https://github.com/Azure/autorest.go/issues/1323

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -729,8 +729,6 @@ export class typeAdapter {
     const recursiveAddReferencedType = function(type: tcgc.SdkType): void {
       switch (type.kind) {
         case 'array':
-          recursiveAddReferencedType(type.valueType);
-          break;
         case 'dict':
           recursiveAddReferencedType(type.valueType);
           break;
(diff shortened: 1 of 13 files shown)


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -733,6 +720,77 @@ export class typeAdapter {
...
+    const referencedModels = new Set<string>();
+
+    const recursiveAddReferencedType = function(type: tcgc.SdkType): void {
+      switch (type.kind) {
+        case 'array':
+          recursiveAddReferencedType(type.valueType);
+          break;
+        case 'dict':
+          recursiveAddReferencedType(type.valueType);
+          break;
```
RickWinter, line 736:
```suggestion
        case 'array':
        case 'dict':
                    recursiveAddReferencedType(type.valueType);
          break;
```

RickWinter (approved):

packages/typespec-go/src/tcgcadapter/types.ts
jhendrixMSFT: Fixed

PR #1331, opened 2024-05-09
Omit param name for unused params in createRequest methods

jhendrixMSFT:
Recent versions of staticcheck flag the unused param, so omit it.


packages/codegen.go/src/helpers.ts
```diff
@@ -107,7 +113,7 @@ export function getCreateRequestParametersSig(method: go.Method | go.NextPageMet
 // e.g. "i, s"
 export function getCreateRequestParameters(method: go.Method): string {
   // split param list into individual params
-  const reqParams = getCreateRequestParametersSig(method).split(',');
+  const reqParams = getCreateRequestParametersSig(method, false).split(',');
```
chlowell, line 116: Is there an example for this? I'm failing to imagine a scenario where we wouldn't want the underscore. If the problem is that we don't have a reliable heuristic for identifying unused params, I wonder whether we should consider another approach like adding a post generation step that lets us actually lint the generated code or inspect it with gopls

jhendrixMSFT: This cases where we pass `false` are where we want just the param names so we can pass them to `fooCreateRequest` (I should probably add a comment explaining this). In `getCreateRequestParameters` we want the complete list of param names so we can pass them from `Foo` to `fooCreateRequest`. It is a bit hokey, but I left it as-is to keep the delta small.

jhendrixMSFT: I refactored this to remove the `boolean` param. While it does duplicate a few lines of code, I think the simplicity it introduces is worth it.

chlowell (approved):

PR #1332, opened 2024-05-09
Support tsp spread params

jhendrixMSFT:
tsp supports spreading a model's fields into discrete params. The values are bundled back into a single model which is then sent over the wire. This requires enumerating the tcgc method params, not the operation params, which correctly models this.
Added type PartialBodyParameter to the code model which represents a param that's part of a model that's been spread. The resultant model that contains the spread params is created inline as it doesn't need to be part of public surface area.


packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadmodel_client.go
```diff
@@ -0,0 +1,220 @@
...
+	"strings"
+)
+
+// SpreadModelClient contains the methods for the Parameters.Spread namespace.
+// Don't use this type directly, use [SpreadClient.NewSpreadModelClient] instead.
+type SpreadModelClient struct {
+	internal *azcore.Client
+}
+
+//   - bodyParameter - This is a simple model.
```
chlowell, line 24: Are the missing method names our problem or the tsp's?

chlowell (approved):

packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadmodel_client.go
jhendrixMSFT: It's the tsp. A lot of the cadl-ranch tests don't have doc comments for the operations. I agree it looks funny. We _could_ just omit the param docs if the operation doesn't have docs. However, given that we'd never ship a real tsp in this way, I don't feel super-strong about changing it.

Issue #1333, opened 2024-05-10
tsp: omit unreferenced base types


jhendrixMSFT:
Somewhat related to https://github.com/Azure/autorest.go/pull/1327

For tsp like `model Foo extends Bar {...}`, we fold `Bar`'s properties into `Foo`. In such a case, if `Bar` isn't actually referenced, it would be ideal to omit it from the generated code.

PR #1337, opened 2024-05-10
Omit base model types by default

jhendrixMSFT:
The properties of a model's base types are folded into the super type. If the base types aren't referenced, omit them.

Fixes https://github.com/Azure/autorest.go/issues/1333


packages/typespec-go/test/armdatabasewatcher/zz_models.go
```diff
@@ -4,7 +4,34 @@
...
 package armdatabasewatcher
 
-import "time"
+import (
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"time"
+)
+
+// ArmOperationStatusResourceProvisioningState - Standard Azure Resource Manager operation status response
+type ArmOperationStatusResourceProvisioningState struct {
```
jhendrixMSFT, line 13: This type is showing up due to the removal of option `remove-unreferenced-types=true`, which is expected. All of the ARM *resource types are still omitted.

RickWinter (commented):
The comment says we need to generate but not emit. Why are we changing this behavior?

Should we be treating internal baseModels the same way?

jhendrixMSFT:
The comment is very focused on languages that use OO inheritance to describe these relationships.

RickWinter (approved):

Issue #1213, opened 2024-04-16
client_factory.go renamed to client.go


Alancere:
The `client.go` generated by typespec-go seems to have the same functionality as `client_factory.go`
![image](https://github.com/Azure/autorest.go/assets/46921893/d1008c77-47a5-444e-a465-170f0e28a2f9)

jhendrixMSFT:
This uses the modeled hierarchical client defined in the tsp. We could consider renaming this to `ClientFactory`. Can that be done in a `client.tsp` file for Go?

tadelesh:
Redesign the logic for TypeSpec as it has pure client hierarchy. Use `@clientName` to rename the namespace to keep backward compatibility.

tadelesh:
Need further discussion:
how go deal with naming collision with client and subclient?
tcgc is forced client name ends with `Cleint` which makes it impossible to rename in `client.tsp`

PR #1345, opened 2024-05-15
Use client factory for tsp codegen

jhendrixMSFT:
To keep compat with existing ARM SDKs, use client factory instead of the hierarchical service client.
We must still adapt it in order for its parameters to propagate to the sub-clients, we just skip adding it to the code model so it isn't generated.

Fixes https://github.com/Azure/autorest.go/issues/1213


chlowell:
Looks like this would make breaking changes in modules generated from tsp with the current code. I guess that's okay because we haven't shipped a stable one of those, but what about modules generated from swagger? Would this make switching to tsp a breaking change?

jhendrixMSFT:
This will prevent a breaking change when migrating from swagger to tsp. The tsp-defined hierarchical client is almost identical to the client factory that we synthesize. e.g. if you diff `armapicenter/zz_client.go` and `armapicenter/zz_client_factory.go`, other than some sorting differences of unexported fields, the only differences in public surface area are the client and client ctor name.

chlowell (approved):

Issue #1341, opened 2024-05-14
`OperationStatusResult.Error` has inconsistent definitions


Alancere:
tsp: https://github.com/Azure/typespec-azure/blob/main/packages/typespec-azure-resource-manager/lib/common-types/types.tsp#L232

![image](https://github.com/Azure/autorest.go/assets/46921893/fc32d369-5306-4a3e-af28-9eb231b7dd52)

jhendrixMSFT:
Repro steps?

PR #1346, opened 2024-05-15
Don't replace error types with azcore.ResponseError

jhendrixMSFT:
Instances of azcore.ResponseError are returned when an operation fails and shouldn't be embedded in modeled error types.
Removed omission of foundation error types. It wasn't a great heuristic and has been superseded by the remove-unreferenced-types option. Include client ctor params when scanning for unreferenced types. Make remove-unreferenced-types the default for regeneration script.

Fixes https://github.com/Azure/autorest.go/issues/1341


packages/typespec-go/.scripts/tspcompile.js
```diff
@@ -117,16 +117,16 @@ function should_generate(name) {
...
-generate('armapicenter', armapicenter, 'test/armapicenter', ['remove-unreferenced-types=true']);
+generate('armapicenter', armapicenter, 'test/armapicenter');
 
 const armlargeinstance = pkgRoot + 'test/tsp/AzureLargeInstance.Management';
-generate('armlargeinstance', armlargeinstance, 'test/armlargeinstance', ['stutter=AzureLargeInstance', 'remove-unreferenced-types=true']);
+generate('armlargeinstance', armlargeinstance, 'test/armlargeinstance', ['stutter=AzureLargeInstance']);
 
 const armdatabasewatcher = pkgRoot + `test/tsp/DatabaseWatcher.Management`;
-generate('armdatabasewatcher', armdatabasewatcher, 'test/armdatabasewatcher');
+generate('armdatabasewatcher', armdatabasewatcher, 'test/armdatabasewatcher', ['remove-unreferenced-types=false']);
```
chlowell, line 129: Do we now get `--option="...remove-unreferenced-types=false" --option="...remove-unreferenced-types=true"` in the command line?

jhendrixMSFT: Yes. Last one appears to win. It was the simplest solution instead of wiring up stuff to change the default value.

jhendrixMSFT: Fair enough. I'll improve the script.

packages/typespec-go/test/armdatabasewatcher/zz_models.go
```diff
@@ -57,6 +54,42 @@ type Datastore struct {
 	KustoClusterDisplayName *string
 }
 
+// ErrorAdditionalInfo - The resource management error additional info.
```
RickWinter, line 57: The structures appear to be the same in all the tests. Is that by chance or are these types that could be shared from the armCore library?

jhendrixMSFT: I believe this, and other types, come from `@azure-tools/typespec-azure-resource-manager`.

packages/typespec-go/.scripts/tspcompile.js
jhendrixMSFT: Done.

chlowell (approved):

Issue #1354, opened 2024-05-20
ResumeToken is missing Comment


Alancere:
tsp: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/devopsinfrastructure/Microsoft.DevOpsInfrastructure/main.tsp

![image](https://github.com/Azure/autorest.go/assets/46921893/4621481c-0447-4484-b9af-fca0d7c3bd28)

PR #1355, opened 2024-05-20
Move resume token doc comment to code model

jhendrixMSFT:
Implicitly applies to all code generators.

Fixes https://github.com/Azure/autorest.go/issues/1354

### DIFF
--- a/packages/codemodel.go/src/gocodemodel.ts
+++ b/packages/codemodel.go/src/gocodemodel.ts
@@ -1502,6 +1502,7 @@ export class ResumeTokenParameter implements ResumeTokenParameter {
     this.paramType = 'optional';
     this.byValue = true;
     this.location = 'method';
+    this.description = 'Resumes the LRO from the provided token.';
   }
 }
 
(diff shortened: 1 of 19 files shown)


packages/codemodel.go/src/gocodemodel.ts
```diff
@@ -1502,6 +1502,7 @@ export class ResumeTokenParameter implements ResumeTokenParameter {
     this.paramType = 'optional';
     this.byValue = true;
     this.location = 'method';
+    this.description = 'Resumes the LRO from the provided token.';
```
chlowell, line 1505: Should we spell out the acronym as `long-running operation (LRO)`? I'm imagining many readers will figure out what the token is for by its name but not understand the acronym

chlowell (approved):

packages/codemodel.go/src/gocodemodel.ts
jhendrixMSFT: Good point, I've made this change.

PR #1370, opened 2024-05-29
Add tsp versioning tests

jhendrixMSFT:
Updated tspcompile script with notes for groups that don't build.


packages/typespec-go/test/cadlranch/versioning/rettypechangedfromgroup/returntypechangedfrom_client_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"context"
+	"rettypechangedfromgroup"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Test(t *testing.T) {
```
RickWinter, line 14: Can we make the test name more descriptive of the scenario.

RickWinter (approved):

packages/typespec-go/test/cadlranch/versioning/rettypechangedfromgroup/returntypechangedfrom_client_test.go
jhendrixMSFT: The test name comes from the method name on the client. In this case, the method name is `Test`.

PR #1371, opened 2024-05-30
Add some tsp coverage for authentication

jhendrixMSFT:
It's hand-written but shows we can support it and gets coverage above 0%.


chlowell:
Is it really meaningful to cover this with handwritten code? Doing so doesn't test the generator.

jhendrixMSFT:
That was my first thought too. However, after thinking about it a bit more I think there is some value.
- if/when we generate this, we shouldn't see a delta as the hand-written code follows the design guidelines
- it verifies we can support the underlying patterns. it does highlight a sharp edge though, how will we set `InsecureAllowCredentialWithHTTP` for these tests

PR #1381, opened 2024-06-11
Rename ParameterType to ParameterKind

jhendrixMSFT:
The name was a bit confusing as Parameter also has a type which is the type system type.

No functional changes.


chlowell (approved):
👍 but I'm not sure about `placement` because it implies a locality that doesn't seem to fit the concept. `kind`?

jhendrixMSFT:
Good idea, updated.

Issue #1393, opened 2024-06-22
rawjson-as-bytes has unstable behavior


jim-minter:
Background: with `rawjson-as-bytes: true`, I would like to be able to specify a swagger definition fragment as follows:

```json
"definitions": {
  "Example": {
    "type": "object",
    "properties": {
      "rawslice": {
        "type": "array",
        "items": {
          "type": "object"
        }
      }
    }
  }
}
```

and get a Go struct that looks like this:

```go
type Example struct {
	Rawslice [][]byte
}
```

But that's **not** what happens.  Instead I get the following, as if `rawjson-as-bytes` were false:

```go
type Example struct {
	Rawslice []any
}
```

Now, the thing that's weird is that if I add a second definition to the swagger fragment that will also use rawjson-as-bytes functionality, the Go type of `Example` **unexpectedly changes** from what I don't want to what I do want.

```json
"definitions": {
  "Example": {
    "type": "object",
    "properties": {
      "rawslice": {
        "type": "array",
        "items": {
          "type": "object"
        }
      }
    }
  },
  "ZZZAnotherThing": {
    "type": "object",
    "properties": {
      "raw": {
        "type": "object"
      }
    }
  }
}
```

results in:

```go
type Example struct {
	Rawslice [][]byte // changed from []any!
}

type ZZZAnotherThing struct {
	// Anything
	Raw []byte
}
```

Unfortunately, however, if I try to `json.Unmarshal` a payload like `{"rawslice": [{}, {}]}` into an `Example`, it fails with an error like `json: cannot unmarshal object into Go value of type []uint8`.

So there's a couple of issues here:
1. First and foremost, it's surely a bug that the behavior here is unstable, which the main reason I'm opening this issue.
2. It would be great if the [][]byte use case that I'm looking for could either be implemented or explicitly validated against.  FWIW I have an acceptable workaround where I can unmarshal a `raw-json-as-bytes` []byte to a []json.RawMessage myself in the application layer, so I don't mind too much whether my feature request is implemented or not.  I feel like it's pretty esoteric and I'd be inclined to block it.

Here's the yaml config used:

```yaml
file-prefix: zz_generated_
generate-fakes: false
go: true
input-file:
- example.json
rawjson-as-bytes: true
track2: true
use-extension:
  "@autorest/modelerfour": "4.27.0"
  "@autorest/go": "4.0.0-preview.63"
version: 3.10.1
```

Many thanks!

PR #1394, opened 2024-06-24
Propagate raw JSON as bytes setting to array element schema

jhendrixMSFT:
This ensures that we generate the type as [][]byte. The original intention was to generate this a []byte but that never actually happened and we can't change it now.

Fixes https://github.com/Azure/autorest.go/issues/1393


chlowell:
Please forgive the stupid question, but why can't we change the generated type? The linked issue implies `[][]byte` doesn't work today; could anyone be broken if we start generating `[]byte` instead?

jhendrixMSFT:
Let me clarify as there are a few angles to this problem.

Enabling `rawjson-as-bytes` was not working for arrays of untyped elements in some cases. The expectation was to generate `[][]byte` (i.e. a slice of raw JSON items) but we were generating `[]any`.

When I originally added the `rawjson-as-bytes` feature, I intended to always emit `[]byte` even if the swagger stated an array of untyped objects. You could argue that what we generate with this fix is technically correct, i.e. if the swagger states an untyped object you get `[]byte` and if it states an array of untyped objects you get `[][]byte`. Whether or not it's better I suppose depends on one's perspective (I can see arguments both ways).

All that said, we have to leave it as `[][]byte` at this point to avoid any breaking changes (I believe KV would be impacted by such a change).

chlowell:
👍 thanks for the context. That makes sense and I see how `[][]byte` is easier than `[]byte` in application code.

chlowell (approved):

Issue #1397, opened 2024-07-02
Client comment


Alancere:
client comment should be operation group, not top-level parent namespace.

tsp: https://github.com/Azure/azure-rest-api-specs/tree/main/specification/mongocluster/DocumentDB.MongoCluster.Management

![image](https://github.com/Azure/autorest.go/assets/46921893/b8900c16-e196-4b34-b338-ac9d99586d93)

PR #1398, opened 2024-07-02
fix client struct type comment

tadelesh:
fix: https://github.com/Azure/autorest.go/issues/1397

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/clients.ts
+++ b/packages/typespec-go/src/tcgcadapter/clients.ts
@@ -76,7 +76,9 @@ export class clientAdapter {
     if (sdkClient.description) {
       description = `${clientName} - ${sdkClient.description}`;
     } else {
-      description = `${clientName} contains the methods for the ${clientName.substring(0, clientName.length - 6)} group.`;
+      // strip clientName's "Client" suffix
+      const groupName = clientName.substring(0, clientName.length - 6)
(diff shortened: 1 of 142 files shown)


packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -76,7 +76,7 @@ export class clientAdapter {
     if (sdkClient.description) {
       description = `${clientName} - ${sdkClient.description}`;
     } else {
-      description = `${clientName} contains the methods for the ${sdkClient.nameSpace} namespace.`;
+      description = `${clientName} contains the methods for the ${clientName.substring(0, clientName.length - 6)} group.`;
```
chlowell, line 79: I wondered about these 6 characters, so I believe it will help future readers to explain a little:
```suggestion
      // strip clientName's "Client" suffix
      const groupName = clientName.substring(0, clientName.length - 6)
      description = `${clientName} contains the methods for the ${groupName} group.`;
```
This also made me wonder: is there a more straightforward way to access the group name?

chlowell (approved):

packages/typespec-go/src/tcgcadapter/clients.ts
tadelesh: for typespec, there will be no operation group concept, replaced with sub-client.

Issue #1358, opened 2024-05-21
Example generation should support TypeSpec


tadelesh:
Detail tasks TBD.

PR #1399, opened 2024-07-15
example generation for typespec

tadelesh:
resolve: https://github.com/Azure/autorest.go/issues/1358

basic design concept:
1. add example in go code model
2. map tcgc example model into go code model, do the mapping in the same code generator adapter
3. use go code model to generate example files, do the generation in the same code generator processor
4. TODO: need to migrate swagger go test generator logic, adapt swagger test modeler example model to go code model, but currently low priority.


packages/codemodel.go/src/client.ts
```diff
@@ -181,6 +184,7 @@ export class Method implements Method {
     this.name = name;
     this.naming = naming;
     this.parameters = new Array<param.Parameter>();
+    this.examples = [];
```
jhendrixMSFT, line 187: `this.examples = new Array<MethodExample>();`

jhendrixMSFT: See linter errors for more changes.

packages/typespec-go/src/tcgcadapter/adapter.ts
```diff
@@ -15,14 +15,14 @@ const headerText = `Copyright (c) Microsoft Corporation. All rights reserved.
...
-export function tcgcToGoCodeModel(context: EmitContext<GoEmitterOptions>): go.CodeModel {
+export async function tcgcToGoCodeModel(context: EmitContext<GoEmitterOptions>): Promise<go.CodeModel> {
   const info = new go.Info('TODO Title');
   const options = new go.Options(headerText, context.options['generate-fakes'] === true, context.options['inject-spans'] === true, context.options['disallow-unknown-fields'] === true);
   if (context.options['azcore-version']) {
     options.azcoreVersion = context.options['azcore-version'];
   }
 
-  const sdkContext = tcgc.createSdkContext(context);
+  const sdkContext = await tcgc.createSdkContext(context);
```
jhendrixMSFT, line 25: Why was this function made async?

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -3,13 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as naming from '../../../naming.go/src/naming.js';
-import * as go from '../../../codemodel.go/src/index.js';
+import { uncapitalize } from '@azure-tools/codegen';
```
jhendrixMSFT, line 6: In the future, it would be ideal to omit such refactoring from PRs with functional changes to reduce the delta, making it easier for reviewers to focus on the actual changes.

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -149,20 +149,20 @@ export class typeAdapter {
   // returns the Go code model type for the specified SDK type.
   // the operation is idempotent, so getting the same type multiple times
   // returns the same instance of the converted type.
-  getPossibleType(type: tcgc.SdkType, elementTypeByValue: boolean, substituteDiscriminator: boolean): go.PossibleType {
+  getPossibleType(type: tcgc.SdkType, elementTypeByValue: boolean, substituteDiscriminator: boolean, isMulti = false): go.PossibleType {
```
jhendrixMSFT, line 152: What does `isMulti` mean/do?

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -352,16 +353,6 @@ export class typeAdapter {
         this.types.set(decimalKey, decimalType);
         return decimalType;
       }
-      case 'eTag': {
```
jhendrixMSFT, line 355: Was `eTag` and the other removed types converted to `string`?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -263,13 +265,15 @@ export class clientAdapter {
       }
     }
 
-    this.adaptMethodParameters(sdkMethod, method);
+    this.adaptMethodParameters(sdkMethod, method, paramMapping);
```
jhendrixMSFT, line 268: It looks like `adaptMethodParameters` is populating `paramMapping` yes? If so, can we change this so `adaptMethodParameters` returns it instead of using an out param?

packages/codegen.go/src/clientFactory.ts
```diff
@@ -87,3 +78,17 @@ export async function generateClientFactory(codeModel: go.CodeModel): Promise<st
...
+    for (const clientParam of values(clients.parameters)) {
+      if (values(allClientParams).where(param => param.name === clientParam.name).any()) {
+        continue;
+      }
+      allClientParams.push(clientParam);
+    }
+  }
+  allClientParams.sort(sortParametersByRequired);
+  return allClientParams;
+}
```
jhendrixMSFT, line 94: Missing new-line.

packages/codegen.go/src/clientFactory.ts
```diff
@@ -87,3 +78,17 @@ export async function generateClientFactory(codeModel: go.CodeModel): Promise<st
   result = contentPreamble(codeModel) + imports.text() + result;
   return result;
 }
+
+export function getAllClientParameters(codeModel: go.CodeModel): Array<go.Parameter> {
```
jhendrixMSFT, line 82: This should probably move to `helpers.js` if it needs to be used in multiple places.

packages/codegen.go/src/example.ts
```diff
@@ -0,0 +1,404 @@
...
+  readonly content: string;
+
+  constructor(name: string, content: string) {
+    this.name = name;
+    this.content = content;
+  }
+}
+
+// Creates the content for all <example>.go files
+export async function generateExamples(codeModel: go.CodeModel): Promise<Array<ExampleContent>> {
```
jhendrixMSFT, line 25: Do you think there's value in adding an opt-out flag to disable example generation? I'm not advocating for or against, just wondering.

packages/typespec-go/src/tcgcadapter/adapter.ts
tadelesh: this is from the latest breaking change of tcgc 0.44.2

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: sure. do you think i need to revert it from this pr?

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: it come from the new multi-part feature of typespec and related change from tcgc, also see https://github.com/Azure/typespec-azure/pull/1237. i think i should move the upgrade of new tcgc to a new pr which could make this pr focus on example only.

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: same from: https://github.com/Azure/typespec-azure/pull/1237

packages/typespec-go/src/tcgcadapter/adapter.ts
tadelesh: move to this pr: https://github.com/Azure/autorest.go/pull/1411

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: will remove after https://github.com/Azure/typespec-azure/pull/1251 release

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: move to: https://github.com/Azure/autorest.go/pull/1411

packages/typespec-go/src/tcgcadapter/types.ts
jhendrixMSFT: Agreed on the consistency. I will work on updating CI to enforce this.

packages/typespec-go/src/tcgcadapter/types.ts
jhendrixMSFT: The motivation for the change makes sense. However, my concern here is that we no longer have a predefined union of type names, so we don't know the possible "sub-type" when writing the code and are left with magic strings.

If it's not possible to define a union for these sub-types, can we at least define some constants, or document the possible sub-types someplace? Maybe it's already there in some form and I just missed it?

tadelesh: sure. created an issue: https://github.com/Azure/typespec-azure/issues/1260.

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: reverted all format change.

packages/typespec-go/src/tcgcadapter/types.ts
ArcturusZhang: So the ask here is to document the `crossLanguageDefinitionId`s of those azure core types?
I do not think we should include a union in TCGC to enable that - the initial motivation of moving them out from the built-in type is trying not to make TCGC strongly bond with azure when we do not have to.
I think we could put those crossLanguageDefinitionId in a document for azure types.

tadelesh: correct. we could put the doc here: https://azure.github.io/typespec-azure/docs/howtos/DataPlane%20Generation%20-%20DPG/07tcgcTypes

jhendrixMSFT: Can you please add a link to this from the tcgc code so it's easy to find? Who will own maintaining it over time?

tadelesh:
@jhendrixMSFT do you have further concern for this pr?

jhendrixMSFT:
I had a few questions that were never answered (see above).

packages/codegen.go/src/example.ts
tadelesh: added.

packages/codegen.go/src/clientFactory.ts
tadelesh: moved.

packages/codegen.go/src/clientFactory.ts
tadelesh: added

packages/typespec-go/src/tcgcadapter/clients.ts
tadelesh: changed.

tadelesh:
> I had a few questions that were never answered (see above).

oh, i just changed the code according to the comment and resolved the comment directly. added comment.

jhendrixMSFT (approved):

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: after the init version, i could help to keep it up-to-date. the scalar types that need special handling should be limited.

PR #1400, opened 2024-07-15
Add release pipeline for typespec-go emitter

jhendrixMSFT:
Pipeline was copied from publish-release-go.yml and slightly modified.

### DIFF
--- a/eng/pipelines/publish-release-typespec.yml
+++ b/eng/pipelines/publish-release-typespec.yml
@@ -1,0 +1,50 @@
+trigger: none
+pr: none
+
+
+extends: 
+  template: /eng/pipelines/templates/stages/1es-redirect.yml
+  parameters:
+    stages:
+      - stage: Release
+
+        variables: 
+          - template: /eng/pipelines/templates/variables/image.yml
+
+        jobs:
+          - job: Release
+            pool:
+              name: $(LINUXPOOL)
+              image: $(LINUXVMIMAGE)
+              os: linux
+
+            steps:
+              - template: /eng/steps/set-env.yaml
+
+              - template: /eng/steps/build-test-typespec.yaml
+
+              - pwsh: |
+                  $currentVersion = node -p -e "require('./packages/typespec-go/package.json').version";
+                  cd packages/typespec-go
+                  npm version --no-git-tag-version $currentVersion
+                  npm pack;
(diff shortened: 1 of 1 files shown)


RickWinter (approved):

jhendrixMSFT:
@benbp can you help create this pipeline when you have some time?

jhendrixMSFT:
@weshaggard is this something you can help with in Ben's absense?

eng/pipelines/publish-release-typespec.yml
```diff
@@ -0,0 +1,50 @@
...
+                    --tag v$currentVersion --notes='TypeSpec emitter for Go SDKs' `
+                    --prerelease --editRelease false `
+                    --assets azure-tools-typespec-go-$currentVersion.tgz `
+                    --target_commitish $(Build.SourceBranchName);
+                displayName: "Publish GitHub Release"
+
+              - script: |
+                  cd packages/typespec-go
+                  echo "//registry.npmjs.org/:_authToken=$(azure-sdk-npm-token)" > ./.npmrc
+                  npm publish --access public
```
weshaggard, line 49: FYI - This is going to stop working in the future as we are losing permission to publish directly to npm and have to use ESRP publishing instead.

RickWinter: Whats the compliant way everyone is using? We should do the same

weshaggard: https://github.com/Azure/azure-sdk-for-js/blob/main/eng/pipelines/templates/steps/npm-release-task.yml

That is the publishing changes we made in the JS repo. The key will be the EsrpRelease step which replaces npm publish.

jhendrixMSFT: Should we do this now? The Shanghai team is waiting on a beta release of `typespec-go` for the MPG effort this month.

weshaggard: I can continue what you are doing for now we just need to get this updated sooner rather than later.

jhendrixMSFT: Opened https://github.com/Azure/autorest.go/issues/1401 to track this.

jhendrixMSFT: Thanks Wes. Let me know when the pipeline is ready so I can publish a release.

PR #1405, opened 2024-07-22
upgrade to latest tsp

tadelesh:
upgrade to typespec 0.58 and typespec-azure 0.44


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -417,9 +417,6 @@ export class typeAdapter {
         return stringType;
       }
       case 'plainTime': {
-        if (type.encode !== 'rfc3339') {
```
tadelesh, line 420: tcgc's encode for plainTime is always `plainTime`, remove to bypass the error

packages/typespec-go/test/armapicenter/zz_models.go
```diff
@@ -558,22 +560,6 @@ type ServiceProperties struct {
...
-	// Resource tags.
-	Tags map[string]*string
-}
-
-// ServiceUpdateProperties - The updatable properties of the Service.
-type ServiceUpdateProperties struct {
-	// Flag used to restore soft-deleted API Center service. If specified and set to 'true' all other properties will be ignored.
-	Restore *bool
-}
-
```
tadelesh, line 576: change of update model is from typespec arm template breaking change

packages/typespec-go/test/cadlranch/azure/core/lro/lrorpcgroup/zz_rpc_client.go
```diff
@@ -45,13 +45,13 @@ func (client *RPCClient) BeginLongRunningRPC(ctx context.Context, generationOpti
 // If the operation fails it returns an *azcore.ResponseError type.
 //
 // Generated from API version 2022-12-01-preview
-func (client *RPCClient) longRunningRPC(ctx context.Context, generationOptions GenerationOptions, options *RPCClientBeginLongRunningRPCOptions) (*http.Response, error) {
+func (client *RPCClient) longRunningRPC(ctx context.Context, body GenerationOptions, options *RPCClientBeginLongRunningRPCOptions) (*http.Response, error) {
```
tadelesh, line 48: body param naming change is from spread behavior breaking change from tcgc

packages/typespec-go/test/cadlranch/azure/special-headers/xmsclientreqidgroup/zz_xmsclientrequestid_client.go
```diff
@@ -44,7 +44,7 @@ func (client *XMSClientRequestIDClient) Get(ctx context.Context, options *XMSCli
 
 // getCreateRequest creates the Get request.
 func (client *XMSClientRequestIDClient) getCreateRequest(ctx context.Context, options *XMSClientRequestIDClientGetOptions) (*policy.Request, error) {
-	urlPath := "/azure/special-headers/x-ms-client-request-id"
+	urlPath := "/azure/special-headers/x-ms-client-request-id/"
```
tadelesh, line 47: suffix slash change is from typespec core breaking change

packages/typespec-go/test/cadlranch/parameters/bodyoptionalgroup/zz_bodyoptionality_client.go
```diff
@@ -68,13 +68,13 @@ func (client *BodyOptionalityClient) requiredExplicitCreateRequest(ctx context.C
 // If the operation fails it returns an *azcore.ResponseError type.
 //   - options - BodyOptionalityClientRequiredImplicitOptions contains the optional parameters for the BodyOptionalityClient.RequiredImplicit
 //     method.
-func (client *BodyOptionalityClient) RequiredImplicit(ctx context.Context, bodyModel BodyModel, options *BodyOptionalityClientRequiredImplicitOptions) (BodyOptionalityClientRequiredImplicitResponse, error) {
+func (client *BodyOptionalityClient) RequiredImplicit(ctx context.Context, name string, options *BodyOptionalityClientRequiredImplicitOptions) (BodyOptionalityClientRequiredImplicitResponse, error) {
```
tadelesh, line 71: spread behavior change

packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadalias_client.go
```diff
@@ -114,17 +114,123 @@ func (client *SpreadAliasClient) spreadAsRequestParameterCreateRequest(ctx conte
...
 // SpreadWithMultipleParameters -
 // If the operation fails it returns an *azcore.ResponseError type.
+//   - requiredString - required string
+//   - optionalInt - optional int
+//   - requiredIntList - required int
+//   - optionalStringList - optional string
 //   - options - SpreadAliasClientSpreadWithMultipleParametersOptions contains the optional parameters for the SpreadAliasClient.SpreadWithMultipleParameters
 //     method.
-func (client *SpreadAliasClient) SpreadWithMultipleParameters(ctx context.Context, id string, xMSTestHeader string, prop1 string, prop2 string, prop3 string, prop4 string, prop5 string, prop6 string, options *SpreadAliasClientSpreadWithMultipleParametersOptions) (SpreadAliasClientSpreadWithMultipleParametersResponse, error) {
+func (client *SpreadAliasClient) SpreadWithMultipleParameters(ctx context.Context, id string, xMSTestHeader string, requiredString string, optionalInt int32, requiredIntList []int32, optionalStringList []string, options *SpreadAliasClientSpreadWithMultipleParametersOptions) (SpreadAliasClientSpreadWithMultipleParametersResponse, error) {
```
tadelesh, line 227: change from cadl-ranch test change

packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadmodel_client.go
```diff
@@ -107,16 +111,15 @@ func (client *SpreadModelClient) spreadCompositeRequestCreateRequest(ctx context
 
 // SpreadCompositeRequestMix -
 // If the operation fails it returns an *azcore.ResponseError type.
-//   - compositeRequestMix - This is a model with non-body http request decorator.
 //   - options - SpreadModelClientSpreadCompositeRequestMixOptions contains the optional parameters for the SpreadModelClient.SpreadCompositeRequestMix
 //     method.
-func (client *SpreadModelClient) SpreadCompositeRequestMix(ctx context.Context, name string, testHeader string, compositeRequestMix CompositeRequestMix, options *SpreadModelClientSpreadCompositeRequestMixOptions) (SpreadModelClientSpreadCompositeRequestMixResponse, error) {
+func (client *SpreadModelClient) SpreadCompositeRequestMix(ctx context.Context, name string, testHeader string, prop string, options *SpreadModelClientSpreadCompositeRequestMixOptions) (SpreadModelClientSpreadCompositeRequestMixResponse, error) {
```
tadelesh, line 116: spread behavior change

packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadalias_client.go
jhendrixMSFT: Hmm, I think the optional params should be placed in the options type.

jhendrixMSFT: Looks like tcgc doesn't properly propagate the optional value for these params. e.g. for `optionalInt` I see it set to `optional: false`.

jhendrixMSFT: Opened https://github.com/Azure/typespec-azure/issues/1218 to track it.

RickWinter: > Looks like tcgc doesn't properly propagate the optional value for these params. e.g. for `optionalInt` I see it set to `optional: false`.

If that's true then we should NOT take the PR until that is fixed.

jhendrixMSFT: No need to gate on this as it's not a regression. In addition, we don't properly handle this in the code generator either, so a fix is required on both sides.

jhendrixMSFT (approved):

packages/typespec-go/test/cadlranch/payload/mediatypegroup/zz_responses.go
```diff
@@ -6,12 +6,14 @@ package mediatypegroup
 
 // MediaTypeStringBodyClientGetAsJSONResponse contains the response from method MediaTypeStringBodyClient.GetAsJSON.
 type MediaTypeStringBodyClientGetAsJSONResponse struct {
-	Value *string
+	ContentType *string
```
jhendrixMSFT, line 9: I just noticed this. Why is this now showing up?

packages/typespec-go/test/cadlranch/azure/special-headers/xmsclientreqidgroup/zz_xmsclientrequestid_client.go
tadelesh: seems from customer ask: https://github.com/microsoft/typespec/issues/3350

packages/typespec-go/test/cadlranch/parameters/spreadgroup/zz_spreadalias_client.go
tadelesh: from my debug, the optional is true in tcgc model, am i miss sth?

packages/typespec-go/test/cadlranch/payload/mediatypegroup/zz_responses.go
tadelesh: not quite sure what this change come from? the content type header is a constant. what should go generate?

jhendrixMSFT: Including the content type header for non-binary responses doesn't add any value. If tcgc is modeling this as a constant then it will be easy enough to filter out. It's just surprising that this showed up now and not months ago.

jhendrixMSFT: We can't just skip it when it's a constant as that removes it from cases like binary responses where it might be useful.

Taking a step back, do we know what change is causing this to show up?

tadelesh: i revisited the tcgc code. this should be by design, since use do define this head in the spec. in previous tcgc, there should be a bug that tsp compiler moved this header into the response's content type property, not put it in the parameters, the latest tcgc fix it.

jhendrixMSFT: Got it, thanks for following up.

PR #1408, opened 2024-07-23
remove useless dev dep

tadelesh:
with latest 0.34.9 of cadl-ranch-specs, the cadl-ranch-api dep has upgraded to 0.4.4, the root cause is a wrong auto-release from empty changelog.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -230,15 +230,12 @@ importers:
       '@azure-tools/cadl-ranch':
         specifier: ~0.13.4
         version: 0.13.4(@typespec/versioning@0.58.0)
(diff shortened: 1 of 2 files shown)


jhendrixMSFT (approved):

jhendrixMSFT:
FYI I had left the changelog empty as the XML work was still in progress. I have another PR I'll post shortly that adds more tests.

tadelesh:
> FYI I had left the changelog empty as the XML work was still in progress. I have another PR I'll post shortly that adds more tests.
got it, thanks for the info. it happened when we try to release a new version for new spread test. timothee said it should be a tooling issue and he will fix it after moving cadl-ranch repo to typespec repo.

PR #1410, opened 2024-07-24
Remove wire format from model types

jhendrixMSFT:
The wire format is a function of the operation and isn't tied to a model. It's also possible for a model to be used with more than one wire format, although that's not a real scenario at present. When determining which SerDe methods to generate, find the operation that uses the model type to determine the wire format.

Precursor for https://github.com/Azure/autorest.go/issues/1300

### DIFF
--- a/packages/autorest.go/src/m4togocodemodel/clients.ts
+++ b/packages/autorest.go/src/m4togocodemodel/clients.ts
@@ -261,7 +261,11 @@ function adaptResponseEnvelope(m4CodeModel: m4.CodeModel, codeModel: go.CodeMode
     if (!modelType) {
       throw new Error(`didn't find type name ${resultProp.schema.language.go!.name} for response envelope ${respEnv.name}`);
     }
-    respEnv.result = new go.ModelResult(modelType, adaptResultFormat(helpers.getSchemaResponse(op)!.protocol));
+    const resultFormat = adaptResultFormat(helpers.getSchemaResponse(op)!.protocol);
+    if (resultFormat !== 'JSON' && resultFormat !== 'XML') {
+      throw new Error(`unexpected result format ${resultFormat} for model ${modelType.name}`);
+    }
+    respEnv.result = new go.ModelResult(modelType, resultFormat);
   } else {
     throw new Error(`unhandled result type for operation ${op.language.go!.name}`);
   }
--- a/packages/codegen.go/src/helpers.ts
+++ b/packages/codegen.go/src/helpers.ts
@@ -583,3 +583,84 @@ export function recursiveUnwrapMapSlice(item: go.PossibleType): go.PossibleType 
 export function star(param: go.Parameter): string {
   return go.isRequiredParameter(param) || param.byValue ? '' : '*';
 }
+
+export type SerDeFormat = 'JSON' | 'XML';
+
+// used by getSerDeFormat to cache results
+const serDeFormatCache = new Map<string, SerDeFormat>();
+
+// returns the wire format for the named model.
+// at present this assumes the formats to be mutually exclusive.
+export function getSerDeFormat(model: go.ModelType | go.PolymorphicType, codeModel: go.CodeModel): SerDeFormat {
+  let serDeFormat = serDeFormatCache.get(model.name);
+  if (serDeFormat) {
+    return serDeFormat;
+  }
+
+  // for model-only builds we assume the format to be JSON
+  if (codeModel.clients.length === 0) {
+    return 'JSON';
+  }
+
+  // recursively walks the fields in model, updating serDeFormatCache with the model name and specified format
+  const recursiveWalkModelFields = function(type: go.PossibleType, serDeFormat: SerDeFormat): void {
+    type = recursiveUnwrapMapSlice(type);
+    if (go.isInterfaceType(type)) {
+      recursiveWalkModelFields(type.rootType, serDeFormat);
+      for (const possibleType of type.possibleTypes) {
+        recursiveWalkModelFields(possibleType, serDeFormat);
+      }
+    } else if (go.isPolymorphicType(type) || go.isModelType(type)) {
+      if (serDeFormatCache.has(type.name)) {
+        // we've already processed this type, don't do it again
+        return;
+      }
+
+      serDeFormatCache.set(type.name, serDeFormat);
+      for (const field of type.fields) {
+        const fieldType = recursiveUnwrapMapSlice(field.type);
+        recursiveWalkModelFields(fieldType, serDeFormat);
+      }
+    }
+  };
+
+  // walk the methods, indexing the model formats
+  for (const client of codeModel.clients) {
+    for (const method of client.methods) {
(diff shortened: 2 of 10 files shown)


packages/codegen.go/src/helpers.ts
```diff
@@ -583,3 +583,84 @@ export function recursiveUnwrapMapSlice(item: go.PossibleType): go.PossibleType
...
+// returns the wire format for the named model.
+// at present this assumes the formats to be mutually exclusive.
+export function getSerDeFormat(model: go.ModelType | go.PolymorphicType, codeModel: go.CodeModel): SerDeFormat {
+  let serDeFormat = serDeFormatCache.get(model.name);
+  if (serDeFormat) {
+    return serDeFormat;
+  }
+
+  // for model-only builds we assume the format to be JSON
+  if (codeModel.clients.length === 0) {
```
jhendrixMSFT, line 601: I don't think model-only builds for XML is a real thing but am following up.

packages/autorest.go/src/m4togocodemodel/clients.ts
```diff
@@ -261,7 +261,11 @@ function adaptResponseEnvelope(m4CodeModel: m4.CodeModel, codeModel: go.CodeMode
     if (!modelType) {
       throw new Error(`didn't find type name ${resultProp.schema.language.go!.name} for response envelope ${respEnv.name}`);
     }
-    respEnv.result = new go.ModelResult(modelType, adaptResultFormat(helpers.getSchemaResponse(op)!.protocol));
+    const resultFormat = adaptResultFormat(helpers.getSchemaResponse(op)!.protocol);
```
RickWinter, line 264: Do you want to call it serdeFormat here to be consistent throughout the code. We refer to it as SerDeFormat elsewhere in the code

packages/codegen.go/src/helpers.ts
RickWinter: Should we throw if we do see them in XML?

packages/codegen.go/src/helpers.ts
```diff
@@ -583,3 +583,84 @@ export function recursiveUnwrapMapSlice(item: go.PossibleType): go.PossibleType
...
+        } else if (go.isMonomorphicResult(resultType) && (resultType.format === 'JSON' || resultType.format === 'XML')) {
+          recursiveWalkModelFields(resultType.monomorphicType, resultType.format);
+        }
+      }
+    }
+  }
+
+  serDeFormat = serDeFormatCache.get(model.name);
+  if (!serDeFormat) {
+    // if we get here there are two possibilities
```
RickWinter, line 657: Should we display a warning or will that be too noisy b/c of unreferenced types?

packages/codemodel.go/src/result.ts
```diff
@@ -97,10 +97,13 @@ export interface PolymorphicResult {
 
   interfaceType: type.InterfaceType;
 
-  // the format in which the result is returned
+  // the format in which the result is returned.
+  // only JSON is supported for polymorphic types.
   format: 'JSON';
 }
 
+export type ModelResultFormat = 'JSON' | 'XML';
```
RickWinter, line 105: Can this share the same SerDeFormat enum?

packages/codegen.go/src/helpers.ts
jhendrixMSFT: In swagger, one can define `consumes` and `produces` to define content types at the global level. For tsp, I don't believe there's such a construct; you can only define the content type via operations. This means that for model-only builds in tsp, we have no way to describe the wire format. So, there would be no way to detect this (see my email thread for more info on this).

packages/codegen.go/src/helpers.ts
```diff
@@ -583,3 +583,84 @@ export function recursiveUnwrapMapSlice(item: go.PossibleType): go.PossibleType
...
+  const recursiveWalkModelFields = function(type: go.PossibleType, serDeFormat: SerDeFormat): void {
+    type = recursiveUnwrapMapSlice(type);
+    if (go.isInterfaceType(type)) {
+      recursiveWalkModelFields(type.rootType, serDeFormat);
+      for (const possibleType of type.possibleTypes) {
+        recursiveWalkModelFields(possibleType, serDeFormat);
+      }
+    } else if (go.isPolymorphicType(type) || go.isModelType(type)) {
+      if (serDeFormatCache.has(type.name)) {
+        // we've already processed this type, don't do it again
```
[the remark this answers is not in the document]
jhendrixMSFT: Technically, a model can be sent/received in both formats as it's a function of the operation. From my internal conversations, it's not a scenario we need to support so I left them as mutually exclusive.

packages/codegen.go/src/models.ts
```diff
@@ -176,12 +176,13 @@ function generateModelDefs(modelImports: ImportManager, serdeImports: ImportMana
...
 
-    const modelDef = new ModelDef(model.name, model.format, model.fields, model.description);
+    const serDeFormat = getSerDeFormat(model, codeModel);
+    const modelDef = new ModelDef(model.name, serDeFormat, model.fields, model.description);
     for (const field of values(modelDef.Fields)) {
       modelImports.addImportForType(field.type);
     }
 
-    if (go.isModelType(model) && model.format === 'xml' && !model.annotations.omitSerDeMethods) {
+    if (go.isModelType(model) && serDeFormat === 'XML' && !model.annotations.omitSerDeMethods) {
```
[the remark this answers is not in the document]
jhendrixMSFT: Leaving as union per offline conversation.

RickWinter (approved):

packages/codegen.go/src/helpers.ts
```diff
@@ -583,3 +583,84 @@ export function recursiveUnwrapMapSlice(item: go.PossibleType): go.PossibleType
...
+        recursiveWalkModelFields(fieldType, serDeFormat);
+      }
+    }
+  };
+
+  // walk the methods, indexing the model formats
+  for (const client of codeModel.clients) {
+    for (const method of client.methods) {
+      for (const param of method.parameters) {
+        if (!go.isBodyParameter(param) || (param.bodyFormat !== 'JSON' && param.bodyFormat !== 'XML')) {
```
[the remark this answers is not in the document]
jhendrixMSFT: Yes, could be `text/plain`, `application/octet-stream` etc. In those cases we just skip them.

chlowell (approved):

packages/codegen.go/src/helpers.ts
jhendrixMSFT: I don't think it would add much value and since we fall back to doing the right thing for the 99% case it would just be noise.

packages/autorest.go/src/m4togocodemodel/clients.ts
jhendrixMSFT: The resultant type is `go.ResultFormat` which is a superset of `SerDeFormat` (includes `Text`).

PR #1413, opened 2024-07-29
Update azcore version

chlowell:
We need to ship this ASAP to prevent Component Governance flagging new libraries for an indirect CVE.


jhendrixMSFT:
You'll need to `rush regenerate` and `rush tspcompile` to regenerate the world. You might also need to `rush modtidy` after.

chlowell:
Should `rush regenerate` also regenerate the baseline for `autorest.gotest`?

jhendrixMSFT:
No, and it looks like those mock modules use very old versions. Might be best to revert that part of the change and let @tadelesh update autorest.gotest separately as I don't understand all the nuance of making changes there.

jhendrixMSFT:
Can you also update the changelog for the tsp emitter to include a note about updating the minimum version of `azcore`? You might need to rebase after my recent merge.

jhendrixMSFT (approved):

PR #1416, opened 2024-08-01
Remove testserver codecoverage publishing

weshaggard:


eng/pipelines/templates/steps/build-test-go.yaml
```diff
@@ -129,12 +128,3 @@ steps:
       testRunner: JUnit
       testResultsFiles: ${{ parameters.GoTestPath }}/**/report.xml
       failTaskOnFailedTests: true
-
-  - ${{if and(ne(variables['Build.Reason'], 'PullRequest'), eq(parameters.PublishAutorestCoverage, true))}}:
```
jhendrixMSFT, line 133: Is this no longer supported?

weshaggard: Correct do to security we cannot use SAS tokens any longer and instead of updating the tooling we decided to stop publishing coverage.

RickWinter: > Correct do to security we cannot use SAS tokens any longer and instead of updating the tooling we decided to stop publishing coverage.

How are we tracking the CADL code coverage now? What is the plan to bring back support for this?
//cc @lmazuel, @allenjzhang

weshaggard: The cadl ranch code coverage is still happening this is only the autorest testserver coverage.

jhendrixMSFT (approved):

eng/pipelines/templates/steps/build-test-go.yaml
RickWinter: What's the plan for bringing this coverage tracking back? Is there an issue tracking that work?

weshaggard: No we are retiring the publishing of the coverage reports for autorest. It still exists with cadl-ranch for the newer typespec code gen.

PR #1425, opened 2024-08-14
Don't prune base models annotated as output types

jhendrixMSFT:
If the type has been marked as output then preserve it.


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -878,7 +878,9 @@ export class typeAdapter {
         // exclude any polymorphic root type from the check
         // as we always need to include the root type even
         // if it's not referenced.
-        if (!parent.discriminatedSubtypes && !baseModels.has(parent.name)) {
+        // also exclude types that have been explicitly annotated
+        // as output types.
+        if (!parent.discriminatedSubtypes && (model.usage & tsp.UsageFlags.Output) !== tsp.UsageFlags.Output && !baseModels.has(parent.name)) {
```
RickWinter: Nit: This was confusing to read. Maybe split the if onto multiple lines? Or put the middle condition into a wrapped paran to make it more visible that `(model.usage & tsp.UsageFlags.Output) !== tsp.UsageFlags.Output` is a condition.

RickWinter (approved):

PR #1427, opened 2024-08-22
adopt typespec and tcgc new version

tadelesh:
three main changes in this PR:
1. we will use relative path from tcgc in example comment
2. tcgc has a breaking change for endpoint: https://github.com/Azure/typespec-azure/pull/1155. currently, endpoint will always be overridable, so there will be multiple endpoints. i just pick the first one as a workaround. i don't know whether we should support it.
3. add `explode` and `allowReserved` support for uri template, which resolve: https://github.com/Azure/autorest.go/issues/1423, https://github.com/Azure/autorest.go/issues/1422


packages/typespec-go/test/cadlranch/authentication/http/customgroup/zz_custom_client.go
```diff
@@ -44,6 +46,8 @@ func (client *CustomClient) Invalid(ctx context.Context, options *CustomClientIn
 
 // invalidCreateRequest creates the Invalid request.
 func (client *CustomClient) invalidCreateRequest(ctx context.Context, _ *CustomClientInvalidOptions) (*policy.Request, error) {
+	host := "{endpoint}"
+	host = strings.ReplaceAll(host, "{endpoint}", client.endpoint)
```
jhendrixMSFT, line 50: We should skip the replacement step when the endpoint is singular.

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -90,23 +90,32 @@ export class clientAdapter {
...
+            paramType = param.type.values[0];
+          }
           // this will either be a fixed or templated host
           // don't set the fixed host for ARM as it isn't used
           if (this.ta.codeModel.type !== 'azure-arm') {
-            goClient.host = param.type.serverUrl;
+            goClient.host = paramType.serverUrl;
           }
-          if (param.type.templateArguments.length === 0) {
+          if (paramType.templateArguments.length === 0) {
```
jhendrixMSFT, line 106: I've been thinking about this and I'd like to remove the host template and require the full URL just like we do in Rust. See https://github.com/Azure/autorest.rust/blob/main/packages/typespec-rust/src/tcgcadapter/adapter.ts#L392-L395

RickWinter (approved):

packages/typespec-go/test/cadlranch/authentication/http/customgroup/zz_custom_client.go
tadelesh: from tcgc result, the current `endpoint` is sth like:
```
@server(
  "{endpoint}",
  "Testserver endpoint",
  {
    endpoint: url = "http://localhost:3000",
  }
)
```
do you mean, we should fall back to constant for this case?

jhendrixMSFT: To take a step back, the original design of track 2 clients is that their constructors take an endpoint, a credential if authentication is required, and an options type. For Go, since we don't emit constructors for data-plane at present, we decided to emit the fixed endpoint as a `const` since it's not configurable. This is an exceptional case as data-plane services require and endpoint and ARM uses the `cloud.Configuration` to determine the endpoint.

I read through the issues and if I correctly understand them, the goal is to provide an option to specify an endpoint even when it's fixed. This essentially means that the endpoint has a client-side default value, effectively pushing the endpoint to the client options.

If/when we decide to emit client constructors for data-plane, I think we'll want to make the endpoint parameter mandatory, never pushing it to client options even if there's a client-side default value. This means that the clients will have an `endpoint` field, and we'd directly consume that in the method bodies.

```go
req, err := runtime.NewRequest(ctx, http.MethodGet, runtime.JoinPaths(client.endpoint, urlPath))
```

tadelesh: Let me split into two parts:
For ARM, we use `cloud.Configuration` to determine the endpoint. So, whatever tcgc give the emitter for the endpoint parameter, we always ignore it, right?
For data plan, we do not generate the ctor at present, how should we deal with the following three cases:
1. User defines a fixed endpoint, tcgc will give emitter an optional endpoint parameter with client default value.
According to your reply, it seems we should put it into client options, right?
2. User defines sth like `{endpoint}`, tcgc will give emitter a required endpoint parameter.
For this case, it should be a client property like the code you referred, right?
3. User defines sth like `https://api.{host}.com`, tcgc will give emitter a union type, one variant is a requirement endpoint parameter like 2, another is required endpoint with `host` templated argument.
How should we deal with this?
4. User defines several different endpoint, tcgc will give emitter a union type with variants of above.
How should we deal with multiple endpoint definition?

jhendrixMSFT: For ARM you're correct, we do nothing (i.e. just what we're doing at present).

For Azure data-plane, I don't think we want to support these scenarios verbatim, and just always require the complete endpoint as the first constructor param.

For unbranded data-plane, we will need to decide how we want to surface this.

Until we have this sorted, it might be best to leave the codegen as it was (i.e. don't change any emitted code with this PR).

CC @JeffreyRichter

tadelesh: got it. refined the logic, please help to review.

jhendrixMSFT (approved):

PR #1428, opened 2024-08-28
support int as string

tadelesh:
this should be added to all numeric types, but this semester we only need to support int/uint from service ask.

resolve: https://github.com/Azure/autorest.go/issues/1402


packages/codemodel.go/src/type.ts
```diff
@@ -130,6 +130,8 @@ export type PrimitiveTypeName = 'any' | 'bool' | 'byte' | 'float32' | 'float64'
 
 export type BytesEncoding = 'Std' | 'URL';
 
+export type PrimitiveEncoding = 'string';
```
jhendrixMSFT, line 133: Looks like this isn't used?

packages/codemodel.go/src/type.ts
```diff
@@ -138,6 +140,7 @@ export interface BytesType {
 // PrimitiveType is a Go integral type
 export interface PrimitiveType {
   typeName: PrimitiveTypeName;
+  encodeAsString?: boolean;
```
jhendrixMSFT, line 143: Can you remove the `?` as I believe that `false` and `undefined` have the same meaning.

packages/codegen.go/src/models.ts
```diff
@@ -372,7 +372,17 @@ function generateJSONMarshallerBody(modelType: go.ModelType | go.PolymorphicType
       } else {
         modelDef.SerDe.needsJSONPopulate = true;
       }
-      marshaller += `\t${populate}(objectMap, "${field.serializedName}", ${receiver}.${field.name})\n`;
+      if (go.isPrimitiveType(field.type) && (field.type.typeName.startsWith('uint') || field.type.typeName.startsWith('int')) && field.type.encodeAsString) {
```
jhendrixMSFT, line 375: Looks like array of these is missing in both cases. If there is no test coverage for that it's ok to omit for now but will need to be addressed at some point. What services are asking for this?

packages/typespec-go/test/cadlranch/payload/multipartgroup/formdata_client_test.go
```diff
@@ -102,43 +102,12 @@ func TestFormDataClient_Complex(t *testing.T) {
...
-				City: to.Ptr("Z"),
-			},
-		},
 		ProfileImage: streaming.MultipartContent{Body: jpgFile},
 	}, nil)
 	require.NoError(t, err)
 	require.Zero(t, resp)
 }
 
-func TestFormDataClient_JSONArrayParts(t *testing.T) {
```
jhendrixMSFT, line 119: Why was this removed?

tadelesh: This tests have been removed from cadl-ranch new version and some other tests added. I could open another pr to add all added tests.

packages/codegen.go/src/models.ts
tadelesh: Batch is asking for this. It seems I also miss the map, let me check the logic.

jhendrixMSFT (approved):

PR #1430, opened 2024-09-18
Throw an error if no media types are defined

jhendrixMSFT:
The message includes the operation ID to help with debugging.

### DIFF
--- a/packages/autorest.go/src/m4togocodemodel/clients.ts
+++ b/packages/autorest.go/src/m4togocodemodel/clients.ts
@@ -315,6 +315,9 @@ function adaptMethodParameter(op: m4.Operation, param: m4.Parameter): go.Paramet
 
   switch (param.protocol.http?.in) {
     case 'body': {
+      if (!op.requests![0].protocol.http!.mediaTypes) {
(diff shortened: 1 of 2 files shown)


packages/autorest.go/src/m4togocodemodel/clients.ts
```diff
@@ -315,6 +315,9 @@ function adaptMethodParameter(op: m4.Operation, param: m4.Parameter): go.Paramet
 
   switch (param.protocol.http?.in) {
     case 'body': {
+      if (!op.requests![0].protocol.http!.mediaTypes) {
+        throw new Error(`no media types defined for operation ${op.operationId}`);
+      }
       let contentType = `"${op.requests![0].protocol.http!.mediaTypes[0]}"`;
       if (op.requests![0].protocol.http!.mediaTypes.length > 1) {
```
chlowell, line 322: Do we still need this condition?

chlowell (approved):

packages/autorest.go/src/m4togocodemodel/clients.ts
jhendrixMSFT: Yeah, that's to handle multiple media types.

PR #1432, opened 2024-09-19
Fake servers honor caller's context in *http.Request

jhendrixMSFT:
The fake dispatch is handled on a separate goroutine so that we can select on the caller's context and the dispatch result.

Fixes https://github.com/Azure/azure-sdk-for-go/issues/23470


packages/autorest.go/test/acr/azacr/fake/zz_authentication_server.go
```diff
@@ -58,21 +58,30 @@ func (a *AuthenticationServerTransport) Do(req *http.Request) (*http.Response, e
...
+			res.err = fmt.Errorf("unhandled API %s", method)
+		}
 
-	return resp, err
+		resultChan <- res
+	}()
+
+	select {
+	case <-req.Context().Done():
+		return nil, req.Context().Err()
```
jhendrixMSFT, line 81: This will leak the dispatch goroutine.

jhendrixMSFT: Fixed.

chlowell (approved):

PR #1449, opened 2024-10-22
Add support for XML in tsp

jhendrixMSFT:
Adapt various tsp decorators that contain XML info including attributes, names, and unwrapped arrays.
Added support for marshaling maps in XML format.
Moved XMLInfo out of Parameter and into body param types.

Fixes https://github.com/Azure/autorest.go/issues/1300


packages/autorest.go/test/autorest/xmlgroup/zz_xml_helper.go
```diff
@@ -15,6 +15,40 @@ import (
 
 type additionalProperties map[string]*string
 
+// MarshalXML implements the xml.Marshaler interface for additionalProperties.
+func (ap additionalProperties) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
```
jhendrixMSFT, line 19: Historically we never needed this. The impl was very storage specific which required only unmarshaling support.

chlowell (approved):

Issue #1441, opened 2024-10-17
tsp codegen for examples is broken with latest tcgc


jhendrixMSFT:
For reference https://github.com/Azure/typespec-azure/issues/1688

Also observed was some of the ARM modules have missing/incorrect parameters.

jhendrixMSFT:
Also impacted is `armloadtestservice`.
```
vet.exe: .\zz_loadtests_client_example_test.go:25:105: not enough arguments in call to clientFactory.NewLoadTestsClient().BeginCreateOrUpdate
        have (context.Context, string, string, nil)
        want (context.Context, string, string, armloadtestservice.LoadTestResource, *armloadtestservice.LoadTestsClientBeginCreateOrUpdateOptions)
```

jhendrixMSFT:
The api-version issue has been fixed.

tadelesh:
confirmed that all the example failure is from setting body parameter name with `@encodedName`. open tcgc issue [here](https://github.com/Azure/typespec-azure/issues/1739).

PR #1453, opened 2024-10-23
Skip api-version param when emitting examples

jhendrixMSFT:
Partial fix for https://github.com/Azure/autorest.go/issues/1441


packages/typespec-go/.scripts/tspcompile.js
```diff
@@ -129,13 +129,13 @@ const armapicenter = pkgRoot +  'test/tsp/ApiCenter.Management';
 generate('armapicenter', armapicenter, 'test/armapicenter');
 
 const armlargeinstance = pkgRoot + 'test/tsp/AzureLargeInstance.Management';
-generate('armlargeinstance', armlargeinstance, 'test/armlargeinstance', ['stutter=AzureLargeInstance']);
+generate('armlargeinstance', armlargeinstance, 'test/armlargeinstance', ['stutter=AzureLargeInstance', 'generate-examples=false']);
```
jhendrixMSFT, line 132: This and `armloadtestservice` are still broken due to missing params (looks to be a tcgc bug).

RickWinter: > This and `armloadtestservice` are still broken due to missing params (looks to be a tcgc bug).

//cc  @izzymsft

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -710,6 +710,10 @@ export class clientAdapter {
       for (const example of sdkMethod.operation.examples) {
         const goExample = new go.MethodExample(example.name, {summary: example.description}, example.filePath);
         for (const param of example.parameters) {
+          if (param.parameter.isApiVersionParam && param.parameter.clientDefaultValue) {
+            // skip the api-version param as it's not a formal parameter
+            continue;
+          }
```
tadelesh, line 716: i'm wondering why the subscription id is not impacted, it should also be a client param

Issue #1444, opened 2024-10-18
Optional body params should conditionally set Content-Type header


jhendrixMSFT:
tcgc change https://github.com/Azure/typespec-azure/issues/1259 makes `content-type` literal param optional if the body is option. This has the side effect of surfacing this as an optional flag in the emitted code. Need to fix this such that if the optional body has been set, then we also set `content-type` and not surface it as an optional flag param.

PR #1456, opened 2024-10-24
Conditionally set Content-Type header

jhendrixMSFT:
When the body param is optional, the Content-Type header should be set IFF the body param is not nil.
Don't expose optional Content-Type literal params as flags.

Fixes https://github.com/Azure/autorest.go/issues/1444


jhendrixMSFT:
@benbp part of the PR has been consistently failing with error "fatal: couldn't find remote ref refs/pull/1456/merge". Have you seen this before?

chlowell (approved):

benbp:
@jhendrixMSFT not sure, though looks like you did get a passing CI run

PR #1457, opened 2024-10-29
fix codegen issues found during current month's release

tadelesh:
1. fix wrong condition for discriminator value comparison
2. fix wrong condition of encoding for path parameter
3. add example literal value handling
4. remove useless comma in example client initialization


packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -490,7 +490,7 @@ export class clientAdapter {
         adaptedParam = new go.HeaderParameter(paramName, param.serializedName, this.adaptHeaderType(param.type, true), paramKind, byVal, location);
       }
     } else if (param.kind === 'path') {
-      adaptedParam = new go.PathParameter(paramName, param.serializedName, param.urlEncode || !param.allowReserved, this.adaptPathParameterType(param.type), paramKind, byVal, location);
+      adaptedParam = new go.PathParameter(paramName, param.serializedName, param.urlEncode && !param.allowReserved, this.adaptPathParameterType(param.type), paramKind, byVal, location);
```
jhendrixMSFT, line 493: It appears that `urlEncode` has been deprecated. Can this just be removed?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -819,12 +819,15 @@ export class clientAdapter {
...
           let concreteType: go.ModelType | go.PolymorphicType;
           if (go.isInterfaceType(goType)) {
-            concreteType = goType.possibleTypes.find(t => t.discriminatorValue?.literal === exampleType.type.discriminatorValue)!;
+            concreteType = goType.possibleTypes.find(t => t.discriminatorValue?.literal === exampleType.type.discriminatorValue || t.discriminatorValue?.literal.value === exampleType.type.discriminatorValue)!;
           } else {
             concreteType = goType;
           }
           const ret = new go.StructExample(concreteType);
           for (const [k, v] of Object.entries(exampleType.value)) {
+            if (concreteType === undefined) {
```
jhendrixMSFT, line 828: Can you please change the definition of `concreteType` to include `undefined` and move this check to before the call to `new go.StructExample(...)`.

Issue #1443, opened 2024-10-18
Fix pruning of unused base types


jhendrixMSFT:
tcgc change https://github.com/Azure/typespec-azure/pull/1659 propagates access and usage flags to base types IFF the base type doesn't have these flags set. This means that for tests like https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/test/tsp/ModelsOnlyWithBaseTypes/main.tsp#L27 where we use usage to determine to omit a base type. While we could fix this test by adding the necessary flags, I suspect that the algorithm will need to change to ensure it works as expected across all tsps.

PR #1463, opened 2024-11-07
Fix removing unreferenced types

jhendrixMSFT:
The remove-unreferenced-types switch has been removed as tcgc does this by default. All we need to handle are Go-specific scenarios. The candidates for type omission include base model types, error types, and Azure.Core types. Any of the type candidates that are directly or indirectly referenced will be preserved.

Fixes https://github.com/Azure/autorest.go/issues/1443


packages/typespec-go/test/armdatabasewatcher/zz_models.go
```diff
@@ -78,42 +78,6 @@ type Datastore struct {
 	KustoClusterDisplayName *string
 }
 
-// ErrorAdditionalInfo - The resource management error additional info.
-type ErrorAdditionalInfo struct {
```
jhendrixMSFT, line 82: This, and other types, were inadvertently resurrected in the last big toolset upgrade due to breaking changes in tcgc. This has now been rectified.

packages/typespec-go/test/cadlranch/azure/client-generator-core/coreusagegroup/zz_models.go
```diff
@@ -10,6 +10,12 @@ type InputModel struct {
 	Name *string
 }
 
+// OrphanModel - Not used anywhere, but access is override to public so still need to be generated and exported with serialization.
+type OrphanModel struct {
```
jhendrixMSFT, line 14: And apparently we were too aggressive in pruning types before.

packages/typespec-go/test/azmodelsonly/zz_models.go
```diff
@@ -24,8 +19,3 @@ type DerivedTwo struct {
 	// REQUIRED
 	Name *string
 }
-
-type PreservedBaseClass struct {
```
jhendrixMSFT, line 28: We need a fix in tcgc to resolve this. Tracked by https://github.com/Azure/typespec-azure/issues/1827

PR #1472, opened 2024-12-03
Update emitter to latest tcgc

jhendrixMSFT:
This fixes an issue where query head/path/query params have different wire names from the parameter names, causing the params to be omitted. Added KV keys to test this scenario.


jhendrixMSFT:
@microsoft-github-policy-service rerun

RickWinter (approved):
Looks Good.

PR #1474, opened 2024-12-10
upgrade tcgc version and add usage filter

tadelesh:
since go does not need initial response and polling response for lro, we filtered them out.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -32,8 +32,8 @@ export class typeAdapter {
       if (enumType.usage === tcgc.UsageFlags.ApiVersionEnum) {
         // we have a pipeline policy for controlling the api-version
         continue;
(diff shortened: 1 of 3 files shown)


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -111,6 +111,23 @@ export class typeAdapter {
...
   }
 
+  private skipSpecificTypes(type: tcgc.SdkModelType | tcgc.SdkEnumType): boolean {
+    if (tcgc.isAzureCoreModel(type)) {
+      // skip core types
+      return true;
+    } else if ((type.usage & tcgc.UsageFlags.Exception) === tcgc.UsageFlags.Exception && (type.usage & tcgc.UsageFlags.Input) === 0 && (type.usage & tcgc.UsageFlags.Output) === 0) {
+      // skip error type
+      return true;
+    } else if ((type.usage & tcgc.UsageFlags.LroPolling) === tcgc.UsageFlags.LroPolling && (type.usage & tcgc.UsageFlags.Input) === 0 && (type.usage & tcgc.UsageFlags.Output) === 0) {
```
jhendrixMSFT, line 121: We will need these at some point, but at least it's all centralized.

PR #1495, opened 2025-02-05
add special handling for `@bodyRoot`/`@body` on model param's property

tadelesh:
this problem was found in this pr: https://github.com/Azure/azure-rest-api-specs/pull/31699. there is an action [operation](https://github.com/Azure/azure-rest-api-specs/blob/79d87fa78471a02293b73433f53a28666498ded7/specification/hybridconnectivity/HybridConnectivity.Management/EndpointResource.tsp#L117C3-L117C33): `listIngressGatewayCredentials` for `EndpointResource` resource which used `@bodyRoot` in a model's property.
previous go emitter's logic of finding parameter's corresponding http parameter is to use the direct naming comparison of the param, but this could not cover `@bodyRoot` or `@body` on model's property, which cause this [exception](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4505829&view=logs&j=0303ead4-2e3b-557b-303e-fe8735d4d18f&t=b29b24f0-8be9-5e66-edce-54fd3565c09c).
this pr just adds logic of handling first level `@bodyRoot` and `@body`. however, it is not a perfect solution, we need to refine param converting logic after we have a clear story about the relation of typespec operation api sig and sdk operation api sig.


packages/typespec-go/test/local/armtest/go.mod
jhendrixMSFT, line 1: Need to execute `go mod tidy`.

packages/typespec-go/test/tsp/Test.Management/BodyRoot.tsp
```diff
@@ -0,0 +1,37 @@
...
+@armResourceOperations
+interface BodyRoots {
+  @doc("Get details of a certificate profile.")
+  get is ArmResourceRead<BodyRoot>;
+
+  @doc("Revoke a certificate under a certificate profile.")
+  action is ArmResourceActionNoContentSync<
+    BodyRoot,
+    {
+      @bodyRoot
```
jhendrixMSFT, line 33: Where can I find the docs for this decorator?

tadelesh: https://typespec.io/docs/libraries/http/reference/decorators/

tadelesh:
@jhendrixMSFT could you review again? this bug block the hybrid connectivity's pr.

jhendrixMSFT:
Can you please update the changelog with this fix and the `package.json` file with the updated version?

packages/typespec-go/test/local/armtest/zz_bodyroots_client.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	urlPath = strings.ReplaceAll(urlPath, "{subscriptionId}", url.PathEscape(client.subscriptionID))
+	if resourceGroupName == "" {
+		return nil, errors.New("parameter resourceGroupName cannot be empty")
+	}
+	urlPath = strings.ReplaceAll(urlPath, "{resourceGroupName}", url.PathEscape(resourceGroupName))
+	if bodyRootName == "" {
+		return nil, errors.New("parameter bodyRootName cannot be empty")
+	}
+	urlPath = strings.ReplaceAll(urlPath, "{bodyRootName}", url.PathEscape(bodyRootName))
```
jhendrixMSFT, line 85: Is this generated code actually correct? I've read the docs for the `@bodyRoot` decorator several times now and it's just not clear to me what it's supposed to do. In this context, it looks like it's simply a path parameter.

tadelesh: this parameter is from this [param](https://github.com/Azure/autorest.go/blob/b8b9881de2b3b094320654e84974486fe91dad79/packages/typespec-go/test/tsp/Test.Management/BodyRoot.tsp#L16-L21). action of `ActionRequest` type is the request body.

tadelesh:
> Can you please update the changelog with this fix and the `package.json` file with the updated version?

added.

packages/typespec-go/test/local/armtest/zz_bodyroots_client.go
tadelesh: from my understanding, if a model that has a property that is used in path, we could use `@bodyRoot` to simplify the method's signature. something like:
```
model Resource {
  @path
  id: string,
  name: string,
  description: string,
}

op create(@bodyRoot resource: Resource): Resource;
```
you will get one http path parameter `id` and a body parameter of model `Resource`.

jhendrixMSFT: In this case, removing the `@bodyRoot` decorator has minimal impact on the method signature (it just changes the param name). However, it does introduce _two_ models named `ActionRequest` in `sdkContext.sdkPackage.models` which is broken.

jhendrixMSFT: Ok I think I get what this decorator does. It allows mixing of body/header/path/query params in the same model (please correct me if I'm wrong).

It seems there are some bugs here though. e.g. if I update action as follows.
```
  action is ArmResourceActionNoContentSync<
    BodyRoot,
    {
      @bodyRoot
      action: ActionRequest;

      @header
      theHeader: string;

      @query
      theQuery: int32;
    }
  >;
```
I don't see any parameters for `theHeader` or `theQuery` in `sdkMethod.parameters`. The operation parameters _do_ exist. I also noticed that the `serializedName` for `theHeader` was transforms to `the-header` which seems wrong.

jhendrixMSFT: Hmm well it seems that `@body` is similar, and if I change the above to `@body` I observe the same result.

jhendrixMSFT (approved):

packages/typespec-go/test/local/armtest/zz_bodyroots_client.go
tadelesh: `sdkMethod.parameters` follows the raw typespec definition. for your tsp, there will be one parameter named `body` (defined in the action template, which is also decorated with `@bodyRoot`, see [here](https://github.com/Azure/typespec-azure/blob/7db5549ce9ac281247420889e0cc40306cb39e06/packages/typespec-azure-resource-manager/lib/operations.tsp#L654-L655)) and its type is a model which contains the `theHeader` and `theQuery` property.
image
image
and `sdkMethod.operation` follows the http operation calculated by typespec http library, it will get the inner action as body parameter. also, http library will rename all header parameter with `-` style. tcgc does not do any specific logic here.
image

Issue #1488, opened 2025-01-22
Refine client factory CTOR logic to prevent potential breaking change


tadelesh:
We encountered one potential SDK breaking change in [this](https://github.com/Azure/azure-rest-api-specs/pull/32062) spec change.
It is about client factory CTOR. This PR introduced a new client into the service, and this new client has a new client level parameter compared with previous clients.
Our current logic for client factory is to gather all client level parameters and put them into the client factory CTOR, thus cause the breaking change of adding a new required parameter: `subscriptionID`.
```go
func NewClientFactory(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
...
}
```
My proposal is only gathering client parameters that is used by all clients, and put the extra required parameters into to `ClientFactory.NewXXXClient` method.
```go
func NewClientFactory(credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
...
}

func (c *ClientFactory) NewGroupQuotaSubscriptionAllocationRequestClient(subscriptionID string) *GroupQuotaSubscriptionAllocationRequestClient {
...
}
```

PR #1496, opened 2025-02-13
refine factory CTOR logic

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1488


packages/codemodel.go/src/package.ts
```diff
@@ -74,6 +74,8 @@ export interface Options {
   sliceElementsByval: boolean;
 
   generateExamples: boolean;
+
+  factoryGatherAllParams: boolean;
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: ok, sure~

packages/autorest.go/.scripts/regeneration.js
```diff
@@ -106,10 +106,10 @@ const blobStorage = './swagger/specification/storage/data-plane/Microsoft.BlobSt
 generateFromReadme("azblob", blobStorage, 'package-2021-12', 'test/storage/azblob', '--module-version=0.1.0 --inject-spans');
 
 const network = './swagger/specification/network/resource-manager/readme.md';
-generateFromReadme("armnetwork", network, 'package-2022-09', 'test/network/armnetwork', '--module=armnetwork --module-version=0.1.0 --azure-arm=true --remove-unreferenced-types');
+generateFromReadme("armnetwork", network, 'package-2022-09', 'test/network/armnetwork', '--module=armnetwork --module-version=0.1.0 --azure-arm=true --remove-unreferenced-types --factory-gather-all-params=true');
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: ok, i will set a case of not setting this flag.

tadelesh (approved):

tadelesh:
@jhendrixMSFT could you also help to review?

packages/codemodel.go/src/package.ts
```diff
@@ -74,6 +74,9 @@ export interface Options {
   sliceElementsByval: boolean;
 
   generateExamples: boolean;
+
+  // whether or not to gather all client parameters for the client factory.
+  factoryGatherAllParams: boolean;
```
jhendrixMSFT, line 79: Do we have any case where we'd _not_ want to do this? i.e. shouldn't this always be true?

JiaqiZhang-Dev: Yes, when it's false, the client factory ctor will only gather common client parameters. There are some cases like armconsumption, armnetwork and so on.

packages/autorest.go/test/consumption/armconsumption/zz_client_factory.go
```diff
@@ -13,23 +13,20 @@ import (
...
+	internal *arm.Client
 }
 
 // NewClientFactory creates a new instance of ClientFactory with the specified values.
 // The parameter values will be propagated to any client created from this factory.
-//   - subscriptionID - Azure Subscription ID.
 //   - credential - used to authorize requests. Usually a credential from azidentity.
 //   - options - pass nil to accept the default values.
-func NewClientFactory(subscriptionID string, credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
+func NewClientFactory(credential azcore.TokenCredential, options *arm.ClientOptions) (*ClientFactory, error) {
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: In this service, some clients don't need `subscriptionID`, like LotsClient, MarketplacesClient and OperationsClient, so the subscriptionID will not been remained in client factory ctor.

packages/codegen.go/src/helpers.ts
```diff
@@ -739,3 +739,32 @@ export function getAllClientParameters(codeModel: go.CodeModel): Array<go.Parame
...
 }
+
+// return common client parameters for all the clients
+export function getCommonClientParameters(codeModel: go.CodeModel): Array<go.Parameter> {
+  const commonClientParams = new Array<go.Parameter>();
+  const paramCount = new Map<string, { count: number, param: go.Parameter }>();
+  var clientCnt = 0;
+  for (const clients of codeModel.clients) {
+    // special cases: some clients always don't contain any parameters (OperationsClient will be depracated in the future)
+    if (new RegExp('^OperationsClient$').test(clients.name)) {
```
JiaqiZhang-Dev, line 750: @jhendrixMSFT Is it ok to ignore the client without any parameter like this? Because `OperationsClient` is a special client, it needs some special logic to process it.

JiaqiZhang-Dev:
@jhendrixMSFT I have done a full tests for all current arm packages with the default config for `factoryGatherAllParams` which means client factory only gathers common parameters of client. There are about 20%(46/244) packages has a breaking change caused by `NewClientFactory`, such as armnetwork, armbilling, armfrontdoor and so on. 

In this month release, there are two services have the breaking change in `NewClientFactory` function which bring a new parameter([quota](https://github.com/Azure/azure-sdk-for-go/pull/24035/files), [hybridconnectivity](https://github.com/Azure/azure-sdk-for-go/pull/24136/files)), so that we need to solve this problem asap.

tadelesh:
> @jhendrixMSFT @tadelesh I have done a full tests for all current arm packages with the default config for `factoryGatherAllParams` which means client factory only gathers common parameters of client. There are about 20%(46/244) packages has a breaking change caused by `NewClientFactory`, such as armnetwork, armbilling, armfrontdoor and so on.
> 
> In this month release, there are two services have the breaking change in `NewClientFactory` function which bring a new parameter([quota](https://github.com/Azure/azure-sdk-for-go/pull/24035/files), [hybridconnectivity](https://github.com/Azure/azure-sdk-for-go/pull/24136/files)), so that we need to solve this problem asap.

are they generated from typespec or swagger? can we add the config based on your test which could prevent future breaking when releasing.

JiaqiZhang-Dev:
> > @jhendrixMSFT @tadelesh I have done a full tests for all current arm packages with the default config for `factoryGatherAllParams` which means client factory only gathers common parameters of client. There are about 20%(46/244) packages has a breaking change caused by `NewClientFactory`, such as armnetwork, armbilling, armfrontdoor and so on.
> > In this month release, there are two services have the breaking change in `NewClientFactory` function which bring a new parameter([quota](https://github.com/Azure/azure-sdk-for-go/pull/24035/files), [hybridconnectivity](https://github.com/Azure/azure-sdk-for-go/pull/24136/files)), so that we need to solve this problem asap.
> 
> are they generated from typespec or swagger? can we add the config based on your test which could prevent future breaking when releasing.

By swagger. Good suggestion! I'll add the config for those services.

jhendrixMSFT:
Can you please update the changelog with the info about this change? Also, please add an entry to the readme about the new command line switch.

JiaqiZhang-Dev:
> Can you please update the changelog with the info about this change? Also, please add an entry to the readme about the new command line switch.

done~

packages/codegen.go/src/helpers.ts
jhendrixMSFT: I think it's probably safe. I updated the condition to apply this check only for ARM.

jhendrixMSFT (approved):

PR #1497, opened 2025-02-17
upgrade to typespec 0.65 and typespec-azure 0.51

JiaqiZhang-Dev:
regular work after typespec monthly release.


tadelesh:
there are some deprecation in tcgc 0.51, could you also help to upgrade?

JiaqiZhang-Dev:
> there are some deprecation in tcgc 0.51, could you also help to upgrade?

@tadelesh Sure, i submitted some changes for deprecation. please help review~

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -215,8 +215,10 @@ export class clientAdapter {
       if (lroOptions) {
         (<go.LROMethod>method).finalStateVia = lroOptions['finalState'];
       }
-      if (sdkMethod.lroMetadata.finalResponse?.resultPath) {
-        (<go.LROMethod>method).operationLocationResultPath = sdkMethod.lroMetadata.finalResponse.resultPath;
+      if (sdkMethod.lroMetadata.finalResponse?.resultSegments) {
+        const segment = sdkMethod.lroMetadata.finalResponse?.resultSegments[0];
+        const resultPath = (<tcgc.SdkBodyModelPropertyType>segment).serializationOptions.json?.name;
+        (<go.LROMethod>method).operationLocationResultPath = resultPath;
```
tadelesh, line 221: the result path should be a json path. you could use map and join to combine all serialized name.

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -362,7 +364,7 @@ export class clientAdapter {
             let serializedName: string | undefined;
             for (const property of opParam.type.properties) {
               if (property.name === param.name) {
-                serializedName = (<tcgc.SdkBodyModelPropertyType>property).serializedName;
+                serializedName = (<tcgc.SdkBodyModelPropertyType>property).serializationOptions.json?.name;
```
tadelesh, line 367: you should separate the logic for json and xml.

packages/typespec-go/src/tcgcadapter/clients.ts
JiaqiZhang-Dev: Can i use ''/" to combine all serialized name as the result path?

tadelesh: json path use `.` to combine each part.

packages/typespec-go/src/tcgcadapter/clients.ts
JiaqiZhang-Dev: done

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -34,7 +34,7 @@ export class clientAdapter {
     }
     for (const sdkClient of sdkPackage.clients) {
       // start with instantiable clients and recursively work down
-      if (sdkClient.initialization.access === 'public') {
+      if (sdkClient.clientInitialization.initializedBy & tcgc.InitializedByFlags.Individually) {
```
jhendrixMSFT, line 37: I see these are flags now which usually implies that one or both can be set. Is that true?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -356,13 +358,27 @@ export class clientAdapter {
         const byVal = isTypePassedByValue(param.type);
         const contentType = this.adaptContentType(opParam.defaultContentType);
         switch (contentType) {
-          case 'JSON':
+          case 'JSON':{
+            // find the corresponding field within the model param so we can get the serialized name
+            let serializedName: string | undefined;
```
jhendrixMSFT, line 363: It seems like this is mostly duplicated code with the `XML` case. Can it be consolidated?

JiaqiZhang-Dev: ok, I'll try to consolidate them.

JiaqiZhang-Dev: done~ I added an inline function `getSerializedNameFromProperty` to combine xml and json logic

jhendrixMSFT (approved):

tadelesh (approved):

JiaqiZhang-Dev:
@jhendrixMSFT Hi Joel, could you please help trigger a release of the CodeGen? thanks~

PR #1498, opened 2025-02-19
Refine model filtering logic

tadelesh:
Remove filtering Azure core model since some instances of template model is in `Azure.Core` namespace. Logic of filtering exception model could cover the filtering needs.

The original problem is found in this [PR](https://github.com/Azure/azure-rest-api-specs/pull/32118). Error log [here](https://github.com/Azure/azure-rest-api-specs/pull/32118/checks?check_run_id=37387576873). Spec playground [here](https://cadlplayground.z22.web.core.windows.net/cadl-azure/?c=aW1wb3J0ICJAdHlwZXNwZWMvaHR0cCI7DQrSGnJlc3TWGnZlcnNpb25pbmfNIGF6dXJlLXRvb2xzL8gsLcYVY29yZcQsDQp1c2luZyBUeXBlU3BlYy5IdHRwO9EWUmVzdNIWVsl3yRxBxF4uQ29yZcVdQHVzZUF1dGgoQWFkT2F1dGgyxA48WyLkAOZzOi8vY29udG9zby7lAJYuY29tLy5kZWZhdWx0Il0%2BKQ0KQHNlcnZpY2Uoew0KICB0aXRsZTogIkPGNiBXaWRnZXQgTWFuYWdlciIsDQp9xDTnAQ9lZCjHKS7GKcco6AC4cykNCm5hbWVzcGFjZekAutMx5gDLZG9jKCLIOyBpbmZvLiIpDQplbnVtIMkX5QCnxitUaGUgMjAyMi0xMS0wMS1wcmV2aWV3IOcAncU8ICDkASFEZXBlbmRlbmN5KOoBP%2BkApS52MV8wX1DGPl8xxTZ2xFhfMTFfMDHIGjogItJt5QES6gC9QSB35QDXxXxAcmVzb3VyY2UoIsYWc8QWbW9kZWznAVhTdWl0ZecA0mtlecgoTmFtZecAvekA6MYaIOQBTugA2XZpc2liaWxpdHkoInJlYWTGF8QiOiBzdHLmAjXORklEIG9mIHTJUCdzIG1hbnVmYWN0dXJlcsdazBNJZNlNZmFrZWQgc2hhcmVkIOUA18dAxhNNxBI%2FOiBGxCdTyhM75wEyaW50ZXJm5AIZ5gEN7QHbRmV0Y2ggYccbIGJ57AD6Z2V0xxhpcyBS5wFlUmVhZDzrAVc%2B7QCwR8Rgc3RhdHVz5AECyVlvcGVyYXTqAi7JW0%2FIGVPGNWlzIEdldMhtzx7aeENyZWF0ZXMgb3IgdXBkxQvJfGFzeW5jaHJvbm91c2x56AHWcG9sbGluZ8liKMYucy74AJvFN2PFaE9yVcVlx2FpcyBTdGFuZGFyZPEAuXMuTG9uZ1J1buQEXcge5gCuyEb6ANZEZWxldGX%2FAMr%2FAMr4AMpkxVzHWWlzIPMAp8Z%2F%2BgCfTGlzdOgAm%2BgDhOUDfCAgbGlzx3zlAdPIUsQyPMQgIOwDmizGEuUCl8QBcGFyYW1ldGVyczrpAVnEOlF1ZXJ5UMkdO8YwfcQH5ACZ6wQg5QMM7QMz6gQT8AMo8ATqdGFn5wFadGFn%2BQOV5gIJZCDkAcHHL8cTQXQ6IHV0Y0RhdGVUaW3kBj59&e=%40azure-tools%2Ftypespec-client-generator-core&options=%7B%7D). You could see `ResourceOperationStatusWidgetSuiteWidgetSuiteError` is a template instance model and used in operation response. But we filtered it before.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -45,7 +45,7 @@ export class typeAdapter {
     const ifaceTypes = new Array<InterfaceTypeSdkModelType>();
     for (const modelType of sdkContext.sdkPackage.models) {
       if (modelType.name.length === 0) {
-        // tcgc creates some unamed models for spread params.
+        // tcgc creates some unnamed models for spread params.
         // we don't use these so just skip them.
         continue;
       } else if (modelType.access === 'internal' && (modelType.usage & tcgc.UsageFlags.Spread) === tcgc.UsageFlags.Spread) {
@@ -112,10 +112,7 @@ export class typeAdapter {
   }
 
   private skipSpecificTypes(type: tcgc.SdkModelType | tcgc.SdkEnumType): boolean {
-    if (tcgc.isAzureCoreModel(type)) {
-      // skip core types
-      return true;
-    } else if ((type.usage & tcgc.UsageFlags.Exception) === tcgc.UsageFlags.Exception && (type.usage & tcgc.UsageFlags.Input) === 0 && (type.usage & tcgc.UsageFlags.Output) === 0) {
+    if ((type.usage & tcgc.UsageFlags.Exception) === tcgc.UsageFlags.Exception && (type.usage & tcgc.UsageFlags.Input) === 0 && (type.usage & tcgc.UsageFlags.Output) === 0) {
       // skip error type
       return true;
     } else if ((type.usage & tcgc.UsageFlags.LroPolling) === tcgc.UsageFlags.LroPolling && (type.usage & tcgc.UsageFlags.Input) === 0 && (type.usage & tcgc.UsageFlags.Output) === 0) {
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -3,13 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as naming from '../../../naming.go/src/naming.js';
-import * as go from '../../../codemodel.go/src/index.js';
+import { uncapitalize } from '@azure-tools/codegen';
+import { values } from '@azure-tools/linq';
 import * as tcgc from '@azure-tools/typespec-client-generator-core';
 import * as tsp from '@typespec/compiler';
 import * as http from '@typespec/http';
-import { values } from '@azure-tools/linq';
-import { uncapitalize } from '@azure-tools/codegen';
+import * as go from '../../../codemodel.go/src/index.js';
+import * as naming from '../../../naming.go/src/naming.js';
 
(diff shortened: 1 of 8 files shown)


tadelesh:
@jhendrixMSFT could you provide your thought for this?

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -45,7 +45,7 @@ export class typeAdapter {
     const ifaceTypes = new Array<InterfaceTypeSdkModelType>();
     for (const modelType of sdkContext.sdkPackage.models) {
       if (modelType.name.length === 0) {
-        // tcgc creates some unamed models for spread params.
+        // tcgc creates some unnamed models for spread params.
```
jhendrixMSFT, line 48: I wonder if this is still true? Don't have to change it now, just curious if you know.

jhendrixMSFT:
Is `ResourceOperationStatusWidgetSuiteWidgetSuiteError` a core type that's a field in the result model?

tadelesh:
> Is `ResourceOperationStatusWidgetSuiteWidgetSuiteError` a core type that's a field in the result model?

yes. it is an instance of `ResourceOperationStatus` model without friendly name. tcgc give such name with just joining all template arguments' name.

packages/typespec-go/src/tcgcadapter/types.ts
tadelesh: tcgc should not give any models without name. but it seems there are some anonymous cases that will break this. it's safe to keep it for now.

packages/typespec-go/test/azure-http-specs/azure/resource-manager/commonpropsgroup/zz_models.go
```diff
@@ -6,6 +6,12 @@ package commonpropsgroup
 
 import "time"
 
+// ErrorResponse - A response containing error details.
+type ErrorResponse struct {
```
jhendrixMSFT, line 10: This shows up because the type has flag `Error` instead of `Exception` set. Is this expected?

jhendrixMSFT:
Note that I rebased and pushed a commit to regenerate content.

packages/typespec-go/test/azure-http-specs/azure/resource-manager/commonpropsgroup/zz_models.go
tadelesh: i think so. after tcgc remove the `Error` usage, then the model will be removed from the model list. i refined the model filtering logic to only keep the model with input or output usage. that could also help to filter out the exception and lro middle step models.

jhendrixMSFT (approved):

PR #1499, opened 2025-02-19
Don't export pager constructors for internal methods

chlowell:


packages/codegen.go/src/operations.ts
```diff
@@ -1331,8 +1331,18 @@ function generateLROBeginMethod(client: go.Client, method: go.LROMethod, imports
 export function fixUpMethodName(method: go.Method): string {
   if (go.isLROMethod(method)) {
     return `Begin${method.name}`;
-  } else if (go.isPageableMethod(method)) {
-    return `New${method.name}Pager`;
+  }
+  if (go.isPageableMethod(method)) {
```
jhendrixMSFT, line 1335: I suspect this is also broken for internal LROs :(

jhendrixMSFT (approved):

Issue #1485, opened 2025-01-13
Go CodeGen should address the errors and warnings from TCGC


lirenhe:
We need to empower our customers both internal and external to easily trouble shooting and fix spec issues that would block the CodeGen.

Today, TCGC could already do the check on the patterns and results from TypeSpec compiler and reports warning and errors for the unsupported case and invalid usage.

However, our CodeGen didn't address those errors and warnings from TCGC in the implementation and in most cases would continue the codegen which would result in unexpected generated code & error. When that happens, the outcome would be difficult to trouble shooting.

Proposals:
- Go CodeGen need to display those warnings and errors in the command line output instead of ignoring them.
- Go CodeGen should properly address the errors and warnings to provide a better user experience.

tadelesh:
tcgc issue here: https://github.com/Azure/typespec-azure/issues/1487.
java impl reference: https://github.com/microsoft/typespec/pull/4382.

jhendrixMSFT:
Regarding "Go CodeGen should properly address the errors and warnings to provide a better user experience.", does this mean we need to do something more that just `context.program.reportDiagnostics(sdkContext.diagnostics);`?

jhendrixMSFT:
Also, is there a reason why tcgc's `createSdkContext()` can't report the diagnostics? It should have access to the `context.program` to do so. This way, emitters don't have to do it (they all get it for free).

tadelesh:
first step is to report any tcgc diagnostics. since tcgc is a typespec lib, it should not report them directly according to the typespec best practices. one simple reason is downstream may call lib multiple times and needs to gather the diagnostics together to report. @JiaqiZhang-Dev could you help with the first step.

PR #1500, opened 2025-02-20
go report TCGC diagnostics

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1485

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/adapter.ts
+++ b/packages/typespec-go/src/tcgcadapter/adapter.ts
@@ -30,6 +30,7 @@ export async function tcgcToGoCodeModel(context: EmitContext<GoEmitterOptions>):
     additionalDecorators: ['TypeSpec\\.@encodedName'],
     disableUsageAccessPropagationToBase: true,
   });
+  context.program.reportDiagnostics(sdkContext.diagnostics);
   let codeModelType: go.CodeModelType = 'data-plane';
   if (sdkContext.arm === true) {
     codeModelType = 'azure-arm';
--- a/packages/autorest.go/CHANGELOG.md
+++ b/packages/autorest.go/CHANGELOG.md
@@ -1,16 +1,11 @@
 # Release History
 
-## 4.0.0-preview.71 (unreleased)
-
-### Other Changes
-
-* Added switch `--factory-gather-all-params` to control the `NewClientFactory` constructor parameters. This switch allows gathering either only common parameters of clients or all parameters of clients.
-
-## 4.0.0-preview.70 (unreleased)
+## 4.0.0-preview.70 (2025-02-21)
 
 ### Other Changes
 
 * Upgraded default `azcore` version to `v1.17.0`.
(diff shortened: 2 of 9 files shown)


JiaqiZhang-Dev:
@chlowell This pr is for reporting tcgc diagnostics. When i run the typespec-go test, there is an error caused by `packages\typespec-go\test\tsp\Internal.Pager\main.tsp`, it seems that pageResult model is not valid. I fixed the tsp file so that it could be pass the test,  could you please check the file change is ok?
image

packages/typespec-go/src/tcgcadapter/adapter.ts
```diff
@@ -30,6 +30,7 @@ export async function tcgcToGoCodeModel(context: EmitContext<GoEmitterOptions>):
     additionalDecorators: ['TypeSpec\\.@encodedName'],
     disableUsageAccessPropagationToBase: true,
   });
+  context.program.reportDiagnostics(sdkContext.diagnostics);
```
jhendrixMSFT, line 33: If there are errors, does the compiler stop at this point or does the emitter continue? If it _does_ continue, should we not?

JiaqiZhang-Dev: I met this case(my last comment with a screenshot), the compiler will stop at this point.

jhendrixMSFT: Got it, thanks for clarifying.

chlowell:
> @chlowell This pr is for reporting tcgc diagnostics. When i run the typespec-go test, there is an error caused by `packages\typespec-go\test\tsp\Internal.Pager\main.tsp`, it seems that pageResult model is not valid. I fixed the tsp file so that it could be pass the test, could you please check the file change is ok?

It's ok, your change doesn't affect the test. Thanks for fixing the TSP

jhendrixMSFT (approved):

tadelesh:
> @tadelesh there are a lot of warning diagnostics from tcgc that we're now seeing with this change. They come from `@azure-tools/typespec-azure-resource-manager`, `@azure-tools/typespec-client-generator-core` and `@azure-tools/typespec-azure-core`. Are these known issues?
> 
> See the [log](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4585492&view=logs&jobId=628df61d-eb78-5279-4293-d5d16eefafa6&j=628df61d-eb78-5279-4293-d5d16eefafa6&t=6b429ad4-dc32-5f29-a57a-c9d79eacaabe) for details.

some of them are expected: `@azure-tools/typespec-client-generator-core/example-value-no-mapping`.
some of them are from wrong config: `@azure-tools/typespec-client-generator-core/example-loading`.
some of them are bugs maybe: `@azure-tools/typespec-client-generator-core/unsupported-kind`, `@azure-tools/typespec-azure-core/union-enums-invalid-kind`.

these warnings should have little impact for generated code. we will disable warning detection (consolidated with other languages) in our sdk automation tool to mitigate the spec review problem.

i'll work with @JiaqiZhang-Dev to see how to resolve all the warnings.

PR #1501, opened 2025-02-25
gotest support new codegen

JiaqiZhang-Dev:
Since codegen supports a new switch `factoryGatherAllParams`, the autorest.gotest should support this feature too.


tadelesh:
close in favor of https://github.com/Azure/autorest.go/pull/1516

PR #1504, opened 2025-02-26
avoid warning log from tcgc example-loading

JiaqiZhang-Dev:
avoid warning log from tcgc example-loading; related issue: https://github.com/Azure/typespec-azure/pull/2267


JiaqiZhang-Dev:
@tadelesh could you please help review this pr?

Issue #1479, opened 2024-12-12
Add doc for tspconfig for go


tadelesh:
1. add doc for the config in code: [here](https://github.com/Azure/autorest.go/blob/7e5d72eba64879bfe7ab71602f22b7b1f51a5032/packages/typespec-go/src/lib.ts#L25). refer [this](https://github.com/microsoft/typespec/blob/b2c14ccb8f240c85a52351d547eaa71b8638886c/packages/openapi3/src/lib.ts#L70).
2. add a doc for how to config go for typespec. refer [this](https://github.com/Azure/azure-sdk-for-java/wiki/TypeSpec-Java-QuickStart#java-specific-options-for-tspconfigyaml).

RickWinter:
@tadelesh please update the milestone for when this work will be completed.

JiaqiZhang-Dev:
@tadelesh Could you please help review these docs? thanks~
1. go-emitter options doc: https://github.com/Azure/autorest.go/pull/1507
2. how to config go for typespec: https://github.com/Azure/azure-sdk-for-go/wiki/TypeSpec-Go-QuickStart

tadelesh:
for wiki, is that possible to be consolidated with the config doc in emitter? you could do that after the emitter config pr merge.

JiaqiZhang-Dev:
> for wiki, is that possible to be consolidated with the config doc in emitter? you could do that after the emitter config pr merge.

ok, thanks for your suggestion~

PR #1507, opened 2025-02-27
add go-emitter options description

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1479


packages/typespec-go/src/lib.ts
```diff
@@ -27,21 +27,81 @@ const EmitterOptionsSchema: JSONSchemaType<GoEmitterOptions> = {
...
+    },
+    'module': {
+      type: 'string',
+      nullable: true,
+      description: 'The name of the Go module written to go.mod. Omit to skip go.mod generation.',
+    },
+    'module-version': {
+      type: 'string',
+      nullable: true,
+      description: 'Semantic version of the Go module without the leading \'v\' written to constants.go. (e.g. 1.2.3).',
```
jhendrixMSFT, line 68: We should mention that when `module-version` is specified, you must also specify `module`. And add the same kind of description to `module` too.

tadelesh (approved):

packages/typespec-go/src/lib.ts
JiaqiZhang-Dev: done

Issue #1508, opened 2025-03-03
Do not throw errors but report diagnostic


JiaqiZhang-Dev:
Emitters should report errors as many as possible therefore if possible, emitters should not crash on purpose (for example throw new Error).

Throwing an error will cause the emitter to crash, and when the emitter crashes, typespec compiler will flag this as a bug in the emitter and recommend the user to open an issue for it.

For those unexpected, it is fine that the emitter would crash, because it is "unexpected", we could never prevent it therefore it should be categorized as a bug.

Ensure do not throw Error unless we really want the emitter to crash, otherwise we should report error diagnostic.

RickWinter:
Per @jhendrixMSFT , Perhaps we throw a special exception which is then caught at the top level. This would be reported as an Emitter error rather than a crash. This aligns with typespec's diagnostics model approach.

PR #1513, opened 2025-03-06
report diagnostic for expected error

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1508

### DIFF
--- a/packages/typespec-go/src/emitter.ts
+++ b/packages/typespec-go/src/emitter.ts
@@ -21,144 +21,159 @@ import { generateServerFactory } from '../../codegen.go/src/fake/factory.js';
 import { generateXMLAdditionalPropsHelpers } from '../../codegen.go/src/xmlAdditionalProps.js';
 import { existsSync } from 'fs';
 import { mkdir, readFile, writeFile } from 'fs/promises';
-import { EmitContext } from '@typespec/compiler';
+import { EmitContext, NoTarget } from '@typespec/compiler';
 import 'source-map-support/register.js';
+import { reportDiagnostic } from './lib.js';
 
 export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
-  const codeModel = await tcgcToGoCodeModel(context);
-  await mkdir(context.emitterOutputDir, {recursive: true});
+  try {
+    const codeModel = await tcgcToGoCodeModel(context);
+    await mkdir(context.emitterOutputDir, {recursive: true});
 
-  // don't overwrite an existing go.mod file, update it if required
-  const goModFile = `${context.emitterOutputDir}/go.mod`;
-  let existingGoMod: string | undefined;
-  if (existsSync(goModFile)) {
-    existingGoMod = (await readFile(goModFile)).toString();
-  }
-  const gomod = await generateGoModFile(codeModel, existingGoMod!);
-  if (gomod.length > 0) {
-    writeFile(goModFile, gomod);
-  }
-
-  let filePrefix = '';
-  if (context.options['file-prefix']) {
-    filePrefix = context.options['file-prefix'];
-    // if a file prefix was specified, ensure it's properly snaked
-    if (filePrefix[filePrefix.length - 1] !== '_') {
-      filePrefix += '_';
+    // don't overwrite an existing go.mod file, update it if required
+    const goModFile = `${context.emitterOutputDir}/go.mod`;
+    let existingGoMod: string | undefined;
+    if (existsSync(goModFile)) {
+      existingGoMod = (await readFile(goModFile)).toString();
     }
-  }
-
-  const clientFactory = await generateClientFactory(codeModel);
-  if (clientFactory.length > 0) {
-    writeFile(`${context.emitterOutputDir}/${filePrefix}client_factory.go`, clientFactory);
-  }
-
-  const constants = await generateConstants(codeModel);
-  if (constants.length > 0) {
-    writeFile(`${context.emitterOutputDir}/${filePrefix}constants.go`, constants);
-  }
-
-  const interfaces = await generateInterfaces(codeModel);
-  if (interfaces.length > 0) {
-    writeFile(`${context.emitterOutputDir}/${filePrefix}interfaces.go`, interfaces);
-  }
-
-  const models = await generateModels(codeModel);
-  if (models.models.length > 0) {
-    writeFile(`${context.emitterOutputDir}/${filePrefix}models.go`, models.models);
-  }
-  if (models.serDe.length > 0) {
-    writeFile(`${context.emitterOutputDir}/${filePrefix}models_serde.go`, models.serDe);
-  }
-
-  const operations = await generateOperations(codeModel);
-  for (const op of operations) {
-    let fileName = op.name.toLowerCase();
-    // op.name is the client name, e.g. FooClient.
-    // insert a _ before Client, i.e. Foo_Client
-    // if the name isn't simply Client.
-    if (fileName !== 'client') {
-      fileName = fileName.substring(0, fileName.length-6) + '_client';
+    const gomod = await generateGoModFile(codeModel, existingGoMod!);
+    if (gomod.length > 0) {
+      writeFile(goModFile, gomod);
     }
-    writeFile(`${context.emitterOutputDir}/${filePrefix}${fileName}.go`, op.content);
-  }
 
-  if (codeModel.options.generateExamples) {
-    const examples = await generateExamples(codeModel);
-    for (const example of examples) {
-      let fileName = example.name.toLowerCase();
+    let filePrefix = '';
+    if (context.options['file-prefix']) {
+      filePrefix = context.options['file-prefix'];
+      // if a file prefix was specified, ensure it's properly snaked
+      if (filePrefix[filePrefix.length - 1] !== '_') {
+        filePrefix += '_';
+      }
+    }
+    const clientFactory = await generateClientFactory(codeModel);
+    if (clientFactory.length > 0) {
+      writeFile(`${context.emitterOutputDir}/${filePrefix}client_factory.go`, clientFactory);
+    }
+
+    const constants = await generateConstants(codeModel);
+    if (constants.length > 0) {
+      writeFile(`${context.emitterOutputDir}/${filePrefix}constants.go`, constants);
+    }
+
+    const interfaces = await generateInterfaces(codeModel);
+    if (interfaces.length > 0) {
+      writeFile(`${context.emitterOutputDir}/${filePrefix}interfaces.go`, interfaces);
+    }
+
+    const models = await generateModels(codeModel);
+    if (models.models.length > 0) {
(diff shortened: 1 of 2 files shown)


packages/typespec-go/src/emitter.ts
```diff
@@ -21,144 +21,159 @@ import { generateServerFactory } from '../../codegen.go/src/fake/factory.js';
...
-    writeFile(goModFile, gomod);
-  }
-
-  let filePrefix = '';
-  if (context.options['file-prefix']) {
-    filePrefix = context.options['file-prefix'];
-    // if a file prefix was specified, ensure it's properly snaked
-    if (filePrefix[filePrefix.length - 1] !== '_') {
-      filePrefix += '_';
+  try {
```
tadelesh, line 29: can we put the try body into another func. try clause is too long to know what kind of exception it will handle.

packages/typespec-go/src/emitter.ts
```diff
@@ -21,144 +21,159 @@ import { generateServerFactory } from '../../codegen.go/src/fake/factory.js';
...
-      if (polymorphics.length > 0) {
-        writeFile(`${fakesDir}/${filePrefix}polymorphic_helpers.go`, polymorphics);
+        const polymorphics = await generatePolymorphicHelpers(codeModel, 'fake');
+        if (polymorphics.length > 0) {
+          writeFile(`${fakesDir}/${filePrefix}polymorphic_helpers.go`, polymorphics);
+        }
       }
     }
+  } catch (error: any) {
+    if (error.constructor === Error) {
```
tadelesh, line 166: is this the idiomatic way for ts to filter customized error? i prefer to define customized error for generator or emitter and use `instance of` here to do the filtering.

packages/typespec-go/src/emitter.ts
JiaqiZhang-Dev: It's really too long, I put the try body into `generate` func.

packages/typespec-go/src/emitter.ts
JiaqiZhang-Dev: I defined a customized  error `DiagnosticError`, since the errors are caused from different packages like codegen.go, codemodel.go, i replace `throw new Error` into `throw new DiagnosticError` in this packages, could you please help review if it's ok? thanks~

tadelesh (approved):

JiaqiZhang-Dev:
@jhendrixMSFT could you please help review this pr? thanks~

jhendrixMSFT:
Please note I rebased this on `main`.

packages/typespec-go/src/emitter.ts
```diff
@@ -161,4 +181,4 @@ export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
       }
     }
   }
-}
```
jhendrixMSFT, line 164: Missing new-line.

jhendrixMSFT:
I took the liberty to expand on this, splitting `DiagnosticError` into different error types and adding some error codes to help determine when we should emit the stack trace.

In addition, the `AdapterError` includes the tsp target so the compiler will provide line number info where the adaptation failed. Here's an example of the output when trying to build `uniongroup` which is not currently supported.

```
../C:/git/Azure/autorest.go/packages/typespec-go/node_modules/@typespec/http-specs/specs/type/union/main.tsp:160:25 - error @azure-tools/typespec-go/AdapterUnsupportedTsp: The emitter encountered a TypeSpec definition that is currently not supported.
Error: unsupported type kind union
    at typeAdapter.getPossibleType (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:247:15)
    at typeAdapter.getModelField (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:586:21)
    at typeAdapter.adaptTypes (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:102:28)
    at tcgcToGoCodeModel (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/adapter.ts:64:6)

> 160 | alias MixedTypesUnion = Cat | "a" | int32 | boolean;
      |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^

Found 1 error, 9 warnings.
```

PTAL and let me know what you think.

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -171,7 +173,7 @@ export class clientAdapter {
   private adaptURIParam(sdkParam: tcgc.SdkPathParameter): go.URIParameter {
     const paramType = this.ta.getPossibleType(sdkParam.type, true, false);
     if (!go.isConstantType(paramType) && !go.isPrimitiveType(paramType)) {
-      throw new Error(`unexpected URI parameter type ${go.getTypeDeclaration(paramType)}`);
+      throw new AdapterError('UnsupportedTsp', `unsupported URI parameter type ${go.getTypeDeclaration(paramType)}`, sdkParam.__raw!.node);
```
jhendrixMSFT, line 176: There are quite a few places where `__raw` is `?` which is unfortunate. @tadelesh are there cases where `__raw` and/or `__raw.node` will be absent? If so, then this will need to be cleaned up. Or is there a better way to get the node? Having the node gives a better user experience as the compiler will output file and line number info as part of the diagnostic.

tadelesh: some types are not defined by user, but created by tcgc. for example, discriminated property of implicit `@discriminator`. for such cases, we do not have `__raw` type.

JiaqiZhang-Dev:
> I took the liberty to expand on this, splitting `DiagnosticError` into different error types and adding some error codes to help determine when we should emit the stack trace.
> 
> In addition, the `AdapterError` includes the tsp target so the compiler will provide line number info where the adaptation failed. Here's an example of the output when trying to build `uniongroup` which is not currently supported.
> 
> ```
> ../C:/git/Azure/autorest.go/packages/typespec-go/node_modules/@typespec/http-specs/specs/type/union/main.tsp:160:25 - error @azure-tools/typespec-go/AdapterUnsupportedTsp: The emitter encountered a TypeSpec definition that is currently not supported.
> Error: unsupported type kind union
>     at typeAdapter.getPossibleType (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:247:15)
>     at typeAdapter.getModelField (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:586:21)
>     at typeAdapter.adaptTypes (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/types.ts:102:28)
>     at tcgcToGoCodeModel (file:///C:/git/Azure/autorest.go/packages/typespec-go/src/tcgcadapter/adapter.ts:64:6)
> 
> > 160 | alias MixedTypesUnion = Cat | "a" | int32 | boolean;
>       |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> Found 1 error, 9 warnings.
> ```
> 
> PTAL and let me know what you think.

It's an excellent design that enhances error handling by providing more specific error types and codes, and the new error files help maintain a clean and modular codebase. Thanks for your improvement! 👍

packages/typespec-go/src/tcgcadapter/clients.ts
jhendrixMSFT: fixed

packages/codegen.go/src/errors.ts
```diff
@@ -0,0 +1,19 @@
...
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ErrorCode } from "../../codemodel.go/src/errors.js";
+
+/**
+ * CodegenError is thrown when the emitter fails some condition
+ * in order to generate a part of the code model.
+ */
+export class CodegenError extends Error {
```
RickWinter, line 12: NIT: Should we call this an EmitterError since its coming from the Emitter?

CodeGen might make more sense as its more broad. At face value it seems too broad to me.

packages/codegen.go/src/example.ts
```diff
@@ -416,7 +417,7 @@ function generateFakeExample(goType: go.PossibleType, name?: string): go.Example
         return new go.NumberExample(goType.values[0].value as number, goType);
     }
   }
-  throw new Error(`do not support to fake example for none primitive type: ${goType}`);
+  throw new CodegenError('InternalError', `do not support to fake example for none primitive type: ${go.getTypeDeclaration(goType)}`);
```
RickWinter, line 420: Why internal error?

Also the grammar of the sense is tough to parse. 
Should we say: "fakes do not support none primitive type: ........"

packages/typespec-go/src/lib.ts
```diff
@@ -117,7 +117,32 @@ const EmitterOptionsSchema: JSONSchemaType<GoEmitterOptions> = {
 
 const libDef = {
   name: '@azure-tools/typespec-go',
-  diagnostics: {},
+  diagnostics: {
+    'InternalError': {
+      severity: 'error',
+      messages: {
+        default: paramMessage`The emitter encountered an internal error during preprocessing. Please open an issue at https://github.com/Azure/autorest.go/issues and include the complete error message.\n${'stack'}`
```
RickWinter, line 124: Should we differentiate between an internal error and a `not implemented` error? I'm not convinced one way or the other so this is just something to think about.

RickWinter (approved):

packages/codegen.go/src/errors.ts
jhendrixMSFT: There are three different types of errors.
- `AdapterError` in `typespec-go/src/errors.ts`
  - indicates an error transformting tcgc code model to Go code model
  - this error can include the originating tsp node so we get source/line number info
- `CodeModelError` in `codemodel.go/src/errors.ts`
  - is a violation of some invariant in the Go code model. this is always a bug in the adapter
- `CodegenError` in `codegen.go/src/errors.ts`
  - since the tsp has been consumed at this point, there's no tsp node, which is why we have a unique error type here
  - with rare exception, this kind of error is always a bug in the codegen portion of the emitter
  - I suspect that with some refactoring, some of these errors currently being thrown can go away

In addition, there are different error codes indicating the type of error. See `codemodel.go/src/errors.ts` for the list of error codes.

packages/codegen.go/src/example.ts
jhendrixMSFT: Reworded the sentence.

For error cases where the codegen portion doesn't handle some case, I classified those as `InternalError` (compared to some case e.g. we expect `X` but got `Y`). We _could_ add a separate error code to separate the two however I don't think it would add a lot of value. In both cases we want to dump the stack and error message informing the user to open an issue. Whether or not it was because some precondition failed, or we just didn't handle some case is less important as the stack will tell us which case it was.

Issue #1515, opened 2025-03-10
Hard-coded monomorphic field name introduces breaking changes


jhendrixMSFT:
The tsp emitter uses a hard-coded name for the value field in a monomorphic response envelope.

https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/src/tcgcadapter/clients.ts#L683

Contrast this to the swagger emitter which recursively constructs the type name.

https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/src/transform/transform.ts#L886

jhendrixMSFT:
Unfortunately, just changing this now could also result in breaking changes. We could add a switch to avoid this, but it's One More Switch which is clunky. In addition, it might not be necessary.

I scanned through the ARM SDKs and all of the monomorphic response envelopes are either of type `string` or an enum type.

I propose that in such cases we leave the field named `Value`.

For arrays/maps of integral types, we change the field name to `Values`. And for arrays/maps of UDTs we follow the same algorithm as the swagger emitter.

This assumes that the investigation was correct, it would be good to double-check before we commit to this.

tadelesh:
i've double confirmed with arm sdk. several cases found:
- array of UDTs, `xxxArray`, [example](https://github.com/Azure/azure-sdk-for-go/blob/2f5ce0599344377c1caadd5ece4110dbb7eed7a2/sdk/resourcemanager/addons/armaddons/response_types.go#L38)
- map of UDTs, `Value`, [example](https://github.com/Azure/azure-sdk-for-go/blob/2f5ce0599344377c1caadd5ece4110dbb7eed7a2/sdk/resourcemanager/logic/armlogic/response_types.go#L315)
- anything, `Interface`, [example](https://github.com/Azure/azure-sdk-for-go/blob/2f5ce0599344377c1caadd5ece4110dbb7eed7a2/sdk/resourcemanager/appservice/armappservice/responses.go#L527)
- basic type, `Value`, [example](https://github.com/Azure/azure-sdk-for-go/blob/2f5ce0599344377c1caadd5ece4110dbb7eed7a2/sdk/resourcemanager/appservice/armappservice/responses.go#L2352)
- array of basic type, `xxxArray`, [example](https://github.com/Azure/azure-sdk-for-go/blob/2f5ce0599344377c1caadd5ece4110dbb7eed7a2/sdk/resourcemanager/containerinstance/armcontainerinstance/responses.go#L73)
[...]

PR #1517, opened 2025-03-11
Fix monomorphic response field name

jhendrixMSFT:
The hard-coded name Value introduced some breaking changes for services migrating to tsp. In addition, as some tsp-based SDKs have shipped with this name, we cannot simply revert to the autorest.go behavior.

The recursiveTypeName implementation is essentially a hybrid of the autorest.go and typespec-go implementations. See the function's doc comments for an explanation of the naming algorithm. This result is based on an analysis of released SDKs and a pending release which is inadvertently being held up due to the result field being renamed.

Fixes https://github.com/Azure/autorest.go/issues/1515

### DIFF
--- a/packages/typespec-go/CHANGELOG.md
+++ b/packages/typespec-go/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Release History
 
-## 0.4.0 (2025-03-11)
+## 0.4.0 (2025-03-12)
 
 ### Breaking Changes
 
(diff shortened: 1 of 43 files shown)


jhendrixMSFT:
autorest.go behavior for reference: https://github.com/Azure/autorest.go/blob/main/packages/autorest.go/src/transform/transform.ts#L1008-L1062

tadelesh (approved):

PR #1527, opened 2025-03-25
Update CI to the latest version of Go

jhendrixMSFT:


jhendrixMSFT:
@benbp how hared would it be to update this to include the N-1 version of Go too?

jhendrixMSFT:
Also, seems like changes to `/eng` should trigger the `typespec-go` CI too.

benbp:
> @benbp how hared would it be to update this to include the N-1 version of Go too?

https://github.com/Azure/autorest.go/pull/1529

PR #1529, opened 2025-03-25
Centralize test version definitions and test go latest + previous

benbp:
@jhendrixMSFT this updates the pipeline configs to test go latest and N-1 versions. Wasn't sure if you wanted all or some of the pipelines to test both versions, so just let me know which ones (and also if you want a variable naming change).

### DIFF
--- a/eng/pipelines/ci-go.yml
+++ b/eng/pipelines/ci-go.yml
@@ -30,7 +30,7 @@ extends:
       - stage: "AutoRest_Go_CI"
         jobs:
           - job: "AutoRest_Go_CI"
-            displayName: "Run AutoRest CI Checks - $(GoVersion)"
+            displayName: "Run AutoRest CI Checks - "
 
             variables:
               - template: /eng/pipelines/templates/variables/globals.yml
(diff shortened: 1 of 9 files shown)


eng/pipelines/ci-go.yml
```diff
@@ -30,16 +30,24 @@ extends:
       - stage: "AutoRest_Go_CI"
         jobs:
           - job: "AutoRest_Go_CI"
-            displayName: "Run AutoRest CI Checks"
+            displayName: "Run AutoRest CI Checks - $(GoVersion)"
```
jhendrixMSFT, line 33: Is it possible for `$(GoVersion)` to resolve in the job names in azdo?

jhendrixMSFT (approved):

eng/pipelines/ci-go.yml
benbp: Oh, whoops! I was assuming this would do it, but maybe not. Let me see if I can figure this out.

We have the same challenge with the updates to the matrix files/env usage in azure-sdk-for-go, but that warrants a different solution.

jhendrixMSFT: No worries, it's not urgent so if it's gonna burn a bunch of time feel free to skip it (I wasn't sure if this was just a bug or a limitation on some azdo stuff).

PR #1530, opened 2025-04-02
codegen time for fakes

gracewilcox:
Fixes https://github.com/Azure/autorest.go/issues/1433

Currently, the code generator is creating extra time helpers that are unused. This PR modifies the code generator to only generate time helpers when fakes actually need them.


packages/codegen.go/src/time.ts
```diff
@@ -51,51 +51,77 @@ export async function generateTimeHelpers(codeModel: go.CodeModel, packageName?:
...
-  // needsSerDeHelpers is only required when time.Time is a struct field
-  let needsSerDeHelpers = false;
-
-  for (const model of codeModel.models) {
-    for (const field of values(model.fields)) {
-      const unwrappedField = recursiveUnwrapMapSlice(field.type);
-      if (!go.isTimeType(unwrappedField)) {
-        continue;
+    for (const respEnv of codeModel.responseEnvelopes) {
+      for (const header of respEnv.headers) {
```
jhendrixMSFT, line 115: If checking response headers is necessary for fakes, why isn't it necessary for non-fakes?

gracewilcox: For fakes, it uses the [`helpers.formatValue`](https://github.com/Azure/autorest.go/blob/4cc6519fd423c84bfff8f0f0c3d53bb5b88f0a58/packages/codegen.go/src/fake/servers.ts#L399C85-L399C134) function to help generate code for the response headers. That function adds the time helpers for `timeUnix` and `timeRFC3339`.

For non-fakes, it uses the [`formatHeaderResponseValue` ](https://github.com/Azure/autorest.go/blob/4cc6519fd423c84bfff8f0f0c3d53bb5b88f0a58/packages/codegen.go/src/operations.ts#L289) function to generate code for response headers. That functions does not use any time helpers.

packages/codegen.go/src/time.ts
```diff
@@ -51,51 +51,77 @@ export async function generateTimeHelpers(codeModel: go.CodeModel, packageName?:
...
+      setHelper(unwrappedResult.dateTimeFormat);
     }
-    const unwrappedResult = recursiveUnwrapMapSlice(respEnv.result.monomorphicType);
-    if (!go.isTimeType(unwrappedResult)) {
-      continue;
+  } else {
+    for (const client of codeModel.clients) {
+      for (const method of client.methods) {
+        for (const param of method.parameters) {
+          if (go.isTimeType(param.type) && go.isBodyParameter(param)) {
```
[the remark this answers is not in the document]
jhendrixMSFT: I think this also applies to body params that are `map[string]*time.Time` or `[]time.Time`.

There are some tests that cover this, but we currently don't build them due to their being "poisoned" with union types. At any rate, here are the older iterations of these tests.

https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/test/http-specs/type/arraygroup/zz_arraydatetimevalue_client.go#L76
https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/test/http-specs/type/dictionarygroup/zz_dictionarydatetimevalue_client.go#L78

packages/codegen.go/src/time.ts
jhendrixMSFT: Ugh, seems like we should consolidate this at some point. Thanks for the explanation.

packages/autorest.gotest/test/integrationtest/output/appplatform/armappplatform/spring/go.mod
jhendrixMSFT, line 1: To unblock CI you can just revert all changes under `autorest.gotest`.

packages/codegen.go/src/time.ts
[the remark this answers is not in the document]
gracewilcox: There is a [current test](https://github.com/Azure/autorest.go/blob/4cc6519fd423c84bfff8f0f0c3d53bb5b88f0a58/packages/typespec-go/test/http-specs/type/arraygroup/fake/zz_arraydatetimevalue_server.go#L112) for a body param that is `[]time.Time`. It doesn't use a time helper. 

[This](https://github.com/Azure/autorest.go/blob/4cc6519fd423c84bfff8f0f0c3d53bb5b88f0a58/packages/codegen.go/src/fake/servers.ts#L654) is the part of the fakes code generator that creates the code for body parameters.
 Looks like the pattern is: 
```go
server.UnmarshalRequestAsJSON[<TYPE>](req)
```

If it's a slice of type `time.Time`, it'll be:
```go
server.UnmarshalRequestAsJSON[[]time.Time](req)
```

And if I'm reading the code correctly, if it's a `map[string]*time.Time`, then:
```go
server.UnmarshalRequestAsJSON[map[string]*time.Time](req)
```

In both situations, there is no time helper.

jhendrixMSFT (commented):
Looks like CI is failing due to another time helper file that's being deleted.

jhendrixMSFT:
Can you please add entries to the changelogs?

packages/autorest.go/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 4.0.0-preview.72 (unreleased)
```
jhendrixMSFT, line 3: Could you set this to today's date? I'm going to release this emitter per request of @JiaqiZhang-Dev

jhendrixMSFT (approved):

PR #1542, opened 2025-04-25
Update tsp toolset

jhendrixMSFT:
Stubbed out test for new scenario that's NYI.

The entries in `package.json` are now sorted. Noisy now but will be better going forward.


jhendrixMSFT:
@gracewilcox FYI

gracewilcox (approved):

PR #1545, opened 2025-04-28
Throw error for optional path parameter

JiaqiZhang-Dev:
Only used for key vault for which Go uses post processing to support it. Not a priority for now. So, emitter should report error if spec uses this feature.

spec: https://github.com/gracewilcox/azure-rest-api-specs/blob/a7cc4073991257a7b68cb7062b5e5b504a787d12/specification/keyvault/Security.KeyVault.Secrets/routes.tsp#L101

resolve: https://github.com/Azure/autorest.go/issues/1575


jhendrixMSFT (changes requested):
This will break our ability to generate KV SDKs. We need to add proper support for optional path params and should wait until the KV tsp files have been updated.

tadelesh:
close since we decided to add full support for optional path param.

PR #1547, opened 2025-04-28
Add dependabot for Go modules

jhendrixMSFT:


RickWinter (commented):
What's the noise level of adding this. Will we have hundreds of PRs to review or will it lump it all into one?

jhendrixMSFT:
Good catch. The intent was to group them all into one PR. That's now done.

PR #1549, opened 2025-04-30
Use pnpm instead of rushjs

jhendrixMSFT:
Deleted various files used by rushjs and added pnpm workspace.
Consolidated tsconfig.json settings.
The utility scripts like gogetu.js etc now require an explicit argument for the cwd else they run against the entire repo (pnpm has no equivalent to RUSH_INVOKED_FOLDER at this time). In addition, the
utility scripts have been converted to common module format.
autorest.go and typespec-go switched to vitest although it isn't used
right now. autorest.gotest remains on jest.
Updates eslint config files to flat format. autorest.gotest will maintain its own config file due to lots of lint.
Fixed up some test issues in autorest.gotest.
CI has been updated to install pnpm.
The CI for typespec-go will now build and install the tarball to be consumed by tspcompile.js (tests the real-world scenario; we already do this for autorest.go).


benbp (approved):

.scripts/semaphore.js
```diff
@@ -1,103 +1,100 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // original sources can be found at https://github.com/abrkn/semaphore.js
```
RickWinter, line 2: Does this license notice still make sense given how much the impl has changed?

jhendrixMSFT: Yeah I think so. The impl itself hasn't changed. It's just been converted to common module format.

PR #1550, opened 2025-05-01
Clean up some lint

jhendrixMSFT:


packages/autorest.go/src/m4togocodemodel/adapter.ts
```diff
@@ -39,14 +41,14 @@ export async function m4ToGoCodeModel(host: AutorestExtensionHost) {
     
     const codeModel = new go.CodeModel(info, type, session.model.language.go!.packageName, options);
     if (session.model.language.go!.host) {
-      codeModel.host = session.model.language.go!.host;
+      codeModel.host = <string | undefined>session.model.language.go!.host;
```
RickWinter, line 44: Wouldn't this need to be defined to get into this if clause?

packages/autorest.go/src/m4togocodemodel/adapter.ts
```diff
@@ -39,14 +41,14 @@ export async function m4ToGoCodeModel(host: AutorestExtensionHost) {
...
     }
     if (session.model.language.go!.module && session.model.language.go!.moduleVersion) {
       codeModel.options.module = new go.Module(session.model.language.go!.module, session.model.language.go!.moduleVersion);
     } else if (session.model.language.go!.module || session.model.language.go!.moduleVersion) {
       throw new Error('--module and --module-version must both or neither be set');
     } else if (session.model.language.go!.containingModule !== '') {
-      codeModel.options.containingModule = session.model.language.go!.containingModule;
+      codeModel.options.containingModule = <string | undefined>session.model.language.go!.containingModule;
```
RickWinter, line 51: Does it make sense to assign undefined to the containingModule?

jhendrixMSFT: This is like the above case. I removed the `| undefined`.

RickWinter (approved):

Issue #1567, opened 2025-05-17
[typespec] example mapping for `QualifiedType`


tadelesh:
current example mapping does not handle `QualifiedType` type in Go, which will crash some service's example generation. see this [example](https://github.com/Azure/azure-rest-api-specs/pull/34508).

PR #1572, opened 2025-05-21
example mapping for QualifiedType

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1567


packages/codemodel.go/src/examples.ts
```diff
@@ -89,6 +89,14 @@ export interface StructExample {
   type: ModelType | PolymorphicType;
 }
 
+export interface QualifiedExample {
+  kind: 'qualified';
+  value: any;
+  type: QualifiedType;
+  packageName: string;
+  exportName: string;
```
tadelesh, line 97: are these two properties redundant from `QualifiedType`?

packages/codegen.go/src/example.ts
```diff
@@ -228,6 +228,9 @@ function getExampleValue(codeModel: go.CodeModel, example: go.ExampleType, inden
       exampleText = getConstantValue(codeModel, example.type.type, example.type.literal.value);
     }
     return `${indent}${getPointerValue(example.type, exampleText, byValue, imports)}`;
+  } else if (example.kind === 'qualified') {
+    imports?.add(example.type.packageName);
+    return `${indent}${getRef(byValue)}${go.getTypeDeclaration(example.type, codeModel.packageName)}(${jsonToGo(example.value, '')})`;
```
tadelesh, line 233: i think we need to deal with the qualified type case by case. the existed three types have different initialization way.

packages/codemodel.go/src/examples.ts
JiaqiZhang-Dev: No, the `packageName` is used to add imports(https://github.com/Azure/autorest.go/pull/1572/files#diff-75792d77625ea70565ef4f1b215d2cb802e3a51198b5466d187198f33cb51d3fR232), the `exportName` is used to generated type declaration.(https://github.com/Azure/autorest.go/blob/main/packages/codemodel.go/src/type.ts#L307)

packages/codegen.go/src/example.ts
JiaqiZhang-Dev: do you mean `MultipartContent`, `ETag` and `ReadSeekCloser`?

packages/codemodel.go/src/examples.ts
JiaqiZhang-Dev: done

packages/codegen.go/src/example.ts
JiaqiZhang-Dev: done

packages/codemodel.go/src/examples.ts
jhendrixMSFT, line 1: Can you please keep the names properly sorted?

packages/codegen.go/src/example.ts
```diff
@@ -295,6 +292,17 @@ function getExampleValue(codeModel: go.CodeModel, example: go.ExampleType, inden
     }
     exampleText += `${indent}}`;
     return exampleText;
+  } else if (example.kind === 'qualified') {
+    imports?.add(example.type.packageName);
+    if (example.type.exportName === 'ETag') {
+      return `${indent}${getRef(byValue)}${go.getTypeDeclaration(example.type, codeModel.packageName)}(${jsonToGo(example.value, '')})`;
+    } else if (example.type.exportName === 'MultipartContent') {
+      // TODO: handle MultipartContent, currently just return nil
+      return `${indent}${getRef(byValue)}${go.getTypeDeclaration(example.type, codeModel.packageName)}(nil)`;
```
tadelesh, line 301: maybe just throw? i don't know if nil could compile.

packages/codemodel.go/src/examples.ts
```diff
@@ -65,6 +65,12 @@ export interface ParameterExample {
   value: ExampleType;
 }
 
+export interface QualifiedExample {
```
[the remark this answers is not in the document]
lirenhe: looks this pattern is using all over this class, consider doing some refactor

packages/codemodel.go/src/examples.ts
JiaqiZhang-Dev: done

packages/codemodel.go/src/examples.ts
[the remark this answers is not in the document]
tadelesh: this is a common patten for go code model. we could refactor in another pr.

Issue #1568, opened 2025-05-17
[typespec] diagnostic should contain enough info for customer to change the spec


tadelesh:
current `NameCollision` diagnostic does not contain the detailed collision info which makes it hard for spec author to fix the spec. example [here](https://dev.azure.com/azure-sdk/public/_build/results?buildId=4878082&view=logs&j=83516c17-6666-5250-abde-63983ce72a49&t=00be4b52-4a63-5865-8e02-c61723ad0692).

PR #1573, opened 2025-05-22
upgrade resource manager&fix issue

JiaqiZhang-Dev:
1. resolve: https://github.com/Azure/autorest.go/issues/1568
 2. upgrade resource manager


tadelesh (approved):

packages/typespec-go/test/azure-http-specs/azure/client-generator-core/accessgroup/_metadata.json
```diff
@@ -1,3 +1,3 @@
 {
-  "emitterVersion": "0.4.8"
+  "emitterVersion": "0.4.9"
```
tadelesh, line 2: could we find a solution not to check it with a real version but a fake version to prevent updating generated code every time emitter upgrade?

JiaqiZhang-Dev: ok, I'll working on it

packages/typespec-go/src/lib.ts
```diff
@@ -134,7 +134,7 @@ const libDef = {
     'NameCollision': {
       severity: 'error',
       messages: {
-        default: 'The emitter automatically renamed one or more types which resulted in a type name collision. Please update the client.tsp to rename the type(s) to avoid the collision.'
+        default: paramMessage`The emitter automatically renamed one or more types which resulted in a type name collision. Please update the client.tsp to rename the type(s) to avoid the collision.\n${'stack'}`
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: Yes, it could provide the detailed collision name, so that user can modify their tsp files. This is the releated issue: https://github.com/Azure/autorest.go/issues/1568
Screenshot 2025-05-22 193157

PR #1578, opened 2025-05-29
fix wrong value example generation for any type with array or dict value

tadelesh:
block this pr: https://github.com/Azure/azure-rest-api-specs/pull/34293


tadelesh:
@jhendrixMSFT could you help to release this and upgrade version in sdk repo to unblock the service's pr?

Issue #1576, opened 2025-05-26
Query parameter re-injection


JiaqiZhang-Dev:
Test case https://github.com/Azure/typespec-azure/pull/2497

Only used for storage for which Go always reinject all params for paging. Emitter should report error if spec uses this feature.

PR #1583, opened 2025-06-03
throw error when paging with re-injection parameters

JiaqiZhang-Dev:
resolve: https://github.com/Azure/autorest.go/issues/1576


JiaqiZhang-Dev:
waiting for https://github.com/Azure/typespec-azure/pull/2773 to merge

JiaqiZhang-Dev:
> waiting for [Azure/typespec-azure#2773](https://github.com/Azure/typespec-azure/pull/2773) to merge

merged, upgrade to dev version

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -206,6 +206,9 @@ export class clientAdapter {
     if (sdkMethod.kind === 'basic') {
       method = new go.Method(methodName, goClient, sdkMethod.operation.path, sdkMethod.operation.verb, statusCodes, naming);
     } else if (sdkMethod.kind === 'paging') {
+      if (sdkMethod.pagingMetadata.nextLinkReInjectedParametersSegments !== undefined && sdkMethod.pagingMetadata.nextLinkReInjectedParametersSegments.length > 0) {
```
tadelesh, line 209: this is a compiler core bug. you could rely on `nextLinkReInjectedParametersSegments` only. let's wait for next typespec release.

jhendrixMSFT (approved):

packages/typespec-go/src/tcgcadapter/clients.ts
JiaqiZhang-Dev: The bug is when developer uses re-injection parameters, the element of `nextLinkReInjectedParametersSegments` is always undefined (eg:[undefined, undefined]), actually we do not rely on the elements, we just judge whether it's empty, so we don't need to wait for the release.

PR #1585, opened 2025-06-06
upgrade version and key vault test

gracewilcox:
tcgc now allows the removal of optional parameters with the `override` decorator.

Upgrading to the newest version of tcgc and expanding the Key Vault test to use this new feature. I also fixed the test so it will actually use the client.tsp file.

### DIFF
--- a/packages/typespec-go/package.json
+++ b/packages/typespec-go/package.json
@@ -53,11 +53,11 @@
     "@azure-tools/typespec-autorest": "0.56.0",
     "@azure-tools/typespec-azure-core": "0.56.0",
     "@azure-tools/typespec-azure-resource-manager": "0.56.2",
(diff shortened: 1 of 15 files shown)


packages/typespec-go/package.json
```diff
@@ -77,7 +77,7 @@
   "peerDependencies": {
     "@azure-tools/typespec-client-generator-core": "0.56.2",
```
jhendrixMSFT, line 78: This needs to be updated to the same version as above.

gracewilcox: fixed in [8c448ff](https://github.com/Azure/autorest.go/pull/1585/commits/8c448ff4b20291738f5d792ab9e34e9508ddb915)

jhendrixMSFT (approved):

jhendrixMSFT (approved):

Issue #1577, opened 2025-05-29
[TypeSpec] Naming logic not consistent


tadelesh:
[...]

PR #1589, opened 2025-06-10
Fix field names for optional params and monomorphic respones

jhendrixMSFT:
This aligns with the code emitted by the autorest.go emitter.

Fixes https://github.com/Azure/autorest.go/issues/1577


jhendrixMSFT:
Added a test to ensure that `Sku` becomes `SKU`.

chlowell (approved):

PR #1592, opened 2025-06-11
Refactor code model method types

jhendrixMSFT:
Switch to using discriminated unions to distinguish method types. Switch on method.kind instead of if/else chains where appropriate.

### DIFF
--- a/packages/autorest.go/src/m4togocodemodel/clients.ts
+++ b/packages/autorest.go/src/m4togocodemodel/clients.ts
@@ -34,27 +34,27 @@ export function adaptClients(m4CodeModel: m4.CodeModel, codeModel: go.CodeModel)
 
       if (helpers.isLROOperation(op) && helpers.isPageableOperation(op)) {
         method = new go.LROPageableMethod(op.language.go!.name, client, httpPath, httpMethod, getStatusCodes(op), naming);
-        (<go.LROPageableMethod>method).finalStateVia = (op.extensions?.['x-ms-long-running-operation-options']?.['final-state-via']);
-        (<go.LROPageableMethod>method).nextLinkName = op.language.go!.paging.nextLinkName;
+        method.finalStateVia = (op.extensions?.['x-ms-long-running-operation-options']?.['final-state-via']);
+        method.nextLinkName = op.language.go!.paging.nextLinkName;
         if (op.language.go!.paging.nextLinkOperation) {
           // adapt the next link operation
           const nextPageMethod = new go.NextPageMethod(op.language.go!.paging.nextLinkOperation.language.go.name, client, httpPath, httpMethod, getStatusCodes(op.language.go!.paging.nextLinkOperation));
           populateMethod(op.language.go!.paging.nextLinkOperation, nextPageMethod, m4CodeModel, codeModel);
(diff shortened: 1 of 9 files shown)


packages/codemodel.go/src/client.ts
```diff
@@ -10,155 +10,145 @@ import * as param from './param.js';
...
   parameters: Array<param.Parameter>;
 
-  // all the constructors for this client, can be empty
+  /** all the constructors for this client, can be empty */
   constructors: Array<Constructor>;
 
-  // contains client methods. can be empty
-  methods: Array<Method | LROMethod | PageableMethod | LROPageableMethod>;
+  /** contains client methods. can be empty */
+  methods: Array<MethodType>;
```
gracewilcox, line 31: Does switching to just using MethodType break anything? Not familiar with the downstream consequences

packages/codemodel.go/src/client.ts
```diff
@@ -176,7 +166,56 @@ export function newClientOptions(modelType: pkg.CodeModelType, clientName: strin
...
 ///////////////////////////////////////////////////////////////////////////////////////////////////
 
-export class Method implements Method {
+interface LROMethodBase extends MethodBase {
+  finalStateVia?: FinalStateVia;
+
+  operationLocationResultPath?: string;
+}
+
+interface MethodBase {
```
gracewilcox, line 175: What's the advantage of switching from Method to MethodBase?

packages/autorest.go/src/m4togocodemodel/clients.ts
```diff
@@ -34,27 +34,27 @@ export function adaptClients(m4CodeModel: m4.CodeModel, codeModel: go.CodeModel)
```
gracewilcox, line 34: Is everything being switched to use go.MethodType? See line 32 still uses the old way

packages/autorest.go/src/m4togocodemodel/clients.ts
```diff
@@ -34,27 +34,27 @@ export function adaptClients(m4CodeModel: m4.CodeModel, codeModel: go.CodeModel)
...
-        (<go.PageableMethod>method).nextLinkName = op.language.go!.paging.nextLinkName;
+        method.nextLinkName = op.language.go!.paging.nextLinkName;
         if (op.language.go!.paging.nextLinkOperation) {
           // adapt the next link operation
           const nextPageMethod = adaptNextPageMethod(op, m4CodeModel, client, codeModel);
-          (<go.PageableMethod>method).nextPageMethod = nextPageMethod;
+          method.nextPageMethod = nextPageMethod;
         }
       } else {
         method = new go.Method(op.language.go!.name, client, httpPath, httpMethod, getStatusCodes(op), naming);
```
gracewilcox, line 60: go.Method?

gracewilcox (commented):
Sorry, just a little confused on the change. Is everything being moved from using go.Method to go.MethodType? Still see some instances in the code base when it's not switched over

packages/autorest.go/src/m4togocodemodel/clients.ts
gracewilcox: joel fixed in 861d4fc018912c79c8db2693be8c5093666d7f86

PR #1601, opened 2025-06-23
Fix header retrival logic to be case-sensitive

### LINKED ISSUE #1600: Metadata headers should be handled in a case-sensitive manner
I'm working with the azblob package to set metadata on a blob. The keys I use are all lowercase. However, when I use blobClient.GetProperties to retrieve a key that is set in lowercase in the code and on the platform, it shows up with a single upper-case letter instead.

This was a known issue in 2017, where it was fixed in the previous iteration of the Azure/azure-sdk-for-go client: See the [issue](https://github.com/Azure/azure-sdk-for-go/issues/645) and [its PR fix](https://github.com/Azure/azure-sdk-for-go/pull/646/files).

The fix was to change Header.Add with using assignment directly to the header map, and to replace Header.Get with accessing the header map. This is the correct way to obtain case-sensitive information from the header. The [official Go documentation](https://pkg.go.dev/net/http#Header.Get) states:

> It is case insensitive... Get assumes that all keys are stored in canonical form. To use non-canonical keys, access the map directly.

And there is similar documentation for [Set](https://pkg.go.dev/net/http#Header.Set). In fact, SetMetadata does directly set the metadata header onto the map, through [this code generation part](https://github.com/Azure/autorest.go/blob/06875ee8644103d768b461aabb34c88fc74e9e86/packages/codegen.go/src/operations.ts#L726), and that does correctly preserve case information when sending it to the API endpoint.

I recommend this fix: whereever there are case-sensitive metadata headers prefixed with `x-ms-meta-`, use the Header map directly. In most if not all of those cases, the metadata is already being enumerated and the key is guaranteed.

I believe I've found where this is [retrieved from the headers](https://github.com/Azure/autorest.go/blob/06875ee8644103d768b461aabb34c88fc74e9e86/packages/codegen.go/src/operations.ts#L250), for example in the generated blob client's [`GetProperties" method](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/generated/zz_blob_client.go#L1576). Fixing this should fix the issue completely.

lbordowitz:
Fixes #1600

### DIFF
--- a/packages/autorest.go/test/acr/azacr/zz_containerregistry_client.go
+++ b/packages/autorest.go/test/acr/azacr/zz_containerregistry_client.go
@@ -118,17 +118,20 @@ func (client *ContainerRegistryClient) createManifestCreateRequest(ctx context.C
 // createManifestHandleResponse handles the CreateManifest response.
 func (client *ContainerRegistryClient) createManifestHandleResponse(resp *http.Response) (ContainerRegistryClientCreateManifestResponse, error) {
 	result := ContainerRegistryClientCreateManifestResponse{}
-	if val := resp.Header.Get("Content-Length"); val != "" {
+	if vals, ok := resp.Header["Content-Length"]; ok && len(vals) > 0 && vals[0] != "" {
+		val := vals[0]
 		contentLength, err := strconv.ParseInt(val, 10, 64)
 		if err != nil {
 			return ContainerRegistryClientCreateManifestResponse{}, err
 		}
 		result.ContentLength = &contentLength
 	}
-	if val := resp.Header.Get("Docker-Content-Digest"); val != "" {
+	if vals, ok := resp.Header["Docker-Content-Digest"]; ok && len(vals) > 0 && vals[0] != "" {
+		val := vals[0]
 		result.DockerContentDigest = &val
 	}
-	if val := resp.Header.Get("Location"); val != "" {
+	if vals, ok := resp.Header["Location"]; ok && len(vals) > 0 && vals[0] != "" {
+		val := vals[0]
 		result.Location = &val
 	}
 	body, err := runtime.Payload(resp)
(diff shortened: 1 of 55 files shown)


jhendrixMSFT:
@lbordowitz thanks for adding this, I'll take a look shortly.

Note that when making a change to generated code, you'll need to rebuild all of the test suites so that the codegen changes are picked up. We have some docs on how to do this [here](https://github.com/Azure/autorest.go/blob/main/docs/development.md#step-3-regenerate-tests-and-samples) which should clean up the CI failures.

jhendrixMSFT:
A little more context/use-case would be useful. I know that usually we don't call `Header.Set()` for casing reasons, so it could be the case that we simply missed it for `x-ms-meta-` (i.e. I don't recall if this was intentional or not).

Storage does have some weirdness about how the metadata headers are handled. So, my biggest concern is, if we make this change, will it break anybody.

Also, just FYI I'll be out-of-office for the remainder of this week, so we'll have to pick this up when I return on 6/30 (hopefully this isn't urgent).

lbordowitz:
I'm replacing deprecated Azure storage clients in `azurerm`, the Azure bucket backend of in OpenTofu, which is the open-source fork of Terraform. One part of how it works involves storing lock/lease information in the metadata. Whether or not that's a wise design decision, it's up to me to keep these backend versions compatible, such that someone using the new version and someone using the old version will be able maintain locks.

That means the metadata keys MUST be exactly the same. The current `azurerm` clients keep everything lowercase. I'm using azure-sdk-for-go in the new backend version, and it uses MIME casing. I need the metadata headers to be case-sensitive to make this compatible with the previous version.

jhendrixMSFT:
When receiving responses, the Go HTTP client will canonicalize the headers and there's no way to opt out of this behavior.  See https://github.com/golang/go/issues/37834 for more info.

PR #1616, opened 2025-07-01
Remove string from Scalar and into its own type

jhendrixMSFT:
Strings were special cased as they don't require any parsing or formatting like the other scalar types. Splitting them into their own type removes the need for special handling.

### DIFF
--- a/packages/codegen.go/src/example.ts
+++ b/packages/codegen.go/src/example.ts
@@ -369,7 +369,6 @@ function getPointerValue(type: go.PossibleType, valueString: string, byValue: bo
       case `bool`:
       case `byte`:
       case `rune`:
-      case `string`:
         prtType = 'Ptr';
         break;
       default:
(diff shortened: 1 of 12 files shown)


packages/codegen.go/src/example.ts
```diff
@@ -427,7 +429,6 @@ function generateFakeExample(goType: go.PossibleType, name?: string): go.Example
       case 'bool':
         return new go.BooleanExample(false, goType);
       case 'byte':
-      case 'string':
       case 'rune':
         return new go.StringExample(`<${name ?? 'test'}>`, goType);
```
RickWinter, line 433: Should this still return a stringexample?

RickWinter (approved):
Are there any tests covering this change? Nothing changed about the about which implies it was indeed all implementation detail.

packages/codegen.go/src/example.ts
jhendrixMSFT: Yeah `byte` and `rune` are still strings.

Issue #1620, opened 2025-07-03
[TypeSpec] The client name seems not deal with the stuttering


tadelesh:
See the breaking changes in this [PR](https://github.com/Azure/azure-rest-api-specs/pull/34320/).
`NotificationHubsClient` seems not deal with the stuttering removal.
![Image](https://github.com/user-attachments/assets/c2d28b0a-946f-4d7d-a5fa-99b56c8e36bc)

jhendrixMSFT:
I have a fix for this, but it might introduce breaking changes in previously released SDKs.

PR #1622, opened 2025-07-07
Fix for some stuttering client names

jhendrixMSFT:
Add the suffix "Client" to client names before checking for stuttering.

Fixes https://github.com/Azure/autorest.go/issues/1620


packages/typespec-go/src/tcgcadapter/adapter.ts
```diff
@@ -95,6 +95,11 @@ function fixStutteringTypeNames(sdkPackage: tcgc.SdkPackage<tcgc.SdkHttpOperatio
   // ensure that enum, client, and struct type names don't stutter
 
   const recursiveWalkClients = function(client: tcgc.SdkClientType<tcgc.SdkHttpOperation>): void {
+    // NOTE: we MUST do this before calling trimPackagePrefix to properly handle
+    // the case where the client name is the same as the package name.
+    if (!client.name.match(/Client$/)) {
```
RickWinter, line 100: What if the client name is already `FooClient`? We would match that and turn it into `FooClientClient`. We should check that the name doesn't already end in `Client`.

RickWinter (approved):

packages/typespec-go/src/tcgcadapter/adapter.ts
jhendrixMSFT: That's what this check does. If the name doesn't end with `Client` then append `Client`.

PR #1636, opened 2025-07-14
Execute steps post code generation

jhendrixMSFT:
Format and run the transforms program when available after successfully emitting code.
Regenerate some missing content.

### DIFF
--- a/packages/typespec-go/CHANGELOG.md
+++ b/packages/typespec-go/CHANGELOG.md
@@ -7,6 +7,13 @@
 * Fixed some cases where a client name could stutter.
 * Force the body paramter to be required for `PATCH` and `PUT` operations.
 
+### Features Added
+
+* When Go tools are found on the path, the following steps happen after successfully generating code.
+  * Execute `gofmt -w .`.
+  * If `/internal/generate/transforms.go` exists in the output directory, `go run` it from the output directory.
+  * If Go tools are not found, the above steps are skipped and a warning is displayed.
+
 ## 0.5.1 (2025-06-26)
 
 ### Other Changes
--- a/packages/typespec-go/src/emitter.ts
+++ b/packages/typespec-go/src/emitter.ts
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
 import 'source-map-support/register.js';
 import { reportDiagnostic } from './lib.js';
 import { CodegenError } from '../../codegen.go/src/errors.js';
+import { execSync } from 'child_process';
 
 export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
   try {
     await generate(context);
+
+    // probe to see if Go tools are on the path
+    try {
+      execSync('go version', { encoding: 'ascii' });
+    } catch {
+      context.program.reportDiagnostic({
+        code: 'GoVersion',
+        severity: 'warning',
+        message: 'skip executing post emitter steps (is go on the path?)',
+        target: NoTarget,
+      });
+
+      // no Go tools available so exit
+      return;
+    }
+
+    try {
+      execSync('gofmt -w .', { cwd: context.emitterOutputDir, encoding: 'ascii' });
+    } catch (err) {
+      context.program.reportDiagnostic({
+        code: 'gofmt',
+        severity: 'error',
+        message: (<Error>err).message,
+        target: NoTarget,
+      });
+
+      return;
+    }
+
+    // check for the transforms file.
+    // if it's there then "go run" it.
+    const transformsFile = `${context.emitterOutputDir}/internal/generate/transforms.go`;
+    if (!existsSync(transformsFile)) {
+      return;
+    }
+    try {
+      execSync(`go run ${transformsFile}`, { cwd: context.emitterOutputDir, encoding: 'ascii' });
+    } catch (err) {
+      context.program.reportDiagnostic({
+        code: 'transforms',
+        severity: 'error',
+        message: (<Error>err).message,
+        target: NoTarget,
+      });
+    }
   } catch (error) {
     if (error instanceof AdapterError) {
       reportDiagnostic(context.program, {
--- a/packages/typespec-go/.scripts/tspcompile.js
+++ b/packages/typespec-go/.scripts/tspcompile.js
@@ -294,7 +294,6 @@ function generate(moduleName, input, outputDir, perTestOptions) {
         logResult(error, stdout, stderr);
         // format on success
         if (error === null) {
-          execSync('gofmt -w .', { cwd: fullOutputDir});
           // Force emitter version to a constant in _metadata.json to avoid unnecessary version drift in committed files
           const metadataPath = `${fullOutputDir}/testdata/_metadata.json`;
           if (existsSync(metadataPath)) {
--- a/packages/typespec-go/test/azure-http-specs/azure/resource-manager/templatesgroup/fake/zz_optionalbody_server.go
+++ b/packages/typespec-go/test/azure-http-specs/azure/resource-manager/templatesgroup/fake/zz_optionalbody_server.go
@@ -26,7 +26,7 @@ type OptionalBodyServer struct {
 
 	// Patch is the fake for method OptionalBodyClient.Patch
 	// HTTP status codes to indicate success: http.StatusOK
-	Patch func(ctx context.Context, resourceGroupName string, widgetName string, options *templatesgroup.OptionalBodyClientPatchOptions) (resp azfake.Responder[templatesgroup.OptionalBodyClientPatchResponse], errResp azfake.ErrorResponder)
+	Patch func(ctx context.Context, resourceGroupName string, widgetName string, properties templatesgroup.Widget, options *templatesgroup.OptionalBodyClientPatchOptions) (resp azfake.Responder[templatesgroup.OptionalBodyClientPatchResponse], errResp azfake.ErrorResponder)
 
 	// Post is the fake for method OptionalBodyClient.Post
 	// HTTP status codes to indicate success: http.StatusOK
@@ -155,13 +155,7 @@ func (o *OptionalBodyServerTransport) dispatchPatch(req *http.Request) (*http.Re
 	if err != nil {
 		return nil, err
 	}
-	var options *templatesgroup.OptionalBodyClientPatchOptions
-	if !reflect.ValueOf(body).IsZero() {
-		options = &templatesgroup.OptionalBodyClientPatchOptions{
-			Properties: &body,
-		}
-	}
-	respr, errRespr := o.srv.Patch(req.Context(), resourceGroupNameParam, widgetNameParam, options)
+	respr, errRespr := o.srv.Patch(req.Context(), resourceGroupNameParam, widgetNameParam, body, nil)
 	if respErr := server.GetError(errRespr, req); respErr != nil {
 		return nil, respErr
 	}
--- a/packages/typespec-go/test/azure-http-specs/azure/resource-manager/templatesgroup/zz_optionalbody_client.go
+++ b/packages/typespec-go/test/azure-http-specs/azure/resource-manager/templatesgroup/zz_optionalbody_client.go
@@ -109,14 +109,15 @@ func (client *OptionalBodyClient) getHandleResponse(resp *http.Response) (Option
 // Generated from API version 2023-12-01-preview
 //   - resourceGroupName - The name of the resource group. The name is case insensitive.
 //   - widgetName - The name of the Widget
+//   - properties - The resource properties to be updated.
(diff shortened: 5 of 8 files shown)


packages/typespec-go/src/emitter.ts
```diff
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
...
 import { reportDiagnostic } from './lib.js';
 import { CodegenError } from '../../codegen.go/src/errors.js';
+import { execSync } from 'child_process';
 
 export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
   try {
     await generate(context);
+
+    // probe to see if Go tools are on the path
+    try {
```
jhendrixMSFT, line 38: I split this up into separate `try/catch` blocks so that each task has a different `code`. Maybe that's not so useful thought?

packages/typespec-go/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
 * Fixed some cases where a client name could stutter.
 * Force the body paramter to be required for `PATCH` and `PUT` operations.
 
+### Features Added
+
+* When Go tools are found on the path, the following steps happen after successfully generating code.
+  * Execute `gofmt -w .`.
+  * If `/internal/generate/transforms.go` exists in the output directory, `go run` it from the output directory.
```
chlowell, line 14: Couldn't this be in`/testdata`?

packages/typespec-go/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@
 * Fixed some cases where a client name could stutter.
 * Force the body paramter to be required for `PATCH` and `PUT` operations.
 
+### Features Added
+
+* When Go tools are found on the path, the following steps happen after successfully generating code.
+  * Execute `gofmt -w .`.
```
gracewilcox, line 13: Could we do `goimports -w .` instead? I know I have some packages that generate with broken inputs, so I always run go imports in my build.go file instead of go fmt

jhendrixMSFT: `goimports` isn't part of the standard Go toolset, so there's no guarantee it will be installed. While we _could_ `go install` it, I'd prefer to avoid that (this is already a slippery slope IMO, and having the emitter make network calls increases the grade).

Can you please let me know what imports are broken? If this is a codegen bug we should fix it.

jhendrixMSFT: Let me emphasize that we _must_ fix such bugs. Emitted SDKs should compile as-is without the need to execute `goimports`.

richardpark-msft: I generally use goimports because sometimes changes you make will make it so you no longer need some imports. 

I think it's okay if we say that it needs to be installed prior to the emitter running, similar to how we require Go be installed.

packages/typespec-go/src/emitter.ts
```diff
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
...
+        severity: 'warning',
+        message: 'skip executing post emitter steps (is go on the path?)',
+        target: NoTarget,
+      });
+
+      // no Go tools available so exit
+      return;
+    }
+
+    try {
```
gracewilcox, line 52: Would be better to have the format command as the last to run. Sometimes transforms can mess with the code, so the formatter needs to be run after the transforms.

packages/typespec-go/src/emitter.ts
jhendrixMSFT: Will leave as-is for now. Can revisit later.

packages/typespec-go/CHANGELOG.md
gracewilcox: > I generally use goimports because sometimes changes you make will make it so you no longer need some imports.
> 
> I think it's okay if we say that it needs to be installed prior to the emitter running, similar to how we require Go be installed.

Ditto on what Richard said. Some of the transforms can effect imports. Would be nice to have this tool run as part of the process.

gracewilcox: Sounds good to me!

packages/typespec-go/src/emitter.ts
```diff
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
...
   try {
     await generate(context);
+
+    // probe to see if Go tools are on the path
+    try {
+      execSync('go version', { encoding: 'ascii' });
+    } catch {
+      context.program.reportDiagnostic({
+        code: 'GoVersion',
+        severity: 'warning',
```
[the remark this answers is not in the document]
richardpark-msft: Agreed - once you've said it should be there it's a failure if it never runs.

packages/typespec-go/src/emitter.ts
```diff
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
...
+          severity: 'error',
+          message: (<Error>err).message,
+          target: NoTarget,
+        });
+      }
+    }
+
+    // format after transforms in case any formatting gets munged
+    try {
+      execSync('gofmt -w .', { cwd: context.emitterOutputDir, encoding: 'ascii' });
```
richardpark-msft, line 70: gofmt -w ./... perhaps?

(or goimports, whichever, but either way recursive)

packages/typespec-go/CHANGELOG.md
jhendrixMSFT: Fixed.

packages/typespec-go/src/emitter.ts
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

packages/typespec-go/src/emitter.ts
```diff
@@ -28,10 +28,56 @@ import { EmitContext, NoTarget } from '@typespec/compiler';
...
+      // no Go tools available so exit
+      return;
+    }
+
+    // check for the transforms file.
+    // if it's there then "go run" it.
+    const transformsFile = `${context.emitterOutputDir}/testdata/generate/transforms.go`;
+    if (existsSync(transformsFile)) {
+      try {
+        execSync(`go run ${transformsFile}`, { cwd: context.emitterOutputDir, encoding: 'ascii' });
```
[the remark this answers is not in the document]
jhendrixMSFT: I think this will also be part of the convention. It's assumed that the transform program is executed from the root of the SDK directory (this appears to be the case for the few that I've inspected).

packages/typespec-go/src/emitter.ts
jhendrixMSFT: This syntax will format all sources for the current module (it's the same syntax we use in `build.go` files).

packages/typespec-go/src/emitter.ts
```diff
@@ -24,14 +24,105 @@ import { generateMetadataFile } from '../../codegen.go/src/metadata.js';
...
+        });
+      }
+    }
+
+    // format after transforms in case any formatting gets munged.
+    // first try goimports which will also fix up any imports that
+    // might be stale due to transforms.
+
+    // probe to see if goimports is installed
+    let goImportsInstalled = false;
```
jhendrixMSFT, line 85: Copilot came up with this algorithm, so be sure to look extra close.

packages/typespec-go/src/emitter.ts
richardpark-msft: Oh, I didn't realize it was recursive by default!

jhendrixMSFT:
So this is a bummer. Here we can see that CI doesn't install `goimports` so we fall back to `gofmt`. However, since I have `goimports` locally installed, the formatting of the `import` blocks is slightly different and as a result, CI fails. This is starting to smell a bit.

richardpark-msft:
> So this is a bummer. Here we can see that CI doesn't install `goimports` so we fall back to `gofmt`. However, since I have `goimports` locally installed, the formatting of the `import` blocks is slightly different and as a result, CI fails. This is starting to smell a bit.

Is the solution to just install goimports? Or are we not allowed to do that in pipelines?

jhendrixMSFT:
We can. But now we have this kind of hidden behavior that you get a different result depending on if `goimports` is installed or not.

packages/typespec-go/src/emitter.ts
jhendrixMSFT: I don't think this is going to work as it's highly dependent on the shell being used.

jhendrixMSFT:
I've also added `go mod tidy` to the list of commands to execute.

I did _not_ add `go get -u` as this can fail in some cases.

packages/typespec-go/src/emitter.ts
```diff
@@ -24,14 +24,88 @@ import { generateMetadataFile } from '../../codegen.go/src/metadata.js';
...
 import 'source-map-support/register.js';
 import { reportDiagnostic } from './lib.js';
 import { CodegenError } from '../../codegen.go/src/errors.js';
+import { execSync } from 'child_process';
 
 export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
   try {
     await generate(context);
+
+    const afterGenerateFileExists = existsSync(`${context.emitterOutputDir}/after_generate.go`);
```
RickWinter, line 37: Rather than trigger based on a file. Can we use the tspconfig.yaml?
i.e.
From this file, https://github.com/Azure/azure-rest-api-specs/blob/f2acb480cbbd549eb2aafff1372dbddf0f8da472/specification/eventgrid/Azure.Messaging.EventGrid.SystemEvents/tspconfig.yaml, We'd add a `post-generate-step` option?
```
  "@azure-tools/typespec-go":
    module: "github.com/Azure/azure-sdk-for-go/{service-dir}/{package-dir}"
    service-dir: "sdk/messaging/eventgrid"
    package-dir: "azsystemevents"
    module-version: "0.0.1"
    emitter-output-dir: "{project-root}"
    slice-elements-byval: true
    flavor: "azure"
```

RickWinter (commented):
Should we use the tspconfig.yaml to trigger the post processing?

packages/typespec-go/src/emitter.ts
jhendrixMSFT: That does have the side-effect of making it explicit which IMO is good. However, there is a one-time cost to go and update the necessary `tspconfig.yaml` files with this info.

RickWinter (approved):

packages/typespec-go/src/emitter.ts
```diff
@@ -63,9 +64,9 @@ export async function $onEmit(context: EmitContext<GoEmitterOptions>) {
     }
 
     // if we have a post-generation transforms file then "go generate" it
-    if (afterGenerateFileExists) {
+    if (goGenerateFileExists) {
       try {
-        execSync('go generate', { cwd: context.emitterOutputDir, encoding: 'ascii' });
+        execSync(`go generate ${goGenerateFile}`, { cwd: context.emitterOutputDir, encoding: 'ascii' });
       } catch (err) {
         context.program.reportDiagnostic({
```
gracewilcox, line 71: Does the error message return the `goGenerateFile` name? Putting the wrong file path could be a common stumbling block, and returning the file name would be useful information.

packages/typespec-go/CHANGELOG.md
```diff
@@ -1,12 +1,21 @@
...
 
 ### Breaking Changes
 
 * Fixed some cases where a client name could stutter.
 * Force the body paramter to be required for `PATCH` and `PUT` operations.
 
+### Features Added
+
+* Added switch `go-generate` to invoke post-generation scripts.
+  * The value is an output-relative path to a `.go` file that will can be used with `go generate`.
```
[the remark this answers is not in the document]
jhendrixMSFT: fixed

packages/typespec-go/src/emitter.ts
jhendrixMSFT: This has been updated so that if `go-generate` is specified and the file isn't found an error is returned including the full path to the file (great idea).

packages/typespec-go/src/lib.ts
```diff
@@ -51,6 +52,11 @@ const EmitterOptionsSchema: JSONSchemaType<GoEmitterOptions> = {
       nullable: true,
       description: 'When true, enables generation of fake servers. The default is false.',
     },
+    'go-generate': {
+      type: 'string',
+      nullable: true,
+      description: `Path to a post-generation 'go generate' script. The path is relative to the directory where the SDK is emitted.`,
+    },
```
gracewilcox, line 59: Is "the directory where the SDK is emitted" the same as the `emitter-output-dir`? If so, it might be good to explicitly use that terminology.

jhendrixMSFT: Fixed

PR #1657, opened 2025-07-21
Fixes for raw JSON usage

jhendrixMSFT:
Added missing import of the `io` package in some cases. Removed unused `var err error` in some implementations of UnmarshalJSON.


packages/codegen.go/src/models.ts
```diff
@@ -443,87 +449,110 @@ function generateJSONUnmarshallerBody(modelType: go.Model | go.PolymorphicModel,
...
-    } else if (field.type.kind === 'time') {
-      unmarshalBody += `\t\t\t\terr = unpopulate${capitalize(field.type.format)}(val, "${field.name}", &${receiver}.${field.name})\n`;
-      modelDef.SerDe.needsJSONUnpopulate = true;
-    } else if (field.type.kind === 'slice' && field.type.elementType.kind === 'time') {
-      imports.add('time');
-      let elementPtr = '*';
-      if (field.type.elementTypeByValue) {
-        elementPtr = '';
+
+  const emitSwitchCase = function(): string {
```
jhendrixMSFT, line 456: This change just stuffs the original code into this helper function and sets `needsErrCheck` as required. The rest of the code is untouched.

RickWinter (approved):

PR #1688, opened 2025-08-22
Report NameCollision diagnostic for duplicate const values

jhendrixMSFT:
This will require the tsp author to explicitly define the names.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -610,13 +610,27 @@ export class typeAdapter {
       if (valueType.enumType.access === 'internal') {
         valueTypeName = naming.getEscapedReservedName(uncapitalize(valueTypeName), 'Type');
       }
-      let value = this.constValues.get(valueTypeName);
-      if (!value) {
-        value = new go.ConstantValue(valueTypeName, type, valueType.value);
-        value.docs.summary = valueType.summary;
-        value.docs.description = valueType.doc;
-        this.constValues.set(valueTypeName, value);
+
+      if (this.constValues.has(valueTypeName)) {
+        // we have a type name collision. while tsp will prevent this
+        // from directly being authored, it can happen due to how names
+        // are computed. consider the following example.
+        //
+        // union ContentTypes { "application/json" | "application/*+json" }
+        //
+        // naming.ensureNameCase() will discard the "*+" chars in the
+        // second value. as a result, both entries will have the name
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
This was observed in AppConfiguration: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/appconfiguration/AppConfiguration/routes.tsp#L152-L158

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -610,13 +610,27 @@ export class typeAdapter {
...
+        // are computed. consider the following example.
+        //
+        // union ContentTypes { "application/json" | "application/*+json" }
+        //
+        // naming.ensureNameCase() will discard the "*+" chars in the
+        // second value. as a result, both entries will have the name
+        // "ApplicationJSON".
+        // instead of being clever, report a NameCollision diagnostic so
+        // that the tsp author can define the desired names.
+        throw new AdapterError('NameCollision', `enum ${type.name} contains entry ${valueTypeName} that collides with an existing value`, valueType.__raw?.node ?? tsp.NoTarget);
```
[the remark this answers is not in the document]
jhendrixMSFT: `valueTypeName` is the colliding name.

packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -610,13 +610,27 @@ export class typeAdapter {
...
-        value.docs.description = valueType.doc;
-        this.constValues.set(valueTypeName, value);
-      }
+
+      if (this.constValues.has(valueTypeName)) {
+        // we have a type name collision. while tsp will prevent this
+        // from directly being authored, it can happen due to how names
+        // are computed. consider the following example.
+        //
+        // union ContentTypes { "application/json" | "application/*+json" }
```
jhendrixMSFT, line 619: We could consider replacing a `*` with the moniker `Wildcard` which would also resolve the issue (we'd still test for collisions and error out of course).

RickWinter (approved):
The change is reasonable to handle the AppConfig scenario...

It's probably better that we throw an error any time we trim the name. I'd prefer renaming is an explicit override in the client.tsp

packages/typespec-go/src/tcgcadapter/types.ts
jhendrixMSFT: Will leave the behavior as-is for now.

RickWinter (approved):

Issue #1693, opened 2025-08-28
should not generate interface for specific response model


JiaqiZhang-Dev:
PR: https://github.com/Azure/azure-rest-api-specs/pull/35817

tsp:
<img width="1702" height="576" alt="Image" src="https://github.com/user-attachments/assets/d87f1403-ed69-48d0-80ad-c366b6440473" />

changelog:
<img width="1419" height="174" alt="Image" src="https://github.com/user-attachments/assets/feb352f0-c7c3-4b54-9ac1-2957702a6e94" />
<img width="1514" height="191" alt="Image" src="https://github.com/user-attachments/assets/492e9558-009f-4fc9-985e-c3ece309282d" />

According to the tsp, the response body is a concrete model `OperationJobExtendedInfo`, but emitter generated code is `OperationExtendedInfoClassification`, which is an interface.

PR #1694, opened 2025-08-29
Fix polymorphic result & Upgrade tsp toolset

JiaqiZhang-Dev:
resolved: https://github.com/Azure/autorest.go/issues/1693

### DIFF
--- a/packages/codemodel.go/src/result.ts
+++ b/packages/codemodel.go/src/result.ts
@@ -103,7 +103,7 @@ export interface ModelResult {
   docs: type.Docs;
 
   /** the type returned in the response envelope */
-  modelType: type.Model;
+  modelType: type.Model | type.PolymorphicModel;
 
   /** the format in which the result is returned */
   format: ModelResultFormat;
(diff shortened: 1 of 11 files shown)


packages/codemodel.go/src/result.ts
```diff
@@ -103,7 +103,7 @@ export interface ModelResult {
   docs: type.Docs;
 
   /** the type returned in the response envelope */
-  modelType: type.Model;
+  modelType: type.Model | type.PolymorphicModel;
```
jhendrixMSFT, line 106: I don't think this is correct as we have `PolymorphicResult` type which is used for polymorphic results.

jhendrixMSFT (changes requested):

packages/codemodel.go/src/result.ts
jhendrixMSFT: Ok I get it now. Can you please update the comment accordingly so it's clear when this will be a `PolymorphicModel`?

JiaqiZhang-Dev: updated

packages/typespec-go/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@
 
 * Updated the minimum version of `azcore` to `v1.19.0`.
 * Report a `NameCollision` diagnostic if two `const` values have the same name.
+* Updated to the latest tsp toolset.
+
+### Bugs Fixed
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: updated

jhendrixMSFT (approved):

PR #1712, opened 2025-09-19
Refactor method types in code model

jhendrixMSFT:
Added definition for vanilla Go methods and their parts (parameters and receivers).  The HTTP methods and parameters were updated to extend these types.  As a result, the client field was renamed to receiver and the responseEnvelope file was renamed to returns.
Renamed synchronous operation method to SyncMethod so it doesn't clash with the vanilla Method type.
Removed Client parameter from some codegen methods as it's not required (the method type contains this info).
Consolidated emitting a method's receiver signature to a shared helper.

### DIFF
--- a/packages/codemodel.go/src/client.ts
+++ b/packages/codemodel.go/src/client.ts
@@ -5,10 +5,14 @@
 
 import { CodeModelError } from './errors.js';
 import { MethodExample } from './examples.js';
+import * as method from './method.js';
 import * as pkg from './package.js';
 import * as param from './param.js';
 import * as result from './result.js';
 import * as type from './type.js';
+
+// temporary until more param types refactor
+export { Receiver } from './method.js';
 
 /** an SDK client */
 export interface Client {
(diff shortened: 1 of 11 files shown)


packages/codemodel.go/src/client.ts
```diff
@@ -221,40 +219,38 @@ interface MethodBase {
...
   parameters: Array<param.MethodParameter>;
 
-  /** the method options type for this methoid */
+  /** the method options type for this method */
   optionalParamsGroup: param.ParameterGroup;
 
-  /** the response type for this method */
-  responseEnvelope: result.ResponseEnvelope;
+  /** the method's return type */
+  returns: result.ResponseEnvelope;
```
RickWinter, line 226: Nit: I'm not a huge fan of the name `returns`... Something about it rubs me, but best I could think of was `returnType`.

RickWinter: On the same note it is a ResponseEnvelope so responseEnvelope is reasonable name...

RickWinter (approved):

packages/codemodel.go/src/client.ts
jhendrixMSFT: The `returns` field comes from the base `Method` type and in those cases, the return type isn't a response envelope type.  This is the same pattern we use in the Rust code model (I'm trying to align parts where it makes sense, so if you work on one, the other should feel familiar).

packages/codemodel.go/src/client.ts
```diff
@@ -161,7 +165,7 @@ export interface NextPageMethod {
   httpStatusCodes: Array<number>;
 
   /** the client to which the method belongs */
```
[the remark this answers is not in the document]
jhendrixMSFT: I tweaked the comment a bit.  For context, in Go, the receiver param is (somewhat) analogous to the `this` param in C++ except that the type/name is explicit.

RickWinter (approved):

Issue #1575, opened 2025-05-26
Support optional path parameter


JiaqiZhang-Dev:
Only used for key vault for which Go uses post processing to support it. Emitter should report error if spec uses this feature.

PR #1727, opened 2025-09-30
Fix handling of optional path parameters

jhendrixMSFT:
Added Spector test coverage.

Fixes https://github.com/Azure/autorest.go/issues/1575

### DIFF
--- a/packages/codegen.go/src/operations.ts
+++ b/packages/codegen.go/src/operations.ts
@@ -761,7 +761,7 @@ function createProtocolRequest(azureARM: boolean, method: go.MethodType | go.Nex
     // replace path parameters
     for (const pp of methodParamGroups.pathParams) {
       let paramValue: string;
-      let optionalPathSep = '';
+      let optionalPathSep = false;
       if (pp.style !== 'optional') {
         // emit check to ensure path param isn't an empty string
         if (pp.kind === 'pathScalarParam') {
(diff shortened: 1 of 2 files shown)


packages/codegen.go/src/operations.ts
```diff
@@ -738,42 +738,95 @@ function createProtocolRequest(azureARM: boolean, method: go.MethodType | go.Nex
...
+    }
+    let client = '';
+    if (param.location === 'client') {
+      client = 'client.';
+    }
+    const paramGroupName = uncapitalize(param.group.name);
+    let optionalParamGroupCheck = `${client}${paramGroupName} != nil && `;
+    if (param.group.required) {
+      optionalParamGroupCheck = '';
+    }
```
RickWinter, line 754: Nit: maybe a meaningless perf optimization. Could set to '' and the only do the larger replacement if !pararm.group.required

packages/typespec-go/test/http-specs/parameters/pathgroup/testdata/_metadata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "emitterVersion": "0.0.0"
+}
```
RickWinter, line 3: New line is missing

RickWinter (approved):

Issue #1594, opened 2025-06-16
[TypeSpec] Binary payload handle for streaming


tadelesh:
For current TypeSpec streaming, TCGC falls back to binary input and output. But the Go codegen seems fail to handle the request binary. See generated code [here](https://github.com/Azure/autorest.go/blob/85e8f7bdc505ddb51b53189e4700fa5d7f2d406b/packages/typespec-go/test/http-specs/streaming/jsonlgroup/zz_jsonlbasic_client.go). The problem is from this [test](https://github.com/Azure/autorest.go/pull/1566/#discussion_r2135354356).

RickWinter:
@jhendrixMSFT Check for duplicate that Laurent mayhave opened, update cost on the primary issue

tadelesh:
The issue is not talking about fully support steaming in TypeSpec. Currently, TCGC will fallback any streaming tsp to just binary input/output. But it seems current Go emitter could not handle binary input/output. Is that because of the content type?

jhendrixMSFT:
Yeah the `application/jsonl` is what's tripping us up here.

This is going to need some design discussion for handling `jsonl`.

RickWinter:
@tadelesh is this needed for the migration work?

PR #1729, opened 2025-10-03
[spector] add tests for http-specs pageable

gracewilcox:
Adding missing tests for spector under http-specs pageable


packages/typespec-go/.scripts/tspcompile.js
```diff
@@ -47,6 +48,7 @@ const httpSpecsGroup = {
   'condreqgroup': ['special-headers/conditional-request'],
   //'repeatabilitygroup': ['special-headers/repeatability'],   // requires union support
   'specialwordsgroup': ['special-words'],
+  'jsonlgroup': ['streaming/jsonl'],
```
jhendrixMSFT, line 51: FYI this produces incorrect codegen and is tracked in https://github.com/Azure/autorest.go/issues/1594

jhendrixMSFT (approved):

PR #1733, opened 2025-10-06
[spector] add http-specs tests

gracewilcox:
Adding spector tests for:
 * http-specs/type/property/nullable/collectionstring
 * http-specs/type/optional/plaindate


gracewilcox:
@microsoft-github-policy-service rerun

PR #1735, opened 2025-10-07
Fix doc for client options param

jhendrixMSFT:
The tsp emitter was using the doc string for the type. This consolidates the doc string.

No functional changes.

### DIFF
--- a/packages/codegen.go/src/operations.ts
+++ b/packages/codegen.go/src/operations.ts
@@ -322,7 +322,7 @@ function generateConstructors(client: go.Client, imports: ImportManager): string
 
     // add client options last
     ctorParams.push(`options ${helpers.formatParameterTypeName(clientOptions)}`);
-    paramDocs.push(helpers.formatCommentAsBulletItem('options', clientOptions.docs));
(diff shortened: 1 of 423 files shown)


packages/codegen.go/src/operations.ts
```diff
@@ -322,7 +322,7 @@ function generateConstructors(client: go.Client, imports: ImportManager): string
 
     // add client options last
     ctorParams.push(`options ${helpers.formatParameterTypeName(clientOptions)}`);
-    paramDocs.push(helpers.formatCommentAsBulletItem('options', clientOptions.docs));
+    paramDocs.push(helpers.formatCommentAsBulletItem('options', { summary: 'Contains optional client configuration. Pass nil to accept the default values.' }));
```
RickWinter, line 325: We lose the link to the type with this change. Won't that make it harder to see what options are available? I like the new syntax but would be nice if we had a link to the *Options type.

//   - options - ### AccessClientOptions contains the optional values for creating a [AccessClient]

jhendrixMSFT: The param types in the function signature will automatically link to the type.  Here's an example.

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7#ClientFactory

Issue #1718, opened 2025-09-23
Support grouped params via override


jhendrixMSFT:
https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/client-generator-core/override/client.tsp#L22-L27

Groups params into a struct. At present, this test causes bad codegen.

PR #1756, opened 2025-10-27
Support @override spector case

JiaqiZhang-Dev:
Related issue: https://github.com/Azure/autorest.go/issues/1718
Spector test: https://github.com/Azure/typespec-azure/blob/main/packages/azure-http-specs/specs/azure/client-generator-core/override/client.tsp

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/clients.ts
+++ b/packages/typespec-go/src/tcgcadapter/clients.ts
@@ -551,6 +551,15 @@ export class clientAdapter {
     // track parameter groups created from model types
     const parameterGroups = new Map<string, go.ParameterGroup>();
 
+    // Helper function to add a parameter to method.parameters and paramMapping
+    const addParameterToMethod = (adaptedParam: go.MethodParameter, opParam: OperationParamType) => {
+      method.parameters.push(adaptedParam);
+      if (!paramMapping.has(opParam)) {
+        paramMapping.set(opParam, new Array<go.MethodParameter>());
+      }
+      paramMapping.get(opParam)?.push(adaptedParam);
+    };
+
     // we must enumerate parameters, not operation.parameters, as it
     // contains the params in tsp order as well as any spread params.
     for (const param of sdkMethod.parameters) {
@@ -705,13 +714,8 @@ export class clientAdapter {
             adaptedPropertyParam.docs.summary = property.summary;
             adaptedPropertyParam.docs.description = property.doc;
             adaptedPropertyParam.group = paramGroup;
-            method.parameters.push(adaptedPropertyParam);
+            addParameterToMethod(adaptedPropertyParam, propertyOpParam);
             paramGroup.params.push(adaptedPropertyParam);
-            
-            if (!paramMapping.has(propertyOpParam)) {
-              paramMapping.set(propertyOpParam, new Array<go.MethodParameter>());
-            }
-            paramMapping.get(propertyOpParam)?.push(adaptedPropertyParam);
           }
           continue; // Skip regular parameter handling
         }
@@ -721,11 +725,7 @@ export class clientAdapter {
 
       adaptedParam.docs.summary = param.summary;
       adaptedParam.docs.description = param.doc;
-      method.parameters.push(adaptedParam);
-      if (!paramMapping.has(opParam)) {
-        paramMapping.set(opParam, new Array<go.MethodParameter>());
-      }
-      paramMapping.get(opParam)?.push(adaptedParam);
+      addParameterToMethod(adaptedParam, opParam);
 
       if (adaptedParam.style !== 'required' && adaptedParam.style !== 'literal') {
         // add optional method param to the options param group


packages/typespec-go/src/tcgcadapter/clients.ts
tadelesh: Do you think these logic could be covered by line 532 - 538?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -561,22 +564,80 @@ export class clientAdapter {
...
+
+        for (const property of modelProperties) {
+          const propertyOpParam = values(allOpParams).where((op: OperationParamType) => {
+            return values(op.correspondingMethodParams).where((methodParam: tcgc.SdkModelPropertyType | tcgc.SdkMethodParameter) => {
               return methodParam === property;
             }).any();
           }).first();
-          if (opParam) {
+
+          if (!propertyOpParam || propertyOpParam.kind === 'body') {
```
jhendrixMSFT: It's unclear why we need this block. Since line 572 evaluated to true, isn't it guaranteed that this is a parameter group?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -561,22 +564,80 @@ export class clientAdapter {
...
+          const isRequired = go.isRequiredParameter(paramStyle);
+          
+          // Use the same naming approach as regular parameters for consistency
+          const paramGroupName = ensureNameCase(param.type.name);
+          const paramName = getEscapedReservedName(ensureNameCase(param.name, isRequired), 'Param');
+          
+          // Remove the model from codeModel.models if it is a parameter group
+          const modelIndex = this.ta.codeModel.models.findIndex(m => m.name === paramGroupName);
+          if (modelIndex >= 0) {
+            this.ta.codeModel.models.splice(modelIndex, 1);
```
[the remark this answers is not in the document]
tadelesh: The function of `@override` does not limit the rewrite of the parameters, thus TCGC does not know the parameter group. TCGC just ensures that all the HTTP parameters have the corresponding method level parameters.

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -561,22 +564,80 @@ export class clientAdapter {
...
-      if (!opParam && param.type.kind === 'model') {
-        for (const property of param.type.properties) {
-          opParam = values(allOpParams).where((opParam: OperationParamType) => {
-            return values(opParam.correspondingMethodParams).where((methodParam: tcgc.SdkModelPropertyType | tcgc.SdkMethodParameter) => {
+      if (!opParam) {
+        throw new AdapterError('InternalError', `didn't find operation parameter for method ${sdkMethod.name} parameter ${param.name}`, sdkMethod.__raw?.node ?? NoTarget);
+      }
+
+      // Check if it is a parameter group
+      if (param.type.kind === 'model' && opParam.kind !== 'body') {
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: done

packages/typespec-go/src/tcgcadapter/clients.ts
JiaqiZhang-Dev: The line 572 just attempt to handle as a parameter group for non-body parameter, if any property doesn't map to HTTP param it's not a parameter group. I removed the `isParameterGroup` flag since it's unclear.

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -636,6 +625,75 @@ export class clientAdapter {
...
+          const paramGroupName = ensureNameCase(param.type.name);
+          const paramName = getEscapedReservedName(ensureNameCase(param.name, isRequired), 'Param');
+          
+          // Remove the model from codeModel.models if it is a parameter group
+          const modelIndex = this.ta.codeModel.models.findIndex(m => m.name === paramGroupName);
+          if (modelIndex >= 0) {
+            this.ta.codeModel.models.splice(modelIndex, 1);
+          }
+          
+          // Check if parameter group already exists
```
jhendrixMSFT: When would we have a case where a parameter group already exists?  Can multiple `sdkMethod.parameters` map to the same parameter group?

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -636,6 +625,75 @@ export class clientAdapter {
...
+        // Validate all properties map to HTTP params
+        for (const property of modelProperties) {
+          const propertyOpParam = values(allOpParams).where((op: OperationParamType) => {
+            return values(op.correspondingMethodParams).where((methodParam: tcgc.SdkModelPropertyType | tcgc.SdkMethodParameter) => {
+              return methodParam === property;
+            }).any();
+          }).first();
+
+          if (!propertyOpParam) {
+            // If any property doesn't map to HTTP param it's not a parameter group
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: This condition is not enough to guarantee it is parameter group, since the param is possible a body parameter. This is an example: https://github.com/microsoft/typespec/blob/main/packages/http-specs/specs/type/model/visibility/main.tsp#L64;

The param is decorated by @body/@bodyRoot, and the opParam is queryProp, which means `param.type.kind === 'model'` and `opParam.kind === 'query'` (met the condition), but the param is not a parameter group.

packages/typespec-go/src/tcgcadapter/clients.ts
JiaqiZhang-Dev: This is an example case, two parameters used same parameter group:

```typespec
model GroupParametersOptions {
  @query param1: string;
  @query param2: string;
}

op groupCustomized(options1: GroupParametersOptions, options2: GroupParametersOptions): void;
```

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -636,6 +625,75 @@ export class clientAdapter {
...
+            paramGroup.docs.summary = `${paramGroupName} contains a group of parameters for the ${method.receiver.type.name}.${method.name} method.`;
+            parameterGroups.set(paramGroupName, paramGroup);
+          }
+
+          // Add each property as a method parameter and associate with the group
+          for (let i = 0; i < modelProperties.length; i++) {
+            const property = modelProperties[i];
+            const propertyOpParam = correspondingOpParams[i];
+            const adaptedPropertyParam = this.adaptMethodParameter(propertyOpParam, method.httpMethod);
+            adaptedPropertyParam.docs.summary = property.summary;
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: I added a  function-local closure `addParameterToMethod` to reduce the duplication.

Issue #1752, opened 2025-10-23
Nameing breaking change for armiothub


JiaqiZhang-Dev:
This PR has some spell naming breaking changes: https://github.com/Azure/azure-sdk-for-go/pull/25430/files#r2453708272.

<img width="724" height="526" alt="Image" src="https://github.com/user-attachments/assets/a5107da6-a853-4e1e-b2e2-7fb6564a653b" />

According to emitter's naming logic, this kind of field name should always be xxxISO, could you help check on this case and whether we could accept this breaking changes? Thanks! https://github.com/Azure/autorest.go/blob/dc53827b71625a944b616ddbbaf68197e68ca225/packages/naming.go/src/mappings.ts#L28

PR #1759, opened 2025-10-30
Remove Iso from the global list of common acronyms

jhendrixMSFT:
Add it for the tsp emitter to avoid breaking changes in autorest generated code.

Fixes https://github.com/Azure/autorest.go/issues/1752


packages/autorest.go/CHANGELOG.md
jhendrixMSFT: Waiting on confirmation that this doesn't introduce any breaking changes in released SDKs before merging.

chlowell (approved):

tadelesh (approved):

Issue #1798, opened 2025-12-12
tsp: support lropaging methods


jhendrixMSFT:
Add support for paged lro methods in the tsp emitter (missing in adapter).

JiaqiZhang-Dev:
blocked two migration PRs
https://github.com/Azure/azure-rest-api-specs-pr/pull/25860
https://github.com/Azure/azure-rest-api-specs/pull/36687

PR #1801, opened 2025-12-15
Add tsp support for pageable LROs

jhendrixMSFT:
It's supported by the code model but was never implemented for tsp. Added an ARM test for generated code.
Updated workspace dependencies.

Fixes https://github.com/Azure/autorest.go/issues/1798


packages/typespec-go/test/tsp/PageableLROs/main.tsp
jhendrixMSFT: This was adapted from one of the pending specs PRs.

chlowell (approved):

Issue #1772, opened 2025-11-19
[http-client-go] Multiples services for one client Implementation


lmazuel:

lirenhe:
Chenjie already started the integration of TCGC into Go emitter.

lirenhe:
Reopen this item with the new requirement that we need to support creating the operation group from services with operations in different api versions.

jhendrixMSFT:
This is now complete.

PR #1810, opened 2025-12-19
Multiple service

tadelesh:
resolve: https://github.com/Azure/autorest.go/issues/1772

No emitter code change. Just upgrade TCGC version and add new tests.


packages/typespec-go/test/azure-http-specs/azure/resource-manager/armmultipleservicegroup/utils_test.go
jhendrixMSFT: This can be simplified, see https://github.com/Azure/autorest.go/blob/main/packages/typespec-go/test/azure-http-specs/azure/resource-manager/methodsubscriptionidgroup/utils_test.go

lirenhe (approved):

PR #1821, opened 2026-01-13
[typespec] update dependencies

gracewilcox:


gracewilcox:
removed 'templatesgroup' spector test due to bug in recently added test: https://github.com/Azure/autorest.go/issues/1822

PR #1825, opened 2026-01-15
Update pnpm lock file and fix lropaging

JiaqiZhang-Dev:


packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -1051,12 +1051,10 @@ export class ClientAdapter {
...
     if (sdkMethod.kind === 'lropaging' || sdkMethod.kind === 'paging') {
       // grab the paged response envelope type from the operation responses
-      for (const httpResp of sdkMethod.operation.responses) {
-        if (httpResp.type) {
-          sdkResponseType = httpResp.type;
-          break;
-        }
-      }
+      sdkResponseType = sdkMethod.kind === 'lropaging' 
+        ? sdkMethod.lroMetadata.finalResponse?.result
```
tadelesh:
```suggestion
        ? sdkMethod.lroMetadata.finalResponse?.envelopResult
```

tadelesh: I'm wrong. We have internal logic to handle such case: https://github.com/Azure/azure-sdk-for-go/blob/8ed7d81fef1d5a1d20918cc9aedc39cc598f69ce/sdk/azcore/internal/pollers/op/op.go#L83.

tadelesh (approved):

packages/typespec-go/src/tcgcadapter/clients.ts
```diff
@@ -1051,12 +1051,10 @@ export class ClientAdapter {
...
+        : sdkMethod.operation.responses[0].type;
```
lirenhe: Can we add this case for LRO paging?  https://github.com/Azure/typespec-azure/pull/3695

PR #1836, opened 2026-01-23
time refactor and add tests

gracewilcox:
Minor refactor and adding tests for the newly supported spector PlainTime scenario.


gracewilcox:
redundant PR, plain time test was already added in: https://github.com/Azure/autorest.go/pull/1835

PR #1850, opened 2026-01-30
Add documentation test

jliusan:
Add documentation test


tadelesh (approved):

packages/typespec-go/test/http-specs/documentationgroup/zz_constants.go
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// Code generated by Microsoft (R) Go Code Generator. DO NOT EDIT.
+
+package documentationgroup
+
+// BulletPointsEnum - This tests really long bullet points in enum documentation to see how wrapping and formatting are handled.
+// This should wrap around correctly and maintain proper indentation for each line.
+// - Simple bullet point. This bullet point is going to be very long to test how text wrapping is handled in bullet points
+// within documentation comments. It should properly indent the wrapped lines.
```
jhendrixMSFT: Hmm this looks wrong.  Should be indented.

jhendrixMSFT (changes requested):
Looks like there are various formatting issues.  We should fix them before committing this test.

PR #1859, opened 2026-02-06
Refine logic of customized datetime encoding format

JiaqiZhang-Dev:
Adopt the proposal for customized datetime encoding format
image

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -258,16 +258,8 @@ export class TypeAdapter {
   }
 
   private getTimeType(type: tcgc.SdkDateTimeType, utc: boolean): go.WireType {
-    try {
-      const encoding = getDateTimeEncoding(type.encode);
-      let datetime = this.types.get(encoding);
-      if (datetime) {
-        return <go.Time>datetime;
-      }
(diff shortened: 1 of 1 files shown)


packages/typespec-go/src/tcgcadapter/types.ts
```diff
@@ -256,15 +257,25 @@ export class TypeAdapter {
...
   }
 
-  private getTimeType(encode: string, utc: boolean): go.Time {
-    const encoding = getDateTimeEncoding(encode);
-    let datetime = this.types.get(encoding);
-    if (datetime) {
-      return <go.Time>datetime;
+  private getTimeType(type: tcgc.SdkDateTimeType, utc: boolean): go.WireType {
+    try {
+      const encoding = getDateTimeEncoding(type.encode);
```
jhendrixMSFT: Instead of using `try/catch` can we change `getDateTimeEncoding` to return `undefined` on unknown encodings?

gracewilcox: Currently, if there's an unknown `datetime` encoding it is `undefined`. Which is not ideal since `models_serde.go` tries to import `datetime.undefined`, causing a compiler error.

example:
image

JiaqiZhang-Dev:
close this PR since we do not adopt unknown datetime encoding format

PR #1861, opened 2026-02-09
[typespec] use client name for response header fields

gracewilcox:
Fixing bug where the name of the response field used the serialized name instead of the actual client name

### DIFF
--- a/packages/typespec-go/package.json
+++ b/packages/typespec-go/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@azure-tools/typespec-go",
-  "version": "0.9.3",
+  "version": "0.10.0",
   "description": "TypeSpec emitter for Go SDKs",
   "type": "module",
   "exports": {


packages/typespec-go/test/azure-http-specs/azure/core/traitsgroup/zz_responses.go
```diff
@@ -19,9 +19,9 @@ type TraitsClientSmokeTestResponse struct {
...
 
-	// The entity tag for the response.
-	ETag *string
-
 	// An opaque, globally-unique, client-generated string identifier for the request.
-	XMSClientRequestID *string
+	ClientRequestID *string
+
+	// The entity tag for the response.
+	EtagHeader *string
```
jhendrixMSFT: Hmm I thought we had code in the emitter to ensure that this is always `ETag` (note the capital `T`).

packages/typespec-go/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Release History
 
-## 0.9.3 (Unreleased)
+## 0.10.0 (2026-02-11)
```
jhendrixMSFT: Be sure to also update the version in `package.json`.

gracewilcox: Thank you! Good catch

Issue #1888, opened 2026-03-18
Add support for `Azure.Core.eTag`


gracewilcox:
We have Etag support in the generator, but typespec added a new specific type `Azure.Core.eTag`. Need to add support in our emitter.

Example in typespec: https://github.com/Azure/azure-rest-api-specs/blob/727095157c5a4d23ed25aa2d90bd5d8551e8ef33/specification/storage/Microsoft.BlobStorage/models.tsp#L639

PR #1889, opened 2026-03-18
Fix handling of Azure.Core.eTag types in headers

jhendrixMSFT:
Allow core Etag types to be used in request/response headers. Fixed codegen to properly convert strings to/from azcore.ETag types. Updated storage tsp to verify correctness.

Fixes https://github.com/Azure/autorest.go/issues/1888


packages/typespec-go/test/tsp/Microsoft.BlobStorage/models.tsp
gracewilcox:
```suggestion
  @Xml.name("Etag") eTag: Azure.Core.eTag;
```

RickWinter (approved):

packages/typespec-go/test/tsp/Microsoft.BlobStorage/models.tsp
gracewilcox:
```suggestion
  sourceIfMatch?: Azure.Core.eTag;
```

packages/typespec-go/test/local/azblob/zz_client.go
```diff
@@ -134,13 +134,13 @@ func (client *Client) acquireLeaseCreateRequest(ctx context.Context, duration in
 	}
 	req.Raw().URL.RawQuery = reqQP.Encode()
 	if options != nil && options.IfMatch != nil {
-		req.Raw().Header["If-Match"] = []string{*options.IfMatch}
+		req.Raw().Header["If-Match"] = []string{string(*options.IfMatch)}
```
gracewilcox: is this right? generating locally and currently in azblob's main branch and it's different

jhendrixMSFT: It's correct here yes.  `options.IfMatch` is an `*azcore.ETag` which needs to be cast to a `string` when setting the header value.

How's it different from what you're observing?

PR #1900, opened 2026-03-27
Add support for some clientOption decorator key/values

jhendrixMSFT:
monomorphicResponseFiedName is applied on operations that return a monomorphic response to customize the field name in the response envelope instead of the default naming algorithm.
omitSerdeMethods is applied on models to elide emitting un/marshalers. Moved some common adapter code to helpers.ts.

### DIFF
--- a/packages/typespec-go/src/tcgcadapter/types.ts
+++ b/packages/typespec-go/src/tcgcadapter/types.ts
@@ -528,7 +528,6 @@ export class TypeAdapter {
                 target: model.__raw?.node,
               };
           }
-          break;
         default:
           return {
             msg: `invalid client option ${name} on model ${model.name}`,


packages/typespec-go/src/tcgcadapter/helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+export type ClientOptionProcessor = (name: string, value: string) => {msg: string, target?: tsp.DiagnosticTarget} | undefined;
+
+/**
+ * processes any clientOption decorator key/value pairs.
+ *
+ * @param program the tsp Program currently in scope, used to report warning diagnostics for unhandled key/value pairs
+ * @param decorators the array of decorators to process, can be empty
+ * @param processor called for each clientOption key/value pair in decorators
+ */
+export function processClientOptions(program: tsp.Program, decorators: Array<tcgc.DecoratorInfo>, processor: ClientOptionProcessor): void {
```
jhendrixMSFT: New code.

PR #1951, opened 2026-04-30
Fix encoding of space characters in query params

jhendrixMSFT:
The Go standard library call url.Values.Encode() encodes space characters as + which is not valid for URIs (it's form-urlencoded). After encoding, replace all + characters with %20 which is the correct encoding for URIs.


jhendrixMSFT:
For context, storage was doing this as an autorest transform.

https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azblob/internal/generated/autorest.md#updating-encoding-url-golang-adds--which-disrupts-encoding-with-service

chlowell (approved):

PR #1955, opened 2026-05-08
Fix example generation for literal typed pointers (e.g. *int32)

JiaqiZhang-Dev:
## Problem

**real case:** https://dev.azure.com/azure-sdk/public/_build/results?buildId=6256508&view=logs&j=b70e5e73-bbb6-5567-0939-8415943fadb9&t=57f096e1-36e2-50f7-6e90-7a231ead9f90
**tsp**: https://github.com/Azure/azure-rest-api-specs/blob/main/specification/securityinsights/resource-manager/Microsoft.SecurityInsights/SecurityInsights/ThreatIntelligenceInformation.tsp#L54

When generating example test files (`*_example_test.go`), literal typed fields like `status?: 1` (an `int32` literal in TypeSpec) produced incorrect Go code:

```go
Status: to.Ptr(1),  // Go infers *int, not *int32
```

This caused compilation errors:
```
cannot use to.Ptr(1) (value of type *int) as *int32 value in struct literal
```

## Root Cause

In `packages/codegen.go/src/core/example.ts`, the `getPointerValue()` function grouped the `'literal'` case with `'string'`, `'any'`, etc., always producing an untyped `to.Ptr(value)`. For numeric literals where the field type is `*int32`, Go infers the untyped `1` as `int`, causing a type mismatch.

## Fix

Unwrap the literal type and recursively delegate to the inner type's handler. For example, a `Literal<Scalar('int32')>` now delegates to the `'scalar'` case, which correctly produces:

```go
Status: to.Ptr[int32](1),
```

## Changes

- **`packages/codegen.go/src/core/example.ts`**: Moved `'literal'` out of the generic `to.Ptr()` case and added recursive unwrapping to delegate to the inner type's handler.
- **`packages/typespec-go/test/tsp/Test.Management/models.tsp`**: Added `AvailabilityProperties` model with a literal `int32` field (`status?: 1`) and referenced it from `BodyRootProperties`.
- **`packages/typespec-go/test/tsp/Test.Management/examples/`**: Added example JSON exercising the literal int32 field.
- **`packages/typespec-go/test/local/armtest/`**: Regenerated output confirming correct `to.Ptr[int32](1)` generation.


packages/codegen.go/src/core/example.ts
```diff
@@ -402,11 +402,13 @@ function getPointerValue(type: go.WireType, valueString: string, byValue: boolea
...
     case 'etag':
-    case 'literal':
     case 'string':
     case 'time':
       if (imports) imports.add('github.com/Azure/azure-sdk-for-go/sdk/azcore/to');
       return `to.Ptr(${valueString})`;
+    case 'literal':
+      // unwrap the literal and delegate to the inner type's case
+      return getPointerValue(type.type as go.WireType, valueString, byValue, imports);
     case 'scalar': {
```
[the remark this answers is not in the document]
JiaqiZhang-Dev: fixed

packages/typespec-go/test/local/armtest/zz_bodyroots_client_example_test.go
```diff
@@ -30,3 +30,39 @@ func ExampleBodyRootsClient_Action() {
 		log.Fatalf("failed to finish the request: %v", err)
 	}
 }
+
+// Generated from example definition: 2025-01-01/BodyRoots_get_with_availability.json
+func ExampleBodyRootsClient_Get() {
```
JiaqiZhang-Dev, line 35: @copilot pls write a put operation test case and example for this, since response is always commented

packages/typespec-go/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release History
 
+## 0.11.3 (unreleased)
+
+### Bugs Fixed
```
JiaqiZhang-Dev, line 5: @copilot pls update changelog for this pr

tadelesh (approved):

jhendrixMSFT (approved):

PR #1964, opened 2026-05-14
Don't adapt client accessors for ARM clients

jhendrixMSFT:
We use a custom client factory instead of hierarchical clients for ARM. This has been working but didn't account for clients with service operations and sub-clients.

Fixes https://github.com/Azure/autorest.go/issues/1963


packages/typespec-go/test/local/armcognitiveservices/zz_client.go
jhendrixMSFT, line 1: Note that we have an older version of this tsp which demonstrates the issue/fix.  Fortunately, the tsp has since been updated in the specs repo, so there will be no breaking change in the shipping SDK.

chlowell (approved):

PR #1965, opened 2026-05-14
Clients can have more than one API version

jhendrixMSFT:
Changed the singular ConstantDef on a client to an array of ConstantDefs. The const names have been changed to use their underlying values with any illegal characters removed.  This also resulted in removing duplicate constant definitions across clients.
The const definitions have moved from *client.go to constants.go.

Fixes https://github.com/Azure/autorest.go/issues/1962


packages/autorest.go/test/network/armnetwork/zz_interfaces_client.go
```diff
@@ -17,12 +17,10 @@ import (
 	"strings"
 )
 
-const defaultInterfacesClientVersion string = "2022-09-01"
-
 // InterfacesClient contains the methods for the NetworkInterfaces group.
 // Don't use this type directly, use NewInterfacesClient() instead.
 //
-// Generated from API version 2022-09-01
+// Generated from API versions 2022-09-01, 2018-10-01
```
jhendrixMSFT: Here's an example where multiple API versions are used within a single client.

tadelesh (approved):

PR #1991, opened 2026-06-05
Improved handling of Content-Type parameters part 2

jhendrixMSFT:
For modeled payloads, set Content-Type inline when it comes from an explicit parameter.
Addendum to db10ab8.

### DIFF
--- a/packages/codegen.go/src/core/operations.ts
+++ b/packages/codegen.go/src/core/operations.ts
@@ -1267,9 +1267,11 @@ function createProtocolRequest(azureARM: boolean, method: go.MethodType | go.Nex
 
       // there's a corner case where we have a content-type param with no body
       // param. for this case we still need to set the header inline.
-      if (methodParamGroups.bodyParam) {
+      const bodyParam = methodParamGroups.bodyParam;
+      if (bodyParam?.bodyFormat === 'binary' || bodyParam?.bodyFormat === 'Text') {
         // if the content-type is from a param, then the param will be passed
-        // explicitly to runtime.SetBody() so don't emit code to set it inline
+        // explicitly to runtime.SetBody() so don't emit code to set it inline.
+        // NOTE: only binary/text payloads call runtime.SetBody().
         if (go.isClientSideDefault(param.style)) {
           text += emitClientSideDefault(param as go.HeaderScalarParameter, param.style, () => '', imports, indent);
           continue;
--- a/packages/typespec-go/test/local/azregressions/zz_client.go
+++ b/packages/typespec-go/test/local/azregressions/zz_client.go
@@ -712,6 +712,45 @@ func (client *Client) optionalBodyPostCreateRequest(ctx context.Context, options
 	return req, nil
 }
 
+// PayloadWithExplicitContentType - explicit content-type param is set when creating the request
+// If the operation fails it returns an *azcore.ResponseError type.
+//   - options - ClientPayloadWithExplicitContentTypeOptions contains the optional parameters for the Client.PayloadWithExplicitContentType
+//     method.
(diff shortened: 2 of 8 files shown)


packages/typespec-go/test/local/azregressions/zz_client.go
```diff
@@ -712,6 +712,45 @@ func (client *Client) optionalBodyPostCreateRequest(ctx context.Context, options
...
+}
+
+// payloadWithExplicitContentTypeCreateRequest creates the PayloadWithExplicitContentType request.
+func (client *Client) payloadWithExplicitContentTypeCreateRequest(ctx context.Context, contentType PayloadWithExplicitContentTypeRequestContentType, thing SomeModel, _ *ClientPayloadWithExplicitContentTypeOptions) (*policy.Request, error) {
+	urlPath := "/payload-with-explicit-content-type"
+	req, err := runtime.NewRequest(ctx, http.MethodPost, runtime.JoinPaths(client.endpoint, urlPath))
+	if err != nil {
+		return nil, err
+	}
+	req.Raw().Header["Content-Type"] = []string{string(contentType)}
```
jhendrixMSFT, line 747: This line was absent before the fix.

packages/codegen.go/src/core/operations.ts
```diff
@@ -1267,9 +1267,11 @@ function createProtocolRequest(azureARM: boolean, method: go.MethodType | go.Nex
 
       // there's a corner case where we have a content-type param with no body
       // param. for this case we still need to set the header inline.
-      if (methodParamGroups.bodyParam) {
+      const bodyParam = methodParamGroups.bodyParam
+      if (bodyParam?.bodyFormat === 'binary' || bodyParam?.bodyFormat === 'Text') {
         // if the content-type is from a param, then the param will be passed
-        // explicitly to runtime.SetBody() so don't emit code to set it inline
+        // explicitly to runtime.SetBody() so don't emit code to set it inline.
+        // NOTE: only binary/text payloads call runtime.SetBody().
```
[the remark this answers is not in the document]
jhendrixMSFT: The tsp additions exercise the codegen, demonstrating the fix, which is sufficient.

packages/codegen.go/src/core/operations.ts
```diff
@@ -1267,9 +1267,11 @@ function createProtocolRequest(azureARM: boolean, method: go.MethodType | go.Nex
 
       // there's a corner case where we have a content-type param with no body
       // param. for this case we still need to set the header inline.
-      if (methodParamGroups.bodyParam) {
+      const bodyParam = methodParamGroups.bodyParam;
+      if (bodyParam?.bodyFormat === 'binary' || bodyParam?.bodyFormat === 'Text') {
```
heaths, line 1271: The `Text` casing is worrying me a little. Should you maybe normalize and compare instead? This seems risky.

heaths (approved):
One suggestion but I see nothing blocking.

packages/codegen.go/src/core/operations.ts
heaths: Or is `bodyFormat` actually declared as such and thus checked by tsc? If so, that seems okay too...though still a bit odd that only `Text`, IIRC, is PascalCase compared to the others I've seen. 🤷‍♂️

jhendrixMSFT: In this case, `bodyFormat` is a union.
```ts
export type BodyFormat = 'JSON' | 'XML' | 'Text' | 'binary';
```
So there's no risk of casing mismatch.

jhendrixMSFT: IIRC the first three entries in the union are splatted directly into the code which is why they're cased that way.

PR #1992, opened 2026-06-08
Use RFC7231 for tsp instead of RFC1123

jhendrixMSFT:
Fixed up some tests to ensure functionality.

### DIFF
--- a/packages/autorest.go/src/generator/generator.ts
+++ b/packages/autorest.go/src/generator/generator.ts
@@ -53,7 +53,7 @@ export async function generateCode(host: AutorestExtensionHost) {
       },
       { filePrefix },
     );
-    await emitter.emit();
+    await emitter.emit('openapi');
   } catch (E) {
     if (debug) {
       console.error(`${fileURLToPath(import.meta.url)} - FAILURE  ${JSON.stringify(E)} ${(<Error>E).stack}`);
--- a/packages/codegen.go/src/core/models.ts
+++ b/packages/codegen.go/src/core/models.ts
@@ -17,9 +17,10 @@ export interface ModelsSerDe {
  *
  * @param pkg contains the package content
  * @param options the emitter options
+ * @param emitter the emitter type
  * @returns the text for the files or the empty string
  */
-export function generateModels(pkg: go.PackageContent, options: go.Options): ModelsSerDe {
+export function generateModels(pkg: go.PackageContent, options: go.Options, emitter: 'openapi' | 'tsp'): ModelsSerDe {
   if (pkg.models.length === 0) {
     return {
       models: '',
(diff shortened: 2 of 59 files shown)


packages/typespec-go/test/http-specs/special-headers/condreqgroup/fake/zz_conditionalrequest_server.go
```diff
@@ -98,7 +99,13 @@ func (c *ConditionalRequestServerTransport) dispatchHeadIfModifiedSince(req *htt
...
 		return nil, &nonRetriableError{errors.New("fake for method HeadIfModifiedSince not implemented")}
 	}
-	ifModifiedSinceParam, err := parseOptional(getHeaderValue(req.Header, "If-Modified-Since"), func(v string) (time.Time, error) { return time.Parse(time.RFC1123, v) })
+	ifModifiedSinceParam, err := parseOptional(getHeaderValue(req.Header, "If-Modified-Since"), func(v string) (time.Time, error) {
+		p, parseErr := strconv.ParseInt(v, 10, 64)
+		if parseErr != nil {
+			return time.Time{}, parseErr
+		}
+		return time.Unix(p, 0).UTC(), nil
+	})
```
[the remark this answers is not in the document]
jhendrixMSFT: these issues have been fixed

packages/autorest.go/src/generator/generator.ts
```diff
@@ -53,7 +53,7 @@ export async function generateCode(host: AutorestExtensionHost) {
       },
       { filePrefix },
     );
-    await emitter.emit();
+    await emitter.emit('openapi');
```
heaths, line 56: Is this specifically for OpenAI, or should it generically be considered for the "unbranded" effort?

packages/codegen.go/src/core/models.ts
```diff
@@ -189,7 +190,18 @@ export function generateModels(pkg: go.PackageContent, options: go.Options): Mod
...
+    let rfc7231 = '';
+    switch (emitter) {
+      case 'openapi':
+        rfc1123 = ' | datetime.RFC1123';
+        break;
+      case 'tsp':
+        rfc7231 = ' | datetime.RFC7231';
+        break;
+    }
+    serdeTextBody += `${indent.get()}datetime.PlainDate | datetime.PlainTime${rfc1123} | datetime.RFC3339${rfc7231} | datetime.Unix\n`;
```
heaths: If they are mutually exclusive, why not just a single local var to concat? At first glance this was a little...off...to mentally process.

heaths (approved):

packages/autorest.go/src/generator/generator.ts
jhendrixMSFT: It's for `OpenAPI` i.e. swagger (it's hard to read given it's all lower case).

Note that this will evaporate once we deprecate Autorest.

packages/codegen.go/src/core/models.ts
jhendrixMSFT: Yeah it is a bit ugly.  I did it this way so that the values are sorted.  Will improve this.

jhendrixMSFT: fixed

heaths (approved):

